PR #44, opened 2020-07-17
Validate PING, DNS and TCP targets

mem:
Add validations for the targets of PING, DNS and TCP checks. Make sure
we are getting something that is either a valid IP address or a valid
hostname. The API will also perform this validation, possibly more
strictly, so this is only meant to give the user some immediate
feedback.

### DIFF
--- a/src/validation.test.ts
+++ b/src/validation.test.ts
@@ -1,0 +1,202 @@
+import { validateCheck, CheckValidation } from 'validation';
+import { Check, CheckType } from 'types';
+
+describe('trivial cases', () => {
+  it('should be a valid http check', () => {
+    const check: Check = {
+      job: 'job-1',
+      target: 'https://grafana.com/',
+      frequency: 10000,
+      timeout: 1000,
+      labels: [],
+      probes: [1],
+      enabled: true,
+      settings: {http: {
+      }},
+    };
+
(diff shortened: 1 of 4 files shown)


src/validation.ts
```diff
@@ -1,5 +1,7 @@
 import { Check, CheckType, Settings, HttpSettings, PingSettings, DnsSettings, TcpSettings, Label } from 'types';
 import { checkType } from 'utils';
+const punycode = require('punycode');
+const ipaddress = require('ip-address');
```
ryantxu, line 4: can we import rather than require?

src/validation.test.ts
```diff
@@ -0,0 +1,19 @@
+import { validateCheck } from 'validation';
+import { Check } from 'types';
+
+describe('validation test', () => {
+  it('should test target', () => {
+    const check: Check = {
```
ryantxu, line 6: @mem -- can you add more tests here?  that will help whoever has to look at this next :)

src/validation.ts
mem: thanks!

src/validation.test.ts
mem: I added a bunch, thanks for bootstrapping! :smile:

PR #48, opened 2020-07-23
fix logs link in summary dashboard

woodsaj:
https://github.com/grafana/grafana/issues/26559 prevents input
placeholders from working correctly when other variables are in the same
field.  This works around the issue by placing the DS_SM_LOGS value into
a "constant" hidden template variable.  This hidden variable is then
used in the link url

### DIFF
--- a/src/dashboards/sm-summary.json
+++ b/src/dashboards/sm-summary.json
@@ -7,6 +7,14 @@
       "type": "datasource",
       "pluginId": "prometheus",
       "pluginName": "Prometheus"
+    },
+    {
+      "name": "DS_SM_LOGS",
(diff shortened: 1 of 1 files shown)


src/dashboards/sm-summary.json
```diff
@@ -559,6 +567,24 @@
...
+        "name": "_DS_SM_LOGS",
+        "options": [
+          {
+            "value": "${DS_SM_LOGS}",
+            "text": "${DS_SM_LOGS}"
+          }
+        ],
+        "query": "${DS_SM_LOGS}",
+        "skipUrlSync": false,
+        "type": "constant"
```
mem, line 587: aha! this is the bit I didn't figure out!

PR #50, opened 2020-07-30
Dns settings refactor

rdubrock:
I notice that the DNS validation fields were tilting a bit, and thought that might be a nice first PR. I also took the opportunity to refactor that component a bit and set up some tests using react testing library. This PR is very much just trying things out and getting familiar with the tooling, any and all feedback very welcome!.

Before:

Screen Shot 2020-07-29 at 8 05 18 PM

After:

Screen Shot 2020-07-29 at 5 13 59 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,10 +13,14 @@
   "devDependencies": {
     "@grafana/data": "7.0.6",
     "@grafana/runtime": "7.0.6",
-    "@grafana/toolkit": "7.0.6",
+    "@grafana/toolkit": "^7.1.1",
     "@grafana/ui": "7.0.6",
+    "@testing-library/dom": "^7.21.6",
+    "@testing-library/react": "^10.4.7",
+    "@testing-library/user-event": "^12.0.17",
+    "@types/ip-address": "latest",
     "@types/lodash": "latest",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -13,10 +13,14 @@
   "devDependencies": {
     "@grafana/data": "7.0.6",
     "@grafana/runtime": "7.0.6",
-    "@grafana/toolkit": "7.0.6",
```
rdubrock, line 16: Not sure we need this, I just gave it a go to see if that version of toolkit included a jest version > 24 (doesn't seem to)

woodsaj:
This looks good to me.

the next step to improving this form and other validation forms that are similar, is to reduce the repitition.
ie, instead of showing the user 6 lists of validation rules, we can make it a single list, and allow users to dynamically select which field is being validated, and whether the expression should be inverted.

eg
image

woodsaj:
To be clear, i think this PR should be merged, and further improvements applied in future PRs.

package.json
ryantxu: sounds like we should make sure it does in the future!

Issue #61, opened 2020-08-06
Entering negative numbers in latitude / longitude is awkward


mem:
When adding a new probe, the latitude and longitude fields require you to enter negative numbers for west and south, but that's pretty weird.

If you click the field the cursor is after the 0. If you hit backspace, that doesn't work. You cannot enter a minus. What you have to do is enter the value back go back to the first character and enter the minus.

PR #63, opened 2020-08-10
Probe editor validations

rdubrock:
Fixes https://github.com/grafana/synthetic-monitoring-app/issues/61

Adds input specific validation via migration to using form state, error messaging, and tests for probe creation/editing

I also moved each input onto it's own line, the introduction of validation messages changes the height of inputs when there is a message and makes horizontally aligning them really jarring when an input moves from a valid => invalid state

Before:

Screen Shot 2020-08-10 at 5 27 46 PM

After:
Screen Shot 2020-08-12 at 3 38 37 PM

### DIFF
--- a/src/components/ProbeStatus.tsx
+++ b/src/components/ProbeStatus.tsx
@@ -1,0 +1,69 @@
+import React, { FC, useState } from 'react';
+import { Badge, BadgeColor, Button, Container, ConfirmModal, Legend, IconName } from '@grafana/ui';
+import { Probe, OrgRole } from 'types';
+import { SMDataSource } from 'datasource/DataSource';
+import { hasRole } from 'utils';
+import { UptimeGauge } from './UptimeGauge';
+
+interface Props {
+  probe: Probe;
+  instance: SMDataSource;
+  onResetToken: () => void;
+}
+
+const ProbeStatus: FC<Props> = ({ probe, instance, onResetToken }) => {
+  const [showResetModal, setShowResetModal] = useState(false);
+
+  if (!probe) {
+    return null;
+  }
+  let isEditor = !probe.public && hasRole(OrgRole.EDITOR);
+  let onlineTxt = 'Offline';
+  let onlineIcon = 'heart-break' as IconName;
+  let color = 'red' as BadgeColor;
+  if (probe.online) {
+    onlineTxt = 'Online';
+    onlineIcon = 'heart';
+    color = 'green';
+  }
+
+  const handleResetToken = () => {
+    onResetToken();
+    setShowResetModal(false);
+  };
+
+  return (
+    <Container margin="md">
+      <Legend>
+        Status: &nbsp;
+        <Badge color={color} icon={onlineIcon} text={onlineTxt} />
+      </Legend>
+      {!probe.public && (
+        <Container>
+          <Button variant="destructive" onClick={() => setShowResetModal(true)} disabled={!isEditor}>
+            Reset Access Token
+          </Button>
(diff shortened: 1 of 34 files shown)


src/components/ProbeStatus.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const handleResetToken = () => {
+    onResetToken();
+    setShowResetModal(false);
+  };
+
+  return (
+    <Container margin="md">
+      <div
+        className={css`
```
ryantxu, line 53: maybe consider this approach:
https://github.com/grafana/grafana/blob/master/contribute/style-guides/styling.md#styling-with-theme

src/components/ProbeStatus.tsx
```diff
@@ -0,0 +1,114 @@
...
+}
+
+const getStyles = stylesFactory((theme: GrafanaTheme) => ({
+  legend: css`
+    margin: ${theme.spacing.sm} ${theme.spacing.sm} 0 0;
+    width: auto;
+  `,
+  container: css`
+    padding-left: ${theme.spacing.sm}
+    margin-bottom: ${theme.spacing.md}
```
s-h-a-d-o-w, line 39: Does that work? Because you're missing semicolons here and IIRC, emotion isn't quite so smart to put them in anyway.
If you use vscode, I'd recommend installing the extension `vscode-styled-components` because that would've thrown an error. (And it provides syntax highlighting :) ).

src/components/ProbeStatus.tsx
```diff
@@ -0,0 +1,114 @@
...
+    color: 'red',
+    icon: 'heart-break',
+  };
+};
+
+const ProbeStatus: FC<Props> = ({ probe, instance, onResetToken }) => {
+  const [showResetModal, setShowResetModal] = useState(false);
+  if (!probe) {
+    return null;
+  }
```
s-h-a-d-o-w, line 67: Did React not complain about this? One of the hooks rules is violated here. I'd recommend installing the react-hooks eslint plugin (and import-order if you're into that sort of thing ;) ). See Alerting UI for how to make it work with grafana's base config.

Because it would throw:
```
React Hook "useTheme" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return? eslintreact-hooks/rules-of-hooks
```

src/page/ChecksPage.tsx
```diff
@@ -290,14 +290,14 @@ export class ChecksPage extends PureComponent<Props, State> {
         </div>
         <section className="card-section card-list-layout-list">
           <ol className="card-list">
-            {filteredChecks.map(check => {
+            {filteredChecks.map((check, index) => {
               const checkId: number = check.id || 0;
               if (!check.id) {
                 return;
               }
```
s-h-a-d-o-w, line 297: I know you didn't write this @rdubrock , still wanted to note that that should probably be refactored next time around. :) 
Because first, a default is used if `check.id` is falsy but then if `check.id` is falsy, we return early? => Invert the order and get rid of the default.

src/page/ChecksPage.tsx
```diff
@@ -290,14 +290,14 @@ export class ChecksPage extends PureComponent<Props, State> {
...
               const checkType = Object.keys(check.settings)[0];
               return (
-                <li className="card-item-wrapper">
+                <li className="card-item-wrapper" key={index}>
```
s-h-a-d-o-w, line 300: If `checkId` happens to be unique, it should be used here instead of `index`.

Issue #51, opened 2020-07-30
DNS is missing list of return codes


mem:
The list of return codes presented to the user should be expanded to include all the values supported by the backend. The default value should be "NOERROR".

mem:
The code used by BBE (github.com/miekg/dns@v1.1.29) currently understands this:

```
NOERROR
BADALG
BADCOOKIE
BADKEY
BADMODE
BADNAME
BADSIG
BADTIME
BADTRUNC
BADVERS
FORMERR
NOTAUTH
NOTIMP
NOTZONE
NXDOMAIN
NXRRSET
REFUSED
SERVFAIL
YXDOMAIN
YXRRSET
```

PR #64, opened 2020-08-11
Dns return codes

rdubrock:
Fix for https://github.com/grafana/synthetic-monitoring-app/issues/51

Things to note
 - NOERROR is now included as a default value for valid DNS response codes
 - The complete list of DNS response codes are now available in the UI
 - The DnsSettings form got a refactor


src/utils.ts
```diff
@@ -177,3 +177,8 @@ export const parseUrl = (url: string) => {
     return;
   }
 };
+
+export function enumToStringArray<T>(enumObject: T) {
+  const uniqueKeys = new Set(Object.keys(enumObject));
+  return Array.from(uniqueKeys);
```
Clarity-89, line 183: Can be simplified to `return [...new Set(Object.keys(enumObject))]`. Also I think it's a good idea to add comments to util functions with a short description of what they do, seems like this one returns unique keys of an object, but the function name doesn't reflect that.

src/components/DnsSettings.tsx
```diff
@@ -1,313 +1,239 @@
...
 interface Props {
   settings: Settings;
   isEditor: boolean;
   onUpdate: (settings: Settings) => void;
 }
 
-interface State extends DnsSettings {
-  showAdvanced: boolean;
-  showValidation: boolean;
+const RESPONSE_CODES = enumToStringArray(DnsResponseCodes).map(responseCode => ({
```
Clarity-89, line 16: Consider moving all these constants to a separate file so they don't add extra clutter.

src/components/DnsSettings.tsx
```diff
@@ -1,313 +1,239 @@
...
-      return;
-    }
-    const validateAuthorityRRS = {
-      failIfMatchesRegexp: validations.failIfMatchesRegexp,
-      failIfNotMatchesRegexp: validations.failIfNotMatchesRegexp,
-    };
-    this.setState({ validateAuthorityRRS }, this.onUpdate);
-  };
+const DnsSettingsForm: FC<Props> = ({ settings, isEditor, onUpdate }) => {
+  const initialValues = {
```
Clarity-89, line 85: Wonder if we could defined a default object and merge it with `settings.dns` instead?
```
const defaultValues = {
  recordType: DnsRecordType.A,
  server: '8.8.8.8'
  //.... 
}
const initialValues = {...defaultValues, ...settings.dns};
```

src/components/DnsSettings.tsx
```diff
@@ -1,313 +1,239 @@
...
-    ];
-    return (
-      <Container>
+  const [dnsSettings, dispatch] = useReducer(dnsSettingsReducer, initialValues);
+  const [showValidation, setShowValidation] = useState(false);
+  const [showAdvanced, setShowAdvanced] = useState(false);
+
+  useEffect(() => {
+    onUpdate({ dns: dnsSettings });
+  }, [dnsSettings]);
```
Clarity-89, line 110: Will this work for the deeply nested values? Afaik shallow comparison is used for hooks dependencies.

src/components/DnsSettings.tsx
```diff
@@ -1,313 +1,239 @@
...
+
+  return (
+    <Container>
+      <HorizontalGroup>
+        <Field label="Record Type" description="DNS record type to query for" disabled={!isEditor}>
+          <Select
+            value={dnsSettings.recordType}
+            options={RECORD_TYPES}
+            onChange={(selected: SelectableValue<DnsRecordType>) =>
+              dispatch({
```
Clarity-89, line 120: Using reducers here or any other state management is not necessary since our Form component (as a wrapper for [react-hook-form](https://react-hook-form.com/v5/api) gives that for free. Here's an example of recent code used in [Enterprise](https://github.com/grafana/grafana-enterprise/blob/master/src/public/reports/ReportPage.tsx#L199). Using `register` with `InputControl` eliminates 99% of use cases for custom form management. Takes a bit of practice to get used to, but in the end makes components way simpler.

Probably the biggest challenge here would be to make it work with the `onUpdate` prop since I assume it has to be called every time a field is updated vs once on form submit.

src/components/DnsSettings.tsx
rdubrock: Good question! I was banking on referential equality working for me here since the reducer always returns a new copy, but that might not be the best idea

src/components/DnsSettings.tsx
rdubrock: That looks like it'll be a major improvement. I'm going to tackle that in it's own PR since it will involve quite a few changes.

src/components/DnsSettings.tsx
Clarity-89: There are some utils for deep dependency comparison but I'm not sure about performance implications of those. 

I might be missing some context but I'm not sure why do we need to store local state at all and not read it straight from the settings and modify them via `onUpdate`?

rdubrock: That's a good question (also don't have a ton of context yet). I think that would work, I'll tinker and see what happens. I think migrating to using Form state would make this all go away anyways

rdubrock: Good suggestion! Much cleaner this way

Issue #15, opened 2020-06-06
If the app cannot talk to the api, it does not display an error


mem:
If you set the API server to something invalid during set up and hit setup, no error is displayed.

PR #71, opened 2020-08-14
Api config error handling

rdubrock:
fixes https://github.com/grafana/synthetic-monitoring-app/issues/15

- Abstracts the tenant api setup step into it's own form
- Shows all types of errors returned from the setup api call (previously just the backend errors were showing up)
- Adds some tests for the tenant api setup form

### DIFF
--- a/src/components/TenantAPISetupForm.tsx
+++ b/src/components/TenantAPISetupForm.tsx
@@ -59,8 +59,8 @@ const TenantAPISetupForm: FC<Props> = ({ onSubmit, submissionError }) => {
                     validate: value => {
                       try {
                         new URL(value);
-                      } catch (e) {
-                        return e.message;
+                      } catch ({ message }) {
+                        return message;
                       }
                     },
                   })}
--- a/src/components/TenantAPISetupForm.tsx
+++ b/src/components/TenantAPISetupForm.tsx
@@ -4,7 +4,7 @@ import { DEFAULT_API_HOST } from './constants';
 
 interface FormValues {
   adminApiToken: string;
-  apiHost: string;
+  apiHost?: string;
 }
 
 interface Props {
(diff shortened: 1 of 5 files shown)


src/components/TenantAPISetupForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+interface Props {
+  onSubmit: (values: FormValues) => void;
+  submissionError?: string;
+}
+
+const TenantAPISetupForm: FC<Props> = ({ onSubmit, submissionError }) => {
+  const [showAdvanced, setShowAdvanced] = useState(false);
+  return (
+    <Form
+      onSubmit={(values: FormValues) => onSubmit(values)}
```
stevenvachon, line 19: This creates a new callback function on each render, which can be often. `onSubmit={onSubmit}` is more efficient and will call with all arguments received (which isn't a problem here).

src/components/TenantAPISetupForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+            isOpen={showAdvanced}
+          >
+            <HorizontalGroup>
+              <Field label="Backend Address" invalid={Boolean(errors.apiHost)} error={errors.apiHost?.message}>
+                <Input
+                  ref={register({
+                    validate: value => {
+                      try {
+                        new URL(value);
+                      } catch (e) {
```
stevenvachon, line 66: You could do
```js
} catch ({ message }) {
```

src/components/TenantAPISetupForm.tsx
rdubrock: I'm running into some collision of react-testing-library and react-hook-form. In the browser `onSubmit={onSubmit}` is just fine, but when running tests the `onSubmit` prop ends up getting passed the entire form element instead of the form values (but will work as expected when using the anonymous function) . I went ahead and changed it, but if you have any good ideas on why that might be I'm all ears!

stevenvachon: It looks like it passed in CI :shrug:

PR #74, opened 2020-08-20
Check search case insensitive

rdubrock:
- Makes filtering list in Check page case insensitive
- Refactors and add tests for check page


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "synthetic-monitoring-app",
-  "version": "0.0.7",
```
rdubrock, line 3: We had some versioning weirdness with the last release, this is just playing catch up to the current version

Issue #77, opened 2020-08-24
Targets can be an invalid URL


rdubrock:
https://suraj/dev is accepted as a valid http check url, but it shouldn't be. We should improve the quality of the URL validation.

PR #78, opened 2020-08-25
chore: use external library for url validation

rdubrock:
Fixes https://github.com/grafana/synthetic-monitoring-app/issues/77 (sort of)
Pulls in an external [url validation library](https://github.com/ogt/valid-url) to handle more URL validation cases

### DIFF
--- a/src/validation.ts
+++ b/src/validation.ts
@@ -2,6 +2,7 @@ import { Check, CheckType, Settings, HttpSettings, PingSettings, DnsSettings, Tc
 import { checkType } from 'utils';
 import * as punycode from 'punycode';
 import { Address4, Address6 } from 'ip-address';
+import validUrl from 'valid-url';
 
 export const CheckValidation = {
   job: validateJob,
@@ -138,12 +139,7 @@ export function validateSettingsTCP(settings: TcpSettings): boolean {
 }
 
 function validateHttpTarget(target: string): boolean {
-  try {
-    const url = new URL(target);
-    return url.protocol === 'https:' || url.protocol === 'http:';
-  } catch (_) {
-    return false;
-  }
+  return Boolean(validUrl.isWebUri(target));
 }
 
 function validateHostname(target: string): boolean {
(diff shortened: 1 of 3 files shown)


src/validation.ts
```diff
@@ -138,12 +139,7 @@ export function validateSettingsTCP(settings: TcpSettings): boolean {
 }
 
 function validateHttpTarget(target: string): boolean {
-  try {
-    const url = new URL(target);
-    return url.protocol === 'https:' || url.protocol === 'http:';
-  } catch (_) {
-    return false;
-  }
+  return Boolean(validUrl.isWebUri(target));
```
leventebalogh, line 142: I think we could also add a few extra test cases to `src/validation.test.ts`, just ensuring that what was false-positive until now (e.g. `target = "https://suraj/dev"`) is validated correctly in the future.

leventebalogh (approved):
LGTM, left one comment for tests 👍

src/validation.ts
electron0zero, line 142: we should add a test for this

rdubrock: Yeah... so I did that, and apparently https://suraj/dev isn't an invalid web uri according to spec. So this doesn't really fix that original case, but should catch other problems we weren't handling

rdubrock: We have a fair few tests right now for URL validation https://github.com/grafana/synthetic-monitoring-app/blob/215302c5c67df6e03c8ec8b17a17a414855490a6/src/validation.test.ts#L130

For the specific case you raised, it validates as a good URL, so I didn't add a case for it. Happy to add more cases, but I'm not sure what isn't covered

electron0zero (approved):
lgtm :+1:

src/validation.ts
mem: http://hostname/ (no domain) should be added to bad targets, because even if it's a valid URL, it's not a URL that we can work with (as probes have no way of resolving that).

rdubrock: Yeah, good point. I'll open another issue to follow up on this

PR #79, opened 2020-08-27
Refactor check health

rdubrock:
Currently we have a bit of prop drilling with the `instance` class which is muddying component APIs, increasing testing complexity, and requiring duplicative code when doing things like fetching metric data. This is the first step in extracting that out by putting the `instance` in a context in the `RootPage`. This allows for creating custom hooks that handle the currently duplicative logic and will allow us to get rid of the prop drilling in future PRs. 

This PR:
- Puts `instance` in a context
- Creates a custom hook that accesses the instance context and fetches metric data
- Refactors the CheckHealth component to use the new custom hook

### DIFF
--- a/src/components/CheckHealth.test.tsx
+++ b/src/components/CheckHealth.test.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import { render, waitFor } from '@testing-library/react';
+import { CheckHealth } from './CheckHealth';
+import { InstanceContext } from './InstanceContext';
+import { Check } from 'types';
+import { getInstanceMock, instanceSettings } from '../datasource/__mocks__/DataSource';
+
+const defaultCheck = {
+  id: 2,
+  tenantId: 1,
+  frequency: 60000,
+  offset: 0,
+  timeout: 2500,
+  enabled: true,
+  labels: [],
+  settings: {
+    ping: {
+      ipVersion: 'V4',
+      dontFragment: false,
+    },
+  },
+  probes: [1],
+  target: 'grafana.com',
+  job: 'tacos',
+  created: 1597928927.7490728,
+  modified: 1597928927.7490728,
+} as Check;
+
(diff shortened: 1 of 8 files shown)


src/components/CheckHealth.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    api: getInstanceMock(instanceSettings),
+  };
+  return render(
+    <InstanceContext.Provider value={{ instance, loading: false }}>
+      <CheckHealth check={check} />
+    </InstanceContext.Provider>
+  );
+};
+
+test('renders without crashing', async () => {
```
rdubrock, line 40: I'm not stoked on either these queries or the assertions, but I don't see any accessible ways to assert a type of icon

besartberisha (commented):
This looks good, but I just wanted to throw it here since as far as I can see the plugin doesn't have state management. 

Maybe it's worth looking into adding redux-toolkit: https://github.com/grafana/grafana/blob/master/contribute/style-guides/frontend.md#state-management as part of refactoring? 
It could probably make it easier to refactor and remove prop drilling all together.

rdubrock:
Good thought! I'll keep that in mind. I'm not yet sure this app has enough state to justify bringing in redux, since for the most part it's form based. I was thinking it could probably get where it needs to go just by using context.

s-h-a-d-o-w:
@rdubrock Since I wasn't here on Friday and missed this, I just wanted to mention that at least in my opinion, the point of Redux isn't necessarily managing huge amounts of state (in e.g. [this](https://www.youtube.com/watch?v=_ISAA_Jt9kI), IIRC it is mentioned that he's seen apps that use thousands of contexts just fine) but improved debuggability. The Alerting UI also doesn't have a lot of state but I've certainly needed to look into how the state changes over time here and there in order to figure out why on earth the state ended up being what it was. ;) (And also to easily inspect the current state across the app.)

PR #85, opened 2020-08-31
chore: add gcom release to ci

rdubrock:


robx:
This looks fine to me, though I don't really have experience with the gcom plugin release side of things.

For context, for the alerting-ui and easystart plugins, we're currently "releasing" them by writing packages to a Google Cloud storage bucket, and installing from there. That's primarily because we don't want them to be available for installation on your local Grafana instance. (And we don't particularly care about whether the plugin is listed there -- these being plugins is rather an implementation detail.)

PR #86, opened 2020-09-01
Probe page refactor

rdubrock:


src/components/ProbeEditor.tsx
```diff
@@ -14,40 +14,40 @@ import {
...
     if (probe.id) {
-      await instance.updateProbe({
+      await instance?.api.updateProbe({
         ...probe,
         ...formValues,
       });
       onReturn(true);
     } else {
-      const info = await instance.addProbe({
+      const info = await instance?.api.addProbe({
```
ryantxu, line 50: `instance?`  

maybe short circuit early if instance is null, or somehow make it not optional?

Issue #87, opened 2020-09-01
Success rate only fetches data once


rdubrock:
Any place the success rate metric shows up, it only fetches on mount. This means that if you don't refresh the page you don't get up to date metrics. We should periodically refresh the data so the user is always seeing the current success rate.

PR #88, opened 2020-09-01
feat: refresh metric data on an interval

rdubrock:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/87

Refetches metric data every 60 seconds.

### DIFF
--- a/src/hooks/useMetricData.ts
+++ b/src/hooks/useMetricData.ts
@@ -11,6 +11,15 @@ export function useMetricData(query: string, options?: (MetricQueryOptions & Use
   const [isFetchingData, setIsFetchingData] = useState(true);
   const [data, setData] = useState<any[]>([]);
   const [error, setError] = useState<string | undefined>();
+  const [refetches, setRefetches] = useState(0);
+
+  // refresh data every 60 seconds
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setRefetches(refetches + 1);
+    }, 60000);
+    return () => clearInterval(interval);
+  });
 
   useEffect(() => {
     const getData = async () => {
@@ -27,7 +36,7 @@ export function useMetricData(query: string, options?: (MetricQueryOptions & Use
     if (!options?.skip) {
       getData();
     }
-  }, [query, instance?.api, options]);
+  }, [query, instance?.api, options, refetches]);
 
   return { loading: isFetchingData, data, error };
 }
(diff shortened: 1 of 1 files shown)


src/hooks/useMetricData.ts
```diff
@@ -11,6 +11,15 @@ export function useMetricData(query: string, options?: (MetricQueryOptions & Use
...
   const [error, setError] = useState<string | undefined>();
+  const [refetches, setRefetches] = useState(0);
+
+  // refresh data every 60 seconds
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setRefetches(refetches + 1);
+    }, 60000);
+    return () => clearInterval(interval);
+  });
```
leventebalogh, line 22: **Idea**
Maybe it could be done using a single `useEffect()` hook?

```javascript
useEffect(() => {
    const getData = async () => {
      if (options?.skip) {
        return;
      }

      setIsFetchingData(true);
      const url = instance?.api?.getMetricsDS()?.url;

      if (!url) {
        return;
      }

      const { error: queryError, data: queryData } = await queryMetric(url, query, options);

      setData(queryData);
      setError(queryError);
      setIsFetchingData(false);
    };
   

    const sixtySec = 60000;
    const interval = setInterval(getData, sixtySec);

    return () => clearInterval(interval);
}, [query, instance?.api, options]);
```

leventebalogh (approved):
Left an idea, but generally LGTM 👍

src/hooks/useMetricData.ts
s-h-a-d-o-w: I feel more strongly about not using a seperate `useEffect()` just to force refetches than you do, @leventebalogh . ;)

There is a bug in the suggestion though (`setIsFetchingData(true)` happens too early) and overall, I'd tweak it a bit like so:
```
useEffect(() => {
  const interval = setInterval(async () => {
    const url = instance?.api?.getMetricsDS()?.url;
    if (options?.skip || !url) {
      return;
    }

    setIsFetchingData(true);

    const { error: queryError, data: queryData } = await queryMetric(url, query, options);

    setData(queryData);
    setError(queryError);
    setIsFetchingData(false);
  }, refetchInterval);

  return () => clearInterval(interval);
}, [query, instance?.api, options]);
```

... with `refetchInterval` either being declared at the top of the module or imported from a `consts` module.

(Sidenote - that `| undefined` in the `useMetricData` function signature seems unnecessary.)

leventebalogh: Good catch and nice tweaks! :)

s-h-a-d-o-w: Thanks!

rdubrock: Thanks y'all!

PR #96, opened 2020-09-04
Probe options

rdubrock:
This PR rearranges the check editing experience:
 - Collapses are restyled to meet design spec
 - Nested collapses are removed and flattened out
 - Label inputs are moved to advanced options in every check type
 - Common fields for all check types are broken up into sections
 - Frequency and timeout fields are converted to a combination slider/input component

Before: 
Screen Shot 2020-09-08 at 10 42 54 AM

After:
Screen Shot 2020-09-08 at 10 16 52 AM

### DIFF
--- a/src/components/CheckEditor/CheckEditor.tsx
+++ b/src/components/CheckEditor/CheckEditor.tsx
@@ -10,17 +10,15 @@ import {
   Switch,
   Select,
   Legend,
-  Collapse,
   Alert,
 } from '@grafana/ui';
 import { SelectableValue } from '@grafana/data';
-import { Check, Label as SMLabel, Settings, CheckType, Probe, OrgRole, APIError } from 'types';
+import { Check, Label as SMLabel, CheckType, Probe, OrgRole, APIError } from 'types';
 import { SMDataSource } from 'datasource/DataSource';
 import { hasRole, checkType, defaultSettings } from 'utils';
-import SMLabelsForm from 'components/SMLabelsForm';
 import * as Validation from 'validation';
 import CheckTarget from 'components/CheckTarget';
-import CheckSettings from './CheckSettings';
+import CheckSettings, { OnUpdateArgs } from './CheckSettings';
 import { ProbeOptions, OnChangeArgs } from './ProbeOptions';
 
 interface Props {
@@ -86,9 +84,11 @@ export default class CheckEditor extends PureComponent<Props, State> {
     this.setState({ check });
   };
 
-  onSettingsUpdate = (settings: Settings) => {
+  onSettingsUpdate = ({ settings, labels }: OnUpdateArgs) => {
     let check = { ...this.state.check } as Check;
+
     check.settings = settings;
+    check.labels = labels ?? [];
     this.setState({ check });
   };
 
@@ -245,30 +245,13 @@ export default class CheckEditor extends PureComponent<Props, State> {
             probes={check.probes}
             onChange={this.onProbeOptionsChange}
           />
-          <Collapse label="Options" collapsible={true} onToggle={this.onToggleOptions} isOpen={showOptions}>
-            <Field
-              label="Labels"
-              description="Custom labels to be included with collected metrics and logs."
-              disabled={!isEditor}
-              invalid={!Validation.validateLabels(check.labels)}
-            >
-              <SMLabelsForm
-                labels={check.labels}
-                onUpdate={this.onLabelsUpdate}
-                isEditor={isEditor}
-                type="Label"
-                limit={5}
-              />
-            </Field>
-            <br />
-            <h3 className="page-heading">{typeOfCheck!.toLocaleUpperCase()} Settings</h3>
-            <CheckSettings
-              settings={check.settings}
-              typeOfCheck={typeOfCheck || CheckType.HTTP}
-              onUpdate={this.onSettingsUpdate}
-              isEditor={isEditor}
-            />
-          </Collapse>
+          <CheckSettings
+            labels={check.labels}
+            settings={check.settings}
+            typeOfCheck={typeOfCheck || CheckType.HTTP}
+            onUpdate={this.onSettingsUpdate}
+            isEditor={isEditor}
+          />
         </div>
         <HorizontalGroup>
           <Button onClick={this.onSave} disabled={!isEditor || !Validation.validateCheck(check)}>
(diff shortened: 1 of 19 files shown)


teddyba:
@rdubrock This looks incredible! a few notes:

- Can we add a few more pixels of padding before the action buttons (save, back)?
- For the Timeout selector, the UI text should read "After [ ] seconds" instead of "Every [ ] seconds" 
- Let's figure out what a better title for "Probe Options" would be. Maybe something like "Monitoring Options" or "Request Settings"

src/components/CheckEditor/CheckSettings.tsx
```diff
@@ -41,29 +48,52 @@ export default class CheckSettings extends PureComponent<Props, State> {
...
       case CheckType.TCP: {
-        return <TcpSettingsForm settings={settings} onUpdate={this.onSettingsChange} isEditor={isEditor} />;
+        return (
+          <TcpSettingsForm
+            settings={settings}
+            labels={labels ?? []}
+            onUpdate={this.onSettingsChange}
+            isEditor={isEditor}
+          />
+        );
```
domasx2, line 96: Not a blocker or anything, but interesting idea about TS:
There's a way to make a `switch` statement or a series of `if`s exhaustive, causing compiler to throw an error if not all  possible values of an enum are covered:
https://dev.to/babak/exhaustive-type-checking-with-typescript-4l3f

I like to use it, to make changing enum values safer in the future

src/components/DnsSettings.tsx
```diff
@@ -37,80 +49,124 @@ function getUpdatedSettings(settings: DnsSettings, action: Action): DnsSettings
...
   const [showValidation, setShowValidation] = useState(false);
   const [showAdvanced, setShowAdvanced] = useState(false);
+  const [showDNSSettings, setShowDNSSettings] = useState(false);
+  const onValidateAnswerChange = useCallback(
+    (validations: DNSRRValidator | undefined) => {
+      const dns = getUpdatedSettings(values, { name: 'validateAnswerRRS', value: validations });
+      onUpdate({ dns }, labels);
+    },
+    // eslint-disable-next-line
+    []
```
domasx2, line 70: going out of your way here to not specify `onUpdate` as dependency :)

domasx2 (approved):

src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -119,6 +120,18 @@ export default class CheckEditor extends PureComponent<Props, State> {
...
+    const { check } = this.state;
+    this.setState({
+      check: {
+        ...check,
+        timeout,
+        frequency,
+        probes,
+      },
+    });
+  };
```
s-h-a-d-o-w, line 133: Using the current state value like this can be dangerous because it may not be up to date. And so whenever using the current state value, it should be done through an [updater function](https://reactjs.org/docs/react-component.html#setstate).
(Also e.g. `onSettingsUpdate` above. I'll leave it up to you to see whether there are even more cases. ;) )

src/components/CheckEditor/CheckProbes.tsx
```diff
@@ -81,16 +82,20 @@ export default class CheckProbes extends PureComponent<CheckProbesProps, CheckPr
           invalid={!Validation.validateProbes(probes)}
           closeMenuOnSelect={false}
         />
-        <Container margin="xs">
-          <HorizontalGroup spacing="md">
+        <div
+          className={css`
+            margin-top: 0.5rem;
```
s-h-a-d-o-w, line 87: I'm not sure about this being specified in `rem`, considering that we're using a 8px grid. But maybe I'm lacking the context for why it's necessary.

src/components/Collapse.tsx
```diff
@@ -0,0 +1,32 @@
...
+    border-right: none;
+    border-left: none;
+    border-bottom: none;
+    margin-bottom: 0;
+  }
+  .panel-container > div:first-of-type {
+    padding: 16px 0px;
+  }
+  .panel-container > div:nth-child(2) {
+    padding: 0 8px;
```
s-h-a-d-o-w, line 24: Might be worthwhile considering using the grafana theme for various hard-coded pixel values. It has design tokens for e.g. spacing and typography.

src/components/DnsSettings.tsx
```diff
@@ -37,80 +49,124 @@ function getUpdatedSettings(settings: DnsSettings, action: Action): DnsSettings
...
-    ...settings.dns,
-  };
+const DnsSettingsForm: FC<Props> = ({ settings, isEditor, labels, onUpdate }) => {
+  const values = useMemo(
+    () => ({
+      ...defaultValues,
+      ...settings.dns,
+    }),
+    [settings.dns]
+  );
```
s-h-a-d-o-w, line 59: As far as I can see, you only use `dns` in this component, so you don't need to provide the full `settings` object.

src/components/DnsSettings.tsx
s-h-a-d-o-w: Also `values`.

Whenever one has to disable eslint, something fishy might be going on. Because unless `onUpdate` and `values` change frequently, this function wouldn't get recreated anyway. And if they to change, wouldn't you want these handlers to get updated?

src/components/CheckEditor/CheckEditor.tsx
rdubrock: Good call!

src/components/CheckEditor/CheckProbes.tsx
rdubrock: Just muscle memory I need to forget. `rem` tends to result in a more stable layout when users vary their font size, zoom, etc, but consistency is more important

s-h-a-d-o-w: I'm aware, [fructosedb](https://github.com/s-h-a-d-o-w/fructosedb) mostly uses `rem`. :)
Although I believe I checked out whether that's still true today and think that browsers have become very good at scaling pixel-based layouts appropriately. (I could be wrong about this though - if I was to start a fresh project right now, I'd double-check.)

src/components/DnsSettings.tsx
rdubrock: Good call! It's a consequence of the state management of the parent component. I want to come back and clean that up via a switch to `Form` in a future PR

PR #99, opened 2020-09-15
Check editor form

rdubrock:
This PR is a montrosity, my apologies ahead of time. The big changes:

- Moved the CheckEditor from a state pyramid to using `react-hook-form`, which is the library `@grafana/ui` uses. It provides a more consistent data flow, and validation handlers, etc. This has cascading consequences and changes for a large section of the codebase
- Changed the existing validations from returning a boolean to either an error message or undefined so that we can give feedback in the UI
- Ditched the `@grafana/ui` `<Collapse>` component in favor of a handrolled one that doesn't remove hidden elements from the page. The model of `react-hook-form` is such that if a component isn't mounted in the DOM, the values are blown away. Since our forms have a bunch of collapsing sections, I needed the collapse to hide things instead of removing them from the DOM completely.
- There is a different data model between what the api provides/expects and what is used to structure the form. All the translation between those two things was moved to `checkFormTranformations`. There is, for sure, some room for improvement there.
- Added tests (not exhaustive coverage, but they cover the trickiest bits), and moved some existing tests around. `react-hook-form` seems to fight RTL's asynchronous methods, so I had to wrap a bunch of things in `act`. I also had a really tough time finding any way to interact with `<Select>` in a testing environment, any advice on that very welcome.
- Refactored the DNS validation inputs to match the newer designs
- Refactored the labels and several of the validation settings inputs that have variable quantities of values to leverage the `useFieldArray` functionality


src/components/CheckEditor/CheckEditor.test.tsx
```diff
@@ -0,0 +1,344 @@
...
+  probes: [],
+  settings: {
+    ping: {
+      ipVersion: IpVersion.V4,
+      dontFragment: false,
+    },
+  },
+} as Check;
+
+const getMinimumCheck = (overrides: any = {}) => ({
```
domasx2, line 26: here comes the types police: `overrides: Partial<Check>` perhaps?

src/components/CheckEditor/CheckEditor.test.tsx
```diff
@@ -0,0 +1,344 @@
...
+  render(
+    <InstanceContext.Provider value={{ instance: { api: instance }, loading: false }}>
+      <CheckEditor check={check} instance={instance} onReturn={onReturn} />
+    </InstanceContext.Provider>
+  );
+  await waitFor(() => expect(screen.getByText('Check Details')).toBeInTheDocument());
+  return instance;
+};
+
+it('renders without crashing', async () => {
```
domasx2, line 75: haha! Kinda implicitly covered by any other test as well :)

besartberisha (commented):
Had some minor comments/suggestions, will try to check more later of it since I only managed to review around ~30% of the PR.

src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -12,207 +12,79 @@ import {
...
-  };
+  const formMethods = useForm<CheckFormValues>({ defaultValues, mode: 'onChange' });
+  const selectedCheckType = formMethods.watch('checkType').value as CheckType;
 
-  onTimeoutUpdate = (event: React.ChangeEvent<HTMLInputElement>) => {
-    let check = { ...this.state.check } as Check;
-    check.timeout = event.target.valueAsNumber * 1000;
-    this.setState({ check });
-  };
+  let isEditor = hasRole(OrgRole.EDITOR);
```
domasx2, line 47: `let`? :raised_eyebrow:

src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -12,207 +12,79 @@ import {
...
-      this.props.onReturn(true);
+      onReturn(true);
     } catch (e) {
-      this.setState({
-        error: {
-          status: e.status,
-          message: e.data?.message ?? 'Something went wrong',
-        },
-      });
+      setError(e);
```
domasx2, line 63: We've discussed this with @s-h-a-d-o-w  a bit,
using hooks after `await` in a component has potential problems:
- possible race condition if called multiple times
- if component is umounted in the meanwhile, hooks like `setError` will throw

Consider using https://github.com/slorber/react-async-hook  (`useAsyncCabllack` specifically) or similar for things like `onSubmit`?

src/components/CheckEditor/checkFormTransformations.ts
```diff
@@ -0,0 +1,339 @@
...
+  DnsValidationFormValue,
+  ResponseMatchType,
+  Check,
+} from 'types';
+
+import { CHECK_TYPE_OPTIONS, IP_OPTIONS, DNS_RESPONSE_CODES } from 'components/constants';
+import { checkType } from 'utils';
+
+export function selectableValueFrom<T>(value: T, label?: string): SelectableValue<T> {
+  const labelValue: unknown = value;
```
domasx2, line 30: Isn't it `T`, not `unknown`?

src/components/CheckEditor/checkFormTransformations.ts
```diff
@@ -0,0 +1,339 @@
...
+  return { label: label ?? (labelValue as string), value };
```
domasx2, line 31: perhaps `String(labelValue)` is safer here

src/components/CheckEditor/checkFormTransformations.ts
```diff
@@ -0,0 +1,339 @@
...
+    case CheckType.TCP: {
+      return {
+        tcp: {
+          ipVersion: IpVersion.V4,
+          tls: false,
+        },
+      };
+    }
+    default:
+      throw new Error(`Cannot find values for invalid check type ${t}`);
```
domasx2, line 73: I'm sure I've mentioned this before, but will plug it again. TS can be made to fail to compile if not all possible enum values are considered: https://dev.to/babak/exhaustive-type-checking-with-typescript-4l3f

I really like this clever type system trick, let me know if I should shut up about it :D

domasx2 (commented):
I only managed to go through half of it today :D will check tomorrow as well if not merged yet

src/components/CheckEditor/CheckProbes.tsx
```diff
@@ -55,7 +57,7 @@ export default class CheckProbes extends PureComponent<CheckProbesProps, CheckPr
 
   render() {
     const { probes } = this.state;
-    const { availableProbes, isEditor } = this.props;
+    const { availableProbes, isEditor, onBlur, invalid, error } = this.props;
```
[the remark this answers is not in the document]
rdubrock: No philisophical reason, just trying to keep the PR from getting bigger

src/components/CheckEditor/CheckEditor.test.tsx
rdubrock: ⌨️ 🚔

src/components/CheckEditor/CheckEditor.test.tsx
rdubrock: True! I had this as a starting spot while working, and just didn't delete it. You think I should get rid of it?

src/components/CheckEditor/checkFormTransformations.ts
rdubrock: 🎉 I was fighting the types here

src/components/CheckEditor/checkFormTransformations.ts
rdubrock: I was thinking about it! I was just juggling too many things in my limited brainspace. I'll add it in today

src/components/CheckEditor/CheckEditor.tsx
rdubrock: 😔

src/components/CheckEditor/checkFormTransformations.ts
rdubrock: It looks like the TS compiler is smart enough to handle this case, I'm guessing since the function has an explicitly typed return and is switching on an enum. If I try to incorporate the function with a `never` return type in the article I get unreachable code warnings:

```
switch(t) {
...all the cases in the enum
}
exhaustiveTypeCheck(t) <---- I get an unreachable code compiler warning here
```

Maybe I'm missing the point, though

src/components/CheckEditor/CheckEditor.tsx
rdubrock: Thanks for that, good to know!

src/components/CheckEditor/CheckEditor.test.tsx
domasx2: Getting rid would speed up tests by a second or two I guess! But doesn't really mater

src/components/CheckEditor/checkFormTransformations.ts
domasx2: You're right, if `default` is removed, you get the same effect because of the return type here. I guess the minor drawback is that it won't throw then if somehow an invalid value appears.

That is the point, to fail to compile if not all enum values are handled, and not have to remember every code location that needs an update when an enum gets modified.

I don't mind unreachable code warning when there's a legit purpose for it. Anyway, up to you how you wanna do it

src/components/CheckEditor/checkFormTransformations.ts
```diff
@@ -0,0 +1,339 @@
...
+
+const getTcpSettingsFormValues = (settings: Settings): TcpSettingsFormValues => {
+  const tcpSettings = settings.tcp ?? (fallbackSettings(CheckType.TCP) as TcpSettings);
+  return {
+    ...tcpSettings,
+    ipVersion: selectableValueFrom(tcpSettings.ipVersion),
+  };
+};
+
+interface GetDnsValidationArgs {
```
domasx2, line 114: A terser way to define this type that also automatically includes all possible values of `ResponseMatchType`:

`type GetDnsValidationArgs = { [key in ResponseMatchType]?: DNSRRValidator };`

src/components/NameValueInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+  label: string;
+  validateName?: (name: string) => string | undefined;
+  validateValue?: (value: string) => string | undefined;
+}
+
+export const NameValueInput: FC<Props> = ({ name, disabled, limit, label, validateName, validateValue }) => {
+  const { register, control, errors } = useFormContext();
+  const { fields, append, remove } = useFieldArray({ control, name });
+  const theme = useTheme();
+  const fieldError = name.split('.').reduce((error, current) => error?.[current], errors);
```
domasx2, line 19: clever!

src/types.ts
```diff
@@ -163,6 +225,8 @@ export interface Check extends BaseObject {
 
   // Link to probes
   probes: number[];
+  id?: number;
+  tenantId?: number;
 }
 
 export interface Settings {
```
domasx2, line 232: are these mutually exclusive, or is it possible for example of both `http` and `ping` properties to be on the `Settings` object?

domasx2 (commented):
`useHookForm` seems cool! The way it attaches to inputs, must be faster than `formik` for large forms

src/components/NameValueInput.tsx
s-h-a-d-o-w: Yeah but kind of a brain twister due to relying on implicit behavior and using `reduce` which doesn't really resemble what's actually going on. Which... I'm btw not completely sure about.

But it looks like you're probably not actually using all parts of `name` but instead rely on implicit discarding of irrelevant parts of it. And then use `reduce` to extract the error that you want if there is one.

So would this work as well?
```
const fieldError = errors?.[name.split('.').slice(-1)[0]];
```

s-h-a-d-o-w: Was just wondering whether the readability could be helped along by for instance renaming `error` to `nestedError` and `current` to `errorPathFragment`.

src/components/CheckEditor/checkFormTransformations.ts
rdubrock: Love it! I'll take all the TS tips you got

src/types.ts
rdubrock: They are mutually exclusive. There must be a better way to type this situation, right?

domasx2: I think there's also a way to make `ping`, `http`, etc properties exclusive, but type description is too gnarly to recommend

s-h-a-d-o-w: You don't need the additional `_type`. You can just check for a property that's exclusive to one of the types (so in this case `'http' in settings`) and TS will narrow it: https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDeBDAXFAzsATgSwDsBzAXwG4AoUSKAIQWQCMNt9jyrxpaB7YACwZwAPnUoUAZgFcCAY2B4eBKPwC2AChQ4iGXgICUyClCh4J6gOQoLp5VqKGkxk1FlLMPADYQAdJ55Emtr6lCYkUBCemNBOLq7uXr7+gfYhziQUzm4EHt5+AUEOlBlAA

rdubrock: Thanks y'all, I'll ruminate on a good way to proceed here

PR #114, opened 2020-09-24
fix: use new metric names

mem:
In https://github.com/grafana/synthetic-monitoring-agent/pull/73 there's
a change in the way that derived metrics are created. Specifically,
names like "probe_success_sum" change to "probe_all_success_sum" to
indicate that this metric contains information about _all_ the runs. The
_sum and _count metrics reflect the fact that they keep a running sum
and running count. We use those instead of simply probe_success in order
to avoid relying on the existence (or not) of the metric in order to
detect failures. The problem with that naming is that it will clash (in
the future) with the name of the original metric, probe_success in this
case. In order to avoid that, we rename the derived metric to
probe_all_success.

This change depends on that PR being merged _and_ deployed.


rdubrock (commented):
Looks good! Does the query at https://github.com/grafana/synthetic-monitoring-app/blob/b5b42165aa976ab91fd9c6826f24472da9c9963e/src/components/CheckHealth.tsx#L45 also need to change?

mem:
> Looks good! Does the query at
> 
> https://github.com/grafana/synthetic-monitoring-app/blob/b5b42165aa976ab91fd9c6826f24472da9c9963e/src/components/CheckHealth.tsx#L45
> 
> also need to change?
> Nevermind!

Thanks for looking at this.

For completeness' sake: anywhere where you see `foo_bar_bucket`, `foo_bar_count` or `foo_bar_sum` most likely needs changing _if there's another metric that is called `foo_bar`_. Those are metrics that are created when the client creates a histogram or a summary. Since the probes are creating these metrics without using the Prometheus client libraries, this is not flagged as an issue. From the point of view of Prometheus, there's a single metric `foo_bar` and the other ones are simply an implementation detail. What the probes do is that they see a metric called `probe_success` and create the other _series_. This is possible because the probe is a long running process, so it can build the distribution on the client side. The probe has been changed to that they create a metric `probe_all_success` that keeps the distribution, and from the point of view of Prometheus that's a different metric.

PR #123, opened 2020-09-30
Switches to checkboxes

### LINKED ISSUE #119: Switch toggles should be checkboxes in most cases

rdubrock:
fixes #119

### DIFF
--- a/src/components/CheckEditor/CheckEditor.test.tsx
+++ b/src/components/CheckEditor/CheckEditor.test.tsx
@@ -313,6 +313,18 @@ describe('HTTP', () => {
     await act(async () => await userEvent.paste(screen.getByLabelText('Client Key', { exact: false }), 'client key'));
     await toggleSection('TLS config');
 
+    // Authentication
+    const authentication = await toggleSection('Authentication');
+    userEvent.click(await within(authentication).findByLabelText('Include bearer authorization header in request'));
+    const bearerTokenInput = await screen.findByPlaceholderText('Bearer token');
+    await act(async () => await userEvent.type(bearerTokenInput, 'a bearer token'));
+
+    userEvent.click(await within(authentication).findByLabelText('Include basic authorization header in request'));
+    const usernameInput = await within(authentication).findByPlaceholderText('Username');
+    const passwordInput = await within(authentication).findByPlaceholderText('Password');
+    await act(async () => await userEvent.type(usernameInput, 'a username'));
+    await act(async () => await userEvent.type(passwordInput, 'a password'));
+
     // Validation
     const validationSection = await toggleSection('Validation');
     const [statusCodeInput, httpVersionInput] = await within(validationSection).findAllByRole('textbox');
@@ -371,6 +383,11 @@ describe('HTTP', () => {
           validHTTPVersions: ['HTTP/1.0'],
           failIfNotSSL: false,
           failIfSSL: false,
+          basicAuth: {
+            password: 'a password',
+            username: 'a username',
+          },
+          bearerToken: 'a bearer token',
           failIfHeaderNotMatchesRegexp: [{ regexp: 'a header regex', allowMissing: true, header: 'Content-Type' }],
           failIfHeaderMatchesRegexp: [],
           failIfBodyMatchesRegexp: [],
--- a/src/components/CheckEditor/CheckEditor.test.tsx
+++ b/src/components/CheckEditor/CheckEditor.test.tsx
@@ -240,7 +240,7 @@ describe('HTTP', () => {
     expect(await within(httpSection).findByPlaceholderText('value')).toHaveValue('headerValue');
 
     await toggleSection('TLS config');
-    expect(await screen.findByLabelText('Skip validation', { exact: false })).toBeChecked();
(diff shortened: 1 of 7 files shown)


src/components/CheckEditor/CheckEditor.test.tsx
```diff
@@ -311,6 +313,18 @@ describe('HTTP', () => {
...
+    const authentication = await toggleSection('Authentication');
+    userEvent.click(await within(authentication).findByLabelText('Include bearer authorization header in request'));
+    const bearerTokenInput = await screen.findByPlaceholderText('Bearer token');
+    await act(async () => await userEvent.type(bearerTokenInput, 'a bearer token'));
+
+    userEvent.click(await within(authentication).findByLabelText('Include basic authorization header in request'));
+    const usernameInput = await within(authentication).findByPlaceholderText('Username');
+    const passwordInput = await within(authentication).findByPlaceholderText('Password');
+    await act(async () => await userEvent.type(usernameInput, 'a username'));
+    await act(async () => await userEvent.type(passwordInput, 'a password'));
```
leventebalogh, line 326: Why exactly do we need the `await`s and the `act()` here?

src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -20,19 +9,43 @@ import { getDefaultValuesFromCheck, getCheckFromFormValues } from './checkFormTr
...
 
 interface Props {
   check?: Check;
   instance: SMDataSource;
   onReturn: (reload: boolean) => void;
 }
 
+const getStyles = (theme: GrafanaTheme) => ({
+  marginBotton: css`
+    margin-bottom: ${theme?.spacing.sm};
```
s-h-a-d-o-w, line 27: No need for `?` operator here.

src/components/HorizonalCheckboxField.tsx
```diff
@@ -0,0 +1,61 @@
...
+  id,
+  name,
+  label,
+  description,
+  className,
+  value,
+  onChange,
+}) => {
+  const styles = useStyles(getStyles);
+  const { register } = useFormContext();
```
leventebalogh, line 43: 👏

src/components/HorizonalCheckboxField.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+interface Props {
+  disabled?: boolean;
+  id?: string;
+  name?: string;
+  label?: string;
+  description?: string;
+  className?: string;
+  value?: boolean;
+  onChange?: () => void;
```
leventebalogh, line 29: Are we sure all these props should be optional?
I mean, in the way how we are going to use this component we most probably always want to specify an `id` and a `name` at least, right?

src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -71,29 +84,25 @@ export const CheckEditor: FC<Props> = ({ check, instance, onReturn }) => {
     <FormContext {...formMethods}>
       <form onSubmit={formMethods.handleSubmit(onSubmit)}>
         <Legend>{check?.id ? 'Edit Check' : 'Add Check'}</Legend>
-        <div
-          className={css`
-            margin-bottom: 8px;
-          `}
-        >
+        <div className={styles.marginBotton}>
```
s-h-a-d-o-w, line 87: Just an idea - With something this general, it might make sense to extract it to shared styles. (Or not use a utility-style name for the class :) )

src/components/HorizonalCheckboxField.tsx
```diff
@@ -0,0 +1,61 @@
...
+}) => {
+  const styles = useStyles(getStyles);
+  const { register } = useFormContext();
+
+  return (
+    <div className={cx(styles.container, className)}>
+      <Checkbox
+        name={name}
+        ref={name ? register : undefined}
+        disabled={disabled}
```
s-h-a-d-o-w, line 50: How is this used when there is no name?

src/components/CheckEditor/CheckEditor.test.tsx
rdubrock: It's because of react-hook-form. I think the validation functions are what do it, they are async by default, so any time you have input into a validated field React will throw `act` warnings in the tests. There might be additional reasons, but I'm not sure

src/components/HorizonalCheckboxField.tsx
rdubrock: Good question! The `name` prop sort of implies registering with a form, but that won't always be the case. The `id` prop should probably be required, though.

src/components/CheckEditor/CheckEditor.tsx
rdubrock: Its funny, when using the`const styles = useStyles(getStyles)` , the theme passed to `getStyles` by the hook is typed as optional

src/components/HorizonalCheckboxField.tsx
rdubrock: To do some local control of a form section, like showing or hiding a form field. So in that case the checkbox value isn't part of the expected form output, I just want to use it to toggle some additional fields to include

PR #124, opened 2020-10-02
Navigation: Make Home the default page

torkelo:
By adding defaultNav it makes clicking on the sidenav app icon go this page.  Noticed that nothing happened when I clicked the synthetic monitoring icon.


rdubrock (approved):
🎉

PR #127, opened 2020-10-09
fix: allow hostnames to start with a number

rdubrock:


mem:
Beware that the probes will still reject this, because this is an invalid host name. We should have made the change in the probes first, deploy that and then change the app. RFC 1123 _does allow_ for host names to start with numbers without restrictions, _as long as something like 1.2.3.4 is treated first like an IP address and only if that fails it's used as a hostname_.

PR #136, opened 2020-10-15
fix: auto update dashboards on UI load

rdubrock:
When the metrics underpinning dashboard queries in the SM dashboard templates change, it breaks the dashboards unless users navigate to `Plugin config` => click on `Update` => click `Update` for each dashboard that has changes. This pr would check for dashboard template changes and present a modal that prompts the user to update them when they visit the SM plugin UI.

Screen Shot 2020-10-19 at 10 44 26 AM

### DIFF
--- a/src/RootPage.tsx
+++ b/src/RootPage.tsx
@@ -3,7 +3,7 @@ import React, { PureComponent } from 'react';
 
 // Types
 import { NavModelItem, AppRootProps, DataSourceInstanceSettings } from '@grafana/data';
-import { GlobalSettings, RegistrationInfo, GrafanaInstances } from './types';
+import { GlobalSettings, RegistrationInfo, GrafanaInstances, DashboardMeta } from './types';
 import { SMDataSource } from 'datasource/DataSource';
 import { findSMDataSources, createNewApiInstance, dashboardUID } from 'utils';
 import { SMOptions } from 'datasource/types';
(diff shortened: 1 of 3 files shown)


domasx2:
Don't have enough context to review this fully, sorry!
But maybe it's worth to ask user if he wants to update? Or not?

src/RootPage.tsx
```diff
@@ -76,6 +82,47 @@ export class RootPage extends PureComponent<Props, State> {
...
+        const templateVersion = templateDashboard?.latestVersion ?? -1;
+        if (templateDashboard && templateVersion > existingDashboard.version) {
+          return {
+            ...existingDashboard,
+            version: templateDashboard.latestVersion,
+            latestVersion: templateDashboard.latestVersion,
+          };
+        }
+        return null;
+      })
```
mem, line 102: this decides whether or not to upgrade the dashboard based on the version.

If the user edits the dashboard, Grafana will update the version.

My understanding is that this happens:

- application ships dashboard version 11
- user makes edits, grafana changes version to 12
- application ships dashboard version 12

-> no upgrade

This is the reason why the dashboard upgrade is user-initiated and not automatic.

Quickly scanning thru the dashboard's JSON, I cannot see anything that indicates that the dashboard has been updated.

Grafana does keep track of the edits, but I don't see anything that allows you to tell a manual edit apart from an edit "made" by the application. You could look at the RO/RW bit, but the user could set it back to RO.

Any ideas as to how to tell user edits apart from application updates?

rdubrock: So I don't believe it's reading the "grafana" version. It's comparing the version number that's in the json template in the plugin with the version number that's stored in the datasource. When a dashboard updates, the current template from the frontend is stored in the datasource. We never actually check grafana itself to see which version it has (that's how it's worked historically in the config page), which I think is fine because I don't believe grafana uses our version number.

PR #139, opened 2020-10-16
fix: update queries for metrics used in the UI

rdubrock:


src/components/CheckHealth.tsx
```diff
@@ -42,7 +42,7 @@ const getIconClassName = (error: string | undefined, noData: boolean, uptime: nu
 
 export const CheckHealth: FC<Props> = ({ check }) => {
   const filter = `instance="${check.target}", job="${check.job}"`;
-  const query = `sum(probe_success{${filter}}) / count(probe_success{${filter}})`;
+  const query = `sum((probe_success{${filter}} OR probe_all_success{${filter}})) / count((probe_success{${filter}} OR probe_all_success{${filter}}))`;
```
rdubrock, line 45: @mem I think this update maybe isn't necessary because I don't see `probe_success` as a metric that we changed, so I can revert this bit if that's the case

mem: right, that wasn't changed

mem (approved):

PR #141, opened 2020-10-23
Installed by default

rdubrock:
What does this PR do?

- It allows for the SM plugin to be initialized via provisioning
- We expect provisioning to happen via cloud, but to keep things working for on-prem users the original method of initialization via an admin key is preserved
- Refactors the client side routing behavior
- Adds a "config" tab to make plugin configuration more discoverable
- Adds a landing page for provisioned plugin users that haven't been initialized yet.
- Refactors datasource discovery/initialization


src/components/PluginTabs.tsx
```diff
@@ -0,0 +1,208 @@
...
+  });
+}
+
+function getNavModel(tabs: Tab[], path: string, activeTab: Tab, logoUrl: string) {
+  const children = tabs.map(tab => ({
+    text: tab.label,
+    id: tab.id,
+    active: tab.id === activeTab.id,
+    url: `${path}?page=${tab.id}`,
+  }));
```
leventebalogh, line 90: **Question:** did we leave the icons out on purpose?

src/components/PluginTabs.tsx
```diff
@@ -0,0 +1,208 @@
...
+    };
+  }
+  return null;
+}
+
+export const PluginTabs: FC<AppRootProps<GlobalSettings>> = ({ query, onNavChanged, path, meta }) => {
+  const { instance } = useContext(InstanceContext);
+  const [hasDismissedDashboardUpdate, setHasDismissedDashboardUpdate] = useState(
+    Boolean(window.sessionStorage.getItem('hasDismissedDashboardUpdate'))
+  );
```
leventebalogh, line 112: **Idea:** these client-side-persistence related logic could possibly live in an other layer.

In that case it would be easier to oversee in the future actually what data are we persisting on the client-side, and it would also be easier to change where we would like to persist them and how.

src/components/ProbeEditor.tsx
```diff
@@ -41,14 +41,18 @@ const ProbeEditor: FC<Props> = ({ probe, onReturn }) => {
     formValues.latitude = Number(formValues.latitude);
     formValues.longitude = Number(formValues.longitude);
 
+    if (!instance.api) {
+      return;
+    }
```
leventebalogh, line 46: **Question:** don't we wanna log something in this case?

src/components/UnprovisionedSetup.tsx
```diff
@@ -0,0 +1,129 @@
...
+      setTenantInfo(tenantInfo);
+      setApiSetup(setupValues);
+    } catch (e) {
+      setApiSetupError(e.message);
+    }
+  };
+
+  const onInstanceSelectionSubmit = async (
+    metricsInstanceId: number | undefined,
+    logsInstanceId: number | undefined
```
leventebalogh, line 59: **Suggestion:** probably we could use optional args here?

By using `metricsInstanceId?: number` instead of `metricsInstanceId: number | undefined`.

src/components/UnprovisionedSetup.tsx
```diff
@@ -0,0 +1,129 @@
...
+        smDatasource,
+      });
+
+      await updatePluginJsonData(pluginId, {
+        logs: {
+          grafanaName: `grafanacloud-${hostedLogs.name}`,
+          hostedId: hostedLogs.id,
+        },
+        metrics: {
+          grafanaName: `grafanacloud-${hostedMetrics.name}`,
```
leventebalogh, line 97: **Question:** are these prefixes going to work in an on-prem setup as well?

src/components/UnprovisionedSetup.tsx
```diff
@@ -0,0 +1,129 @@
...
+      return;
+    }
+
+    if (!apiSetup || !apiSetup.apiHost || !tenantInfo?.accessToken) {
+      setInstanceSelectionError('Something went wrong setting up Synthetic Monitoring. Please refresh the page.');
+      return;
+    }
+
+    const hostedMetrics = tenantInfo?.instances.find(tenantInstance => tenantInstance.id === metricsInstanceId);
+    const hostedLogs = tenantInfo?.instances.find(tenantInstance => tenantInstance.id === logsInstanceId);
```
leventebalogh, line 72: **Nitpick**

The following is probably shorter to write and scan:
```javascript
tenantInfo?.instances.find(({ id })  => id === logsInstanceId);
```

src/datasource/DataSource.ts
```diff
@@ -128,7 +128,10 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
         url: `${this.instanceSettings.url}/sm/check/list`,
       })
       .then((res: any) => {
-        return res.data;
+        if (Array.isArray(res.data)) {
+          return res.data;
+        }
+        return [];
```
leventebalogh, line 134: **Nitpick**

Maybe a ternary is easier to read in this case:
```javascript
.then((res: any) => Array.isArray(res.data) ? res.data : [])
```

src/fetchers.ts
```diff
@@ -0,0 +1,141 @@
...
+export const updatePluginJsonData = async (pluginId: string, jsonData: any) => {
+  const pluginSettings = await fetchPluginSettings(pluginId);
+  await getBackendSrv().post(`api/plugins/${pluginId}/settings`, {
+    ...pluginSettings,
+    jsonData: {
+      ...(pluginSettings.jsonData ?? {}),
+      ...jsonData,
+    },
+  });
+};
```
leventebalogh, line 19: **Nitpick**
Hmm... maybe it's just me, but I probably wouldn't look for POST / PUT / DELETE methods under a module called `fetchers`.

leventebalogh:
I need to remember not to leave more than ~9-10 reviews per PR as Github collapses them 😄

README.md
```diff
@@ -1,37 +1,59 @@
...
+apps:
+  - type: grafana-synthetic-monitoring-app
+    name: grafana-synthetic-monitoring-app
+    disabled: false
+    jsonData:
+      apiHost: https://synthetic-monitoring-api.grafana.net
+      grafanaInstanceId: <instanceId>
+      logs:
+        grafanaName: <Name of a Loki datasource pointed to a Grafana Cloud Loki instance>
+        hostedId: <Grafana Cloud Loki instance ID>
```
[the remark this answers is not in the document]
rdubrock: Good question, I couldn't find it, but it's possible I was looking in the wrong place. This is the id of the actual instance in hosted grafana, it's stored in the datasource under the 'username' field of the datasource, I'm not sure if frontend code has access to that or not

src/components/InstanceProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+  metricInstanceName?: string;
+  logsInstanceName?: string;
+  meta: AppPluginMeta<GlobalSettings>;
+}
+
+export const InstanceProvider: FC<Props> = ({ children, metricInstanceName, logsInstanceName, meta }) => {
+  const [instances, setInstances] = useState<GrafanaInstances | null>(null);
+  const [instancesLoading, setInstancesLoading] = useState(true);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
rdubrock: Good call, I've actually got that in the repo, I should use it.

src/components/InstanceSelection.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { FC, useState } from 'react';
+import { Alert, Button } from '@grafana/ui';
+import { HostedInstance } from 'types';
+import { InstanceList } from 'components/InstanceList';
+
+interface Props {
+  logsInstances: HostedInstance[];
+  metricsInstances: HostedInstance[];
+  onSubmit: (metricsId: number | undefined, logsId: number | undefined) => void;
```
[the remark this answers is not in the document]
rdubrock: It doesn't! But you're right using the `?` is better

src/components/PluginTabs.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const [activeTab, setActiveTab] = useState(findActiveTab(tabs, query.page, apiInitialized));
+  const logoUrl = meta.info.logos.large;
+
+  function skipDashboardUpdate() {
+    window.sessionStorage.setItem('hasDismissedDashboardUpdate', 'true');
+    setHasDismissedDashboardUpdate(true);
+  }
+
+  // Prompt user to update dashboards that are out of date
+  useEffect(() => {
```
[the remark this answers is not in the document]
rdubrock: Good thought. I went with this pattern because useMemo doesn't seem to play nice with asynchronous functions, and I need to make a network call

src/components/InstanceProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { SMDataSource } from 'datasource/DataSource';
+import { Spinner } from '@grafana/ui';
+import { AppPluginMeta } from '@grafana/data';
+import { UnprovisionedSetup } from 'components/UnprovisionedSetup';
+
+async function fetchDatasources(metricInstanceName: string, logsInstanceName: string): Promise<GrafanaInstances> {
+  const dataSourceSrv = getDataSourceSrv();
+  const smApi = (await dataSourceSrv.get('Synthetic Monitoring').catch(e => undefined)) as SMDataSource | undefined;
+  const metrics = await dataSourceSrv.get(metricInstanceName).catch(e => undefined);
+  const logs = await dataSourceSrv.get(logsInstanceName).catch(e => undefined);
```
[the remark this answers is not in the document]
rdubrock: When a user logs in for the first time (especially when running on prem) it's possible/likely that none of the datasources the plugin depends on exist. So this is supposed to be 'get the datasources if they exist, if they don't we're going to have to kick the user into a different flow' instead of guaranteeing their existence

src/components/InstanceProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+      });
+    } else {
+      setInstances({});
+      setInstancesLoading(false);
+    }
+  }, [logsInstanceName, metricInstanceName]);
+
+  if (instancesLoading || !instances) {
+    return <Spinner />;
+  }
```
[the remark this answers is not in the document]
rdubrock: I think I chucked the `!instances` in there just to satisfy the ts compiler. I think you're correct, that should be a seperate error case.

src/components/InstanceProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+  if (!instances.metrics || !instances.logs) {
+    return <UnprovisionedSetup pluginId={meta.id} pluginName={meta.name} />;
+  }
+  return (
+    <InstanceContext.Provider value={{ meta, instance: instances, loading: instancesLoading }}>
```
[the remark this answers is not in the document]
rdubrock: Totally, this is just an artifact of the way things used to be, and would require changes across most of the app. I want to follow up on this in a seperate PR since this one is already pretty big

src/components/PluginTabs.tsx
rdubrock: The app hasn't ever had any icons in the tabs, not sure if that was a conscious decision or not. @teddyba maybe you have some thoughts on icons in the tabs?

src/components/UnprovisionedSetup.tsx
rdubrock: Yeah, it's a little funny because you can run the plugin on prem, but you _have_ to send your data to a cloud instance. So the plugin has to connect to datasources that connect to an instance running in cloud. I name it this way just because that's how cloud names the datasources, so it will be consistent between the two (and make it clear to the user which instance is holding synthmon data)

PR #157, opened 2020-12-01
Series calculator

### LINKED ISSUE #153: Provide an active series estimate during check creation
We should give an indication of how many active series a check will generate when creating.

rdubrock:
What is this for?:
It provides customers with an approximation of how much usage they will be generating with a specific check. Closes #153 

series calculator


src/checkUsageCalc.ts
```diff
@@ -0,0 +1,63 @@
...
+  checksPerMonth: number;
+  activeSeries: number;
+  logsGbPerMonth: number;
+}
+
+enum CheckSeries {
+  PING = 76,
+  HTTP = 128,
+  DNS = 79,
+  TCP = 59,
```
leventebalogh, line 19: **Question:** probably a dumb question, but what do these numbers mean exactly?

src/checkUsageCalc.ts
```diff
@@ -0,0 +1,63 @@
...
+    case CheckType.TCP:
+      return CheckSeries.TCP;
+    case CheckType.DNS:
+      return CheckSeries.DNS;
+    case CheckType.HTTP:
+      return CheckSeries.HTTP;
+  }
+};
+
+const getMonthlyChecks = (frequencySeconds: number) => {
```
leventebalogh, line 35: **Nitpick:** I would probably call it `getChecksPerMonth()` instead as we are referring like that to its return value everywhere else.

leventebalogh (approved):
Left a few nitpick comments, but generally looks good to me.

src/checkUsageCalc.ts
nathanrodman: These numbers are the number of active series generated per check. Different types of checks generate different numbers of metrics

leventebalogh: Thanks for the explanation Nathan 👍

Issue #164, opened 2020-12-04
Check pagination isn't implemented


rdubrock:
In the list of checks we are showing pagination controls, but we are rendering all the checks without paginating them. We need to implement the pagination behavior.

PR #165, opened 2020-12-10
Paginate checks

nathanrodman:
Fixes issue #164 

The pagination component was in use but the check list wasn't actually paginating the results.


leventebalogh (approved):
LGTM 👍

PR #169, opened 2020-12-28
docs: update provisioning flow, remove api key flow

### LINKED ISSUE #168: Docs: update to show when provisioning is required
We run into a case where installing via provisioning is required, and should be called out in the docs. Use case:

- I log into grafana cloud, and initialize the SM tenant from the grafana instance in cloud (provisioned flow)
- I decide I want to connect an on prem grafana instance to the same SM tenant without provisioning (api key flow)
- If I install SM and try to initialized with an api key, I get the too many tenants error. Initializing via API key in this case is not allowed, the only way to connect to the tenant is via the provisioning flow

rdubrock:
fixes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,19 +17,7 @@ For each check, users can select 1 or more 'public' probe locations distributed 
 
 ### Configuration
 
-Synthetic Monitoring requires a Grafana Cloud account, but can be set up to run in a local Grafana instance in two ways:
-
-#### Via an admin key
-
-Install the [Synthetic Monitoring plugin](https://grafana.com/grafana/plugins/grafana-synthetic-monitoring-app)
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,4 +58,4 @@ During the Initialization process, the Synthetic Monitoring backend will:
 3. The dashboards included with the App will then be imported.
 4. Finally, the Synthentic Monitoring Backend will be notified that the service is ready.
 
-Users can then create checks to monitor their remote targets.
+Users can then create checks to monitor their remote targets for the selected Cloud stack.
```
electron0zero, line 61: > their remote targets for the selected Cloud stack

this bit is confusing my brain, maybe we can reword it to be bit more clear

electron0zero (approved):
small nit, rest LGTM :+1: :shipit:

PR #171, opened 2020-12-29
fix: annotation and labels in alerts not getting sent to ruler

rdubrock:
We have a bug with creating alerts. We didn't quite map the labels and annotations correctly when we moved to adding multiple alert rules.


src/components/CheckEditor/CheckEditor.test.tsx
```diff
@@ -647,6 +665,18 @@ describe('Alerting', () => {
...
+        labels: [
+          {
+            name: 'steve',
+            value: 'mcsteveson',
+          },
+        ],
+        annotations: [
+          {
+            name: 'bob',
+            value: 'mcbobson',
```
nathanrodman, line 677: Using this as my next alias

nathanrodman (commented):
nice catch. LGTM

PR #177, opened 2021-01-14
fix: renavigate to checks page after deleting check

nathanrodman:
A fix for check deletion. When the user would delete a check it would open a modal and delete the check via an API but it wasn't closing the modal after deletion and navigating back to the checks page.


rdubrock (commented):
🎉

PR #179, opened 2021-01-15
Metric reduction form option

nathanrodman:
Adds a field to the form for choosing standard (basic) or advanced metrics. Still need to work out the wording. @teddyba the Checkbox component in Grafana UI doesn't allow for elements in the description field, just a simple string. Meaning we can't put an anchor tag linking to the docs page. I can see about opening a PR to make a change but we also might be able to come up with a different solution.

Screenshot_2021-01-15 Synthetic Monitoring Synthetic Monitoring - Grafana


rdubrock (approved):
looks great 🦧

PR #181, opened 2021-01-22
Alerting v2

rdubrock:
Adds an alerting tab to the plugin, which can create default alert rules and edit the generated rules. 

Screen Shot 2021-01-22 at 11 27 53 AM

Screen Shot 2021-01-21 at 3 21 59 PM

Screen Shot 2021-01-22 at 11 26 56 AM
Screen Shot 2021-01-22 at 11 26 22 AM

### DIFF
--- a/src/components/AlertRuleForm.tsx
+++ b/src/components/AlertRuleForm.tsx
@@ -132,10 +132,15 @@ export const AlertRuleForm: FC<Props> = ({ rule, onSubmit }) => {
   const formMethods = useForm<AlertFormValues>({
     defaultValues,
   });
-  const { register, control, handleSubmit, errors, watch } = formMethods;
+  const { register, control, handleSubmit, errors, watch, reset } = formMethods;
   const currentValues = watch() as AlertFormValues;
   const currentLabels = watch('labels');
   const currentAnnotations = watch('annotations');
+
+  const onCancel = () => {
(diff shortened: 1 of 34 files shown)


rdubrock:
TBD: 
- Test styles with light mode
- Second pass at field validations in the alert rule form
- Update the alert expression to represent a percentage instead of an absolute probe success number

src/components/AlertRuleForm.tsx
```diff
@@ -0,0 +1,312 @@
...
+      setIsOpen(false);
+      return Promise.resolve();
+    } catch (e) {
+      return Promise.reject(e.message ?? 'Something went wrong');
+    }
+  });
+
+  if (!defaultValues) {
+    return (
+      <Collapse label={rule.alert} isOpen={isOpen} onToggle={() => setIsOpen(!isOpen)} collapsible>
```
nathanrodman, line 179: We're no longer using a <Collapse/> for the form, instead replacing it with a button. This should match

rdubrock: Yes! Thanks a bunch

nathanrodman (approved):
LGTM :shipit: 

Great work, Russ!

rdubrock:
Thanks a bunch for your help!

PR #187, opened 2021-01-29
fix: update tcp query response to send and parse base64

### LINKED ISSUE #186: TCP Query/Response field is failing on valid values
Adding a "query/response" option to a TCP check shouws a "server failed, 400" message.

Response from the API is:
```
{
"msg": "failed to decode updated check",
"err": "illegal base64 data at input byte 4"
}
```

rdubrock:
Resolves #186 

The api is expecting a base64 encoded byte value, and we are sending a plain string. This PR translates user input to base64 if it isn't base64 already, and translates it back when editing.

There is also a little fix for ca cert fields that were passing the entire errors object to field errors instead of the just the error message, which makes the FieldError component error out.

### DIFF
--- a/src/components/CheckEditor/checkFormTransformations.ts
+++ b/src/components/CheckEditor/checkFormTransformations.ts
@@ -26,6 +26,7 @@ import {
   HttpRegexValidationType,
   HeaderMatch,
   DnsResponseCodes,
+  TCPQueryResponse,
 } from 'types';
 
 import {
@@ -37,6 +38,7 @@ import {
   fallbackCheck,
 } from 'components/constants';
 import { checkType } from 'utils';
+import isBase64 from 'is-base64';
 
 export function selectableValueFrom<T>(value: T, label?: string): SelectableValue<T> {
   const labelValue = String(value);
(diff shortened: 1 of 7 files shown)


src/components/CheckEditor/checkFormTransformations.ts
```diff
@@ -183,11 +185,26 @@ const getHttpSettingsFormValues = (settings: Settings): HttpSettingsFormValues =
...
 
+const getTcpQueryResponseFormValues = (queryResponses?: TCPQueryResponse[]) => {
+  if (!queryResponses) {
+    return undefined;
+  }
+  return queryResponses.map(({ send, expect, startTLS }) => {
+    return {
+      startTLS,
+      send: atob(send),
+      expect: atob(expect),
```
nathanrodman, line 196: can both `btoa()` and `atob()` methods be in a try/catch block? If there is an invalid character it throws an exception

rdubrock: Good call!

nathanrodman (approved):
LGTM 🚢

PR #195, opened 2021-02-09
Frequency control fix

### LINKED ISSUE #192: Check cannot be created if "frequency" control is touched
Yo!

I'm trying to set up checks for a personal project and ran into an issue: if I touch "frequency" controls at all, form can no longer be saved. Seems there's a bunch of errors from slider component and then 400 from backend. Seems like `frequency: null` is being sent to backend here. I have to refresh the page and try again w/o touching "frequency" to be able to save a check.

Grafana v7.4.0 (c2203b9859)

Also: no idea what "60 every 60 seconds" means. 60 checks every minute? Would like to have one check per minute

synt monitoring err

rdubrock:
Fixes #192 

Several things that changed in 7.4 were breaking changes, most notably the <Slider> component. This PR catches up on the changes, including a bunch of type changes. The `Select` component started behaving differently as well, so I just mocked it with a bare HTML select in order to make the pages in general more testable.


nathanrodman (approved):
LGTM 🚢

PR #206, opened 2021-02-17
fix: add plugin enable action

rdubrock:
When I updated the config page, I mistakenly believed it would no longer be possible for users (both on-prem and in cloud) to have the plugin installed but disabled, and removed the enable function. This PR does a few things:

- Adds and "enable" action to the plugin config page
- Disables the plugin in Grafana on disable (instead of just with the SM API)
- Pins the plugin on enable

### DIFF
--- a/src/components/ConfigActions.test.tsx
+++ b/src/components/ConfigActions.test.tsx
@@ -6,12 +6,12 @@ import { getInstanceMock } from 'datasource/__mocks__/DataSource';
 import { AppPluginMeta } from '@grafana/data';
 import { GlobalSettings } from 'types';
 
-const renderConfigActions = ({ hasApi = true } = {}) => {
+const renderConfigActions = ({ hasApi = true, enabled = true } = {}) => {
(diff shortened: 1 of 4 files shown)


src/components/ConfigActions.test.tsx
```diff
@@ -6,12 +6,12 @@ import { getInstanceMock } from 'datasource/__mocks__/DataSource';
...
 import { GlobalSettings } from 'types';
 
-const renderConfigActions = ({ hasApi = true } = {}) => {
+const renderConfigActions = ({ hasApi = true, enabled = true } = {}) => {
   const instance = hasApi ? getInstanceMock() : undefined;
   const meta = {} as AppPluginMeta<GlobalSettings>;
   return render(
     <InstanceContext.Provider value={{ instance: { api: instance }, loading: false, meta }}>
-      <ConfigActions />
+      <ConfigActions enabled={enabled} pluginId="steve" />
```
nathanrodman, line 14: Good ol' Steve

nathanrodman (approved):

PR #212, opened 2021-02-18
chore: expose errors when fetching alerts

rdubrock:
Current errors fetching alert rules result in an endless spinner. This PR surfaces the error in the UI and stops the spinning.

### DIFF
--- a/src/hooks/useAlerts.ts
+++ b/src/hooks/useAlerts.ts
@@ -49,28 +49,31 @@ export const defaultRules = {
   ],
 };
 
-const fetchSMRules = async (alertRulerUrl: string) => {
-  try {
-    return await getBackendSrv()
-      .fetch<any>({
-        method: 'GET',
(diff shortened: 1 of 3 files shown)


src/hooks/useAlerts.ts
```diff
@@ -49,28 +49,31 @@ export const defaultRules = {
...
+      },
+    })
+    .toPromise()
+    .then((response) => {
+      const alertGroup = parse(response.data);
+      return { rules: alertGroup.rules };
+    })
+    .catch((e) => {
+      if (e.status === 404) {
+        return { rules: [] };
```
besartberisha, line 73: Not sure if I understand correctly, but based on this, the `alertRules` will always have an empty array, meaning in the `Alerting.tsx` component `{!alertRules && <Spinner />}` will spin even if an error is thrown.  

Probably a simple fix would be ` {!!alertRules.length && <Spinner />}`?

rdubrock: Good catch!

PR #214, opened 2021-02-25
Check card redesign

rdubrock:
Screen Shot 2021-02-26 at 12 27 41 PM
Screen Shot 2021-02-26 at 12 27 52 PM
Screen Shot 2021-02-26 at 12 27 31 PM


src/components/LabelFilterInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const LabelFilterInput = ({ checks, labelFilters, onChange, className }: Props) => {
+  const styles = useStyles(getStyles);
+  const aggregatedLabels = useMemo(
+    () =>
+      checks.reduce<AggregateLabels>((acc, check) => {
+        check.labels?.forEach(({ name, value }) => {
+          if (acc[name]) {
+            acc[name].push(value);
+          }
+          acc[name] = [value];
```
domasx2, line 39: it seems like this line should be inside an `else` branch

src/components/CheckCardLabel.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const getStyles = (theme: GrafanaTheme) => ({
+  container: css`
+    background-color: #9933cc;
+    border-radius: 1px;
+  `,
+});
+
+export const CheckCardLabel = ({ label, onLabelSelect }: Props) => {
```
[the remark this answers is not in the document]
rdubrock: The new eslint config in 7.4 gets grouchy when I use `FC<Props>`. It trips the `react/prop-types` rule. Did you bump into that in your repos?

src/components/LabelFilterInput.tsx
rdubrock: Yes! Good catch!

PR #223, opened 2021-03-05
fix: increase job name length limit

### LINKED ISSUE #75: Check job name length limit is restrictive
The current length limit for a check job name is 32 characters, which feels restrictive.

rdubrock:
Fixes #75 

In addition to increasing the job name limit, this PR adds some text truncation and a view dashboard button in the list view.

### DIFF
--- a/src/components/CheckListItem/CheckItemActionButtons.tsx
+++ b/src/components/CheckListItem/CheckItemActionButtons.tsx
@@ -2,7 +2,7 @@ import { GrafanaTheme, AppEvents } from '@grafana/data';
 import { Button, ButtonGroup, ConfirmModal, IconButton, useStyles } from '@grafana/ui';
 import React, { useContext, useState } from 'react';
 import { css } from 'emotion';
-import { Check, CheckType, OrgRole } from 'types';
+import { Check, OrgRole } from 'types';
 import { dashboardUID, checkType as getCheckType, hasRole } from 'utils';
 import { InstanceContext } from 'components/InstanceContext';
 import appEvents from 'grafana/app/core/app_events';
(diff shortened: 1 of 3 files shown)


src/components/CheckListItem/CheckItemActionButtons.tsx
```diff
@@ -47,8 +47,10 @@ export const CheckItemActionButtons = ({ check, showViewDashboard, onRemoveCheck
 
   return (
     <ButtonGroup className={styles.actionButtonGroup}>
-      {showViewDashboard && (
-        <Button onClick={() => showDashboard(check, checkType)} size="sm" variant="link">
+      {viewDashboardAsIcon ? (
+        <IconButton name="apps" onClick={showDashboard} />
```
samjewell, line 51: Is this attr `name="apps"` actually used anywhere? And should it be `name="dashboard"` or something instead?
I'm still learning my way around, so I might be looking in the wrong place, but I was expecting to find this `name` attr or something similar rendered directly into the DOM, onto this element.

src/components/CheckListItem/CheckItemActionButtons.tsx
```diff
@@ -17,17 +17,17 @@ const getStyles = (theme: GrafanaTheme) => ({
...
 
-export const CheckItemActionButtons = ({ check, showViewDashboard, onRemoveCheck }: Props) => {
+export const CheckItemActionButtons = ({ check, viewDashboardAsIcon, onRemoveCheck }: Props) => {
   const styles = useStyles(getStyles);
   const [showDeleteModal, setShowDeleteModal] = useState(false);
   const checkType = getCheckType(check.settings);
   const { instance } = useContext(InstanceContext);
 
-  const showDashboard = (check: Check, checkType: CheckType) => {
+  const showDashboard = () => {
```
samjewell, line 30: Nice change here ❤️

samjewell (commented):
This looks nice to me 👍 , great change Russ. I've just asked one question in the comments.
Quick question: How do you switch an org from the old interface to the new one, for the checks page? Thanks 🙏

PR #241, opened 2021-03-22
fix: add length validation for label names and values, truncate long …

### LINKED ISSUE #239: long values as a label value causes vague error
A check with additional labels (under Advanced), generates a "400 something went wrong" error when saved, rather than a better error message, or pop-up error when blurring the field. 

image

### LINKED ISSUE #240: Long label name results in an uneditable check
If you enter a long label name under the advanced section of a check, you then can't edit the check from the check list. 

image

rdubrock:
fixes #240 
fixes #239 

This PR does a few label related things:
- plumbs in the label name/value validation
~- Increases the validation value to 128 (from 32)~ Length validation increase will be included later
- Truncates arbitrarily long label values in the check list view
- Fixes a bug where deleting all the labels from a check wouldn't get persisted to the API

### DIFF
--- a/src/validation.ts
+++ b/src/validation.ts
@@ -101,8 +101,8 @@ export function validateLabel(label: Label): string | undefined {
 }
 
 export function validateLabelName(name: string): string | undefined {
-  if (name.length > 32) {
-    return 'Label names must be 32 characters or less';
+  if (name.length > 128) {
+    return 'Label names must be 128 characters or less';
   }
   if (!labelRegex.test(name)) {
     return 'Invalid label name';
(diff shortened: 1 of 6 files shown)


src/validation.ts
```diff
@@ -101,8 +101,8 @@ export function validateLabel(label: Label): string | undefined {
 }
 
 export function validateLabelName(name: string): string | undefined {
-  if (name.length > 32) {
-    return 'Label names must be 32 characters or less';
+  if (name.length > 128) {
```
mem, line 104: if you can keep this at 32, we can merge the rest of the changes now

PR #264, opened 2021-04-30
chore: remove deprecated datasource method

rdubrock:
We were using the deprecated `datasourceRequest` method for all our datasource interactions. Changing to the future proofed `fetch`. The `fetch` method returns an observable, and in order to limit the scope of the change I just converted them to promises.

### DIFF
--- a/__mocks__/@grafana/runtime.ts
+++ b/__mocks__/@grafana/runtime.ts
@@ -1,7 +1,9 @@
 import * as runtime from '@grafana/runtime';
 
 export const getBackendSrv = () => ({
-  datasourceRequest: jest.fn().mockImplementation(() => ({ ok: true })),
+  fetch: {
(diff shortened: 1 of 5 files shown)


__mocks__/@grafana/runtime.ts
```diff
@@ -1,7 +1,9 @@
 import * as runtime from '@grafana/runtime';
 
 export const getBackendSrv = () => ({
-  datasourceRequest: jest.fn().mockImplementation(() => ({ ok: true })),
+  fetch: {
+    toPromise: () => jest.fn().mockImplementation(() => ({ ok: true })),
```
leventebalogh, line 5: **Nitpick:** maybe `jest.fn().mockResolvedValue({ ok: true })` as a bit of syntactic sugar ✨

leventebalogh (commented):
LGTM 👍

leventebalogh (approved):

PR #265, opened 2021-04-30
feat: add the ability to sort checks by success rate

### LINKED ISSUE #250: Additional sort options for checks
Currently the we appear to be able to sort checks based on name, and filter on name, endpoint, type, or labels. It would be nice to be able to filter or sort based on success rate. Seems small, but when managing hundreds of checks it can be time consuming trying to find a specific subset that may be at less than 100% success rate with out bouncing across multiple screens and running custom queries.

rdubrock:
Closes #250 

This PR adds the ability to sort checks by the result of the success rate query. It also adds some of the plumbing to do the same for probes, but didn't actually add the functionality (because we don't have any sorting for probes at the moment).

The changes are fairly extensive because we were previously fetching a success rate for each individual check at the moment it was rendered. This wouldn't allow for sorting by success because checks are paginated and we wouldn't have values for every check. 

This PR changes the success query to fetch all the series in the metric and mapping the returned values to a check instead of sending a query for each individual check. This allows us to fetch the results for _all_ the checks with one request and store them in context. We can then have the ability to sort by success.

Screen Shot 2021-04-30 at 12 03 01 PM

### DIFF
--- a/src/__mocks__/utils.ts
+++ b/src/__mocks__/utils.ts
@@ -9,7 +9,10 @@ const queryMetric = jest.fn().mockImplementation(() => {
   return Promise.resolve({
     data: [
       {
-        metric: {},
+        metric: {
+          instance: 'tacos',
+          job: 'burritos',
+        },
         value: [1598535155, '1'],
       },
     ],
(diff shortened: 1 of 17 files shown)


nathanrodman (approved):
LGTM - great work and thanks for adding more tests

PR #277, opened 2021-06-03
Feature: add "reachability" to dashboards

mem:
Add a "reachability" panel to all dashboards (and a reachability column
to the summary). This reflects the fact that some probes might fail to
reach a given target because of network errors between the probe and the
target, not associated to either of them.

Also redefine uptime a measure of how much time the target was reachable
by any probe vs unreachable from all probes.

Fixes #

### DIFF
--- a/src/dashboards/sm-dns.json
+++ b/src/dashboards/sm-dns.json
@@ -22,13 +22,13 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "7.0.3"
+      "version": "7.5.7"
     },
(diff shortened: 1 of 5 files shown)


mem (commented):
These changes are the result of making some changes with a current Grafana version (7.5.7) and exporting the resulting dashboard with the "sharing" setting enabled. Part of the reason why the change is so large is because it's switching the table plugin from the old one to the new one.

rdubrock (approved):
Let's hold off on merging this until the v8 changes go in

PR #280, opened 2021-06-03
Feature: add filtering by custom labels

mem:
This adds dropdown menus to the summary dashboard that allow the user to
select a set of three labels to use as filters for the dashboard.

Each filter is composed of _two_ dropdowns: the first one specifies the
label to use for filtering and the second one the corresponding value.

There's a fourth, hidden, dropdown that is used to compute the effective
value of the "check type" dropdown, after applying the filters, so that
the dashboard doesn't show rows with empty panels.

The defaults in the dashboard are such that all the checks are selected.

### DIFF
--- a/src/dashboards/sm-dns.json
+++ b/src/dashboards/sm-dns.json
@@ -22,13 +22,13 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "7.0.3"
+      "version": "7.5.7"
     },
     {
       "type": "panel",
       "id": "grafana-worldmap-panel",
       "name": "Worldmap Panel",
-      "version": "0.2.1"
+      "version": "0.3.2"
     },
     {
       "type": "panel",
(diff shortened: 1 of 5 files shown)


rdubrock:
The update modal isn't actually updating the dashboard when I run it locally, so I'll try and add a fix for that to go with this update

rdubrock (approved):
I think the problems I was seeing with updates were related to grafana versions. Let's merge this after v8

electron0zero:
Grafana v8 has been out for a while, let's merge and release this.

samjewell:
I wanted a visual on the change in this PR, so in the end I imported the dashboards into my own instance, [here](https://samjewell.grafana.net/d/smsummaryv2attempt2/sm-summary-v2-attempt-2?orgId=1)

Just in case this is useful to anyone else, here's a screenshot of the change:
image

samjewell:
> This adds dropdown menus to the summary dashboard that allow the user to
select a set of three labels to use as filters for the dashboard

Why three labels? Why not just add one or two to start with? This seems like a powerful feature, but also one which increases complexity. Might adding just one deliver the lion's share of the benefit here, and help us keep complexity down?

PR #291, opened 2021-06-09
Visualization view

rdubrock:
This PR adds a third view type to the checks page that gives an overview of all checks, regardless of how many there are.

- The color of the hexagon is informed by the same query/data the `Success Rate` viz is using in the normal check list.
- The view can be sorted/filtered by the same controls as the check list views 
- ~It's currently behind a feature flag~
- The data refreshes on a cadence

checkviz


rdubrock:
The git history got a little sideways on this one since it started as a branch of a branch

src/components/Autosizer.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const el = useRef<HTMLDivElement>(null);
+  const styles = useStyles2(getStyles);
+  const [size, setSize] = useState({ width: 0, height: 0 });
+
+  useEffect(() => {
+    const rect = el.current?.getBoundingClientRect();
+    if (rect) {
+      setSize({ width: rect.width, height: rect.height });
+    }
+  }, [el]);
```
bfmatei, line 36: Shouldn't this be `el.current`?

bfmatei (approved):
One small question, but otherwise looks pretty much fine to me.

electron0zero (approved):

src/components/Autosizer.tsx
rdubrock: I had thought so too! When I do that, though, I get this eslint warning:

```
React Hook useEffect has an unnecessary dependency: 'el.current'. Either exclude it or remove the dependency array. Mutable values like 'el.current' aren't valid dependencies because mutating them doesn't re-render the component.
```

mem (approved):

src/components/Autosizer.tsx
bfmatei: Usually I would destructure the `el` object, get the current and use it directly, but this is an absolute nitpick.

PR #313, opened 2021-06-14
chore: add some analytics to the setup flow

rdubrock:
Following the example from the onboarding plugin, checking to see if GA is present in the page and sending events. I deliberately only added a few events to start


mem:
GA being "google analytics"?

rdubrock:
> GA being "google analytics"?

Yup!

PR #314, opened 2021-06-15
feat: format dashboard log panel to include the probe name

### LINKED ISSUE #261: Dashboard Error Logs should include probe.
Minor UX issue; it would be helpful if the error logs in the included dashboard included slightly more information. The values are available in the datasource, but the current implementation of the logs panel or the table panel don't provide for view that includes the probe value with out a loss of other functionality

**Why is this needed**:
When reviewing logs in the UI it can be difficult to quickly identify a widespread or single probe failure.

rdubrock:
Screen Shot 2021-06-15 at 12 25 47 PM
closes #261

### DIFF
--- a/src/dashboards/sm-dns.json
+++ b/src/dashboards/sm-dns.json
@@ -84,11 +84,7 @@
     {
       "circleMaxSize": "10",
       "circleMinSize": "2",
-      "colors": [
-        "#37872D",
-        "#FA6400",
-        "#C4162A"
-      ],
+      "colors": ["#37872D", "#FA6400", "#C4162A"],
       "datasource": "${DS_SM_METRICS}",
       "decimals": 2,
       "description": "What's the error percentage for each probe that is observing the given target.",
@@ -194,17 +190,13 @@
       "options": {
         "colorMode": "value",
         "fieldOptions": {
-          "calcs": [
-            "lastNotNull"
-          ]
+          "calcs": ["lastNotNull"]
         },
         "graphMode": "none",
         "justifyMode": "auto",
         "orientation": "horizontal",
         "reduceOptions": {
-          "calcs": [
-            "mean"
-          ],
+          "calcs": ["mean"],
           "fields": "",
           "values": false
         },
@@ -276,17 +268,13 @@
       "options": {
         "colorMode": "value",
         "fieldOptions": {
-          "calcs": [
-            "lastNotNull"
-          ]
+          "calcs": ["lastNotNull"]
         },
         "graphMode": "none",
         "justifyMode": "auto",
         "orientation": "horizontal",
         "reduceOptions": {
-          "calcs": [
-            "mean"
-          ],
+          "calcs": ["mean"],
           "fields": "",
           "values": false
         },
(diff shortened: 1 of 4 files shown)


src/dashboards/sm-dns.json
```diff
@@ -84,11 +84,7 @@
     {
       "circleMaxSize": "10",
       "circleMinSize": "2",
-      "colors": [
-        "#37872D",
-        "#FA6400",
-        "#C4162A"
-      ],
+      "colors": ["#37872D", "#FA6400", "#C4162A"],
```
samjewell, line 87: [nit] Personally I'd split this commit into two commits, or even better, 2 PRs. That way the git-blame will be cleaner and easier to interact with in future, and code-review is easier too.

Why have you made this change, and is it in scope of the review? Of course I have no objections to this change, but I hope we won't see these lines of code flip-flopping back and forth between PRs.

src/dashboards/sm-dns.json
```diff
@@ -919,7 +897,7 @@
       },
       "targets": [
         {
-          "expr": "{probe=~\"$probe\",instance=\"$instance\", job=\"$job\", probe_success=\"0\"}",
+          "expr": "{probe=~\"$probe\", instance=\"$instance\", job=\"$job\", probe_success=\"0\"} | logfmt | line_format \"level={{.level}} target={{.target}} probe={{.probe}} msg={{.msg}}\"",
```
samjewell, line 900: [maybe nit]
I can see that this does what the Github issue requested. But I wonder if we lose some value here too. By using `line_format` here you only allow those four items in the log-line to be rendered, so you lose other content such as:
- duration_seconds `… msg="Check failed" duration_seconds=10.000953946`
- err `… msg="Timeout reading from socket" err="read udp 0.0.0.0:46062: raw-read udp 0.0.0.0:46062: i/o timeout"`
- ip `… msg="Resolved target address" ip=68.183.27.198`
- ip_protocol `… msg="Resolving target address" ip_protocol=ip4`
- type `… msg="Beginning check" type=ping timeout_seconds=10`

For me `duration_seconds` and `err` feel like they might be nice ones to try to keep.

There's another visualisation option which we could try as an alternative to using `line_format` here, which is to enable the "Unique labels" option:
UniqueLabels enabled

I've created a dashboard and panel with this option enabled, which you can have a play with here:
https://samjewell.grafana.net/d/uQV39cg7k/unique-labels-option-enabled?orgId=1

That way we preserve the extra log content while also showing the unique labels like the Probe location.

samjewell (approved):
Nice work addressing this. I'm fine for you to ship this change as it is, but I have made a couple of suggestions for you to consider. Cheers 👏 🚢 👍 🎖️

src/dashboards/sm-dns.json
rdubrock: This was just the project json formatter having a field day, I think that this file had never been run through it before. Good points!

src/dashboards/sm-dns.json
rdubrock: Aha! This is exactly what I was trying to accomplish. I'll switch to that

PR #320, opened 2021-06-22
Viz icon overlay

rdubrock:
Our check visualization view isn't very accessible, because the only dimension differentiating the check status is color. This PR adds a toggle which will overlay icons onto the viz so they it's easier to see a distinction.

Without icons simulating achromatopsia:
Screen Shot 2021-06-22 at 1 48 05 PM

With icons simulating achromatopsia:
Screen Shot 2021-06-22 at 1 48 10 PM

### DIFF
--- a/src/components/CheckList.tsx
+++ b/src/components/CheckList.tsx
@@ -14,7 +14,18 @@ import {
   CheckType,
 } from 'types';
 import appEvents from 'grafana/app/core/app_events';
-import { Button, Icon, Select, Input, Pagination, InfoBox, Checkbox, useStyles, RadioButtonGroup } from '@grafana/ui';
+import {
+  Button,
+  Icon,
+  Select,
+  Input,
+  Pagination,
+  InfoBox,
+  Checkbox,
+  useStyles,
+  RadioButtonGroup,
+  InlineSwitch,
+} from '@grafana/ui';
 import { unEscapeStringFromRegex, escapeStringForRegex, GrafanaTheme, AppEvents, SelectableValue } from '@grafana/data';
 import { hasRole, checkType as getCheckType, matchStrings } from 'utils';
 import {
@@ -160,6 +171,7 @@ export const CheckList = ({ instance, onAddNewClick, checks, onCheckUpdate }: Pr
   const [selectAll, setSelectAll] = useState(false);
   const [viewType, setViewType] = useState(getViewTypeFromLS() ?? CheckListViewType.Card);
   const [sortType, setSortType] = useState<CheckSort>(CheckSort.AToZ);
+  const [showVizIconOverlay, setShowVizIconOverlay] = useState(false);
   const [bulkActionInProgress, setBulkActionInProgress] = useState(false);
   const styles = useStyles(getStyles);
   const successRateContext = useContext(SuccessRateContext);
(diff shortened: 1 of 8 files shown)


samjewell (commented):
This looks like a nice change, thanks for putting it together :-)

I just had a few questions:
1. Could the icons always show? Rather than having to toggle them on (ie. remove that toggle completely)? If we could keep them there always then both the code and the UX will be simpler.
2. Does the toggle itself get shown/hidden as the user switches onto and off the Visualisation View? Or is the toggle always shown? It seems to me that the toggle should live "inside" the visualisation view, as it only applies there. But the visual layout of the page doesn't imply this "inside" positioning to me.
3. Are these icons self-explanatory do you think, and/or that users will discover what they represent by interacting with the page? I wondered about introducing a key to explain each icon, but I expect that's overkill. I do wonder about reusing the Tick/Cross icons - we already use them for "Enabled/Disabled" on the other tabs, but they have a different meaning here. Maybe we could simply render the success rate (e.g. 94%) straight onto the hexagon itself, at least for red checks, so you can  immediately see the damage without hovering.
4. I'm a bit late to the party, but I'm wondering why we've chosen hexagons here, rather than a grid of squares/rectangles? I wonder if a standard heatmap visualisation from Grafana might be a nice alternative, and one where we can change the intensity of the red in response to the actual failure rate. That way we can solve the accessibility challenges of the heatmap once, and have that fix apply everywhere.

rdubrock:
Good thoughts @samjewell! 

>     1. Could the icons always show? Rather than having to toggle them on (ie. remove that toggle completely)? If we could keep them there always then both the code and the UX will be simpler.

They definitely could! The code complexity savings aren't huge, just because most of the trickiness is rendering svgs on top of svgs. The toggle is just a render/don't render condition. It might be that people just really like having the icons more. I'll check with Teddy. 

>     2. Does the toggle itself get shown/hidden as the user switches onto and off the Visualisation View? Or is the toggle always shown? It seems to me that the toggle should live "inside" the visualisation view, as it only applies there. But the visual layout of the page doesn't imply this "inside" positioning to me.

Good point. I'll check with Teddy on that. I thought about putting the toggle more "inside" the visualization, but it looked sort of lonely floating around by itself.

>     3. Are these icons self-explanatory do you think, and/or that users will discover what they represent by interacting with the page? I wondered about introducing a key to explain each icon, but I expect that's overkill. I do wonder about reusing the Tick/Cross icons - we already use them for "Enabled/Disabled" on the other tabs, but they have a different meaning here. Maybe we could simply render the success rate (e.g. 94%) straight onto the hexagon itself, at least for red checks, so you can  immediately see the damage without hovering.

I think it might be kind of hard to parse numbers, because they'll be small and there will be lots of them. The user feedback we've gotten is that folks wanted a quick "global up/down" view of all their checks at once. So the percentage is less important to them than "what's down right now". Totally agree on the icons, I'm not 100% happy with them. I've been playing around with other options like arrows, but I found that really difficult to parse



>     4. I'm a bit late to the party, but I'm wondering why we've chosen hexagons here, rather than a grid of squares/rectangles? I wonder if a standard heatmap visualisation from Grafana might be a nice alternative, and one where we can change the intensity of the red in response to the actual failure rate. That way we can solve the accessibility challenges of the heatmap once, and have that fix apply everywhere.

Just chose hexagons because they seemed fun, honestly. I think original inspiration came from the [polystat panel](https://github.com/grafana/grafana-polystat-panel). The trick here is that this visualization isn't in a dashboard, so we can't directly leverage any of the dashboard panels in this view. I think there is room for using a heatmap in our dashboards for sure, though

electron0zero (approved):

samjewell:
> Good thoughts @samjewell!

Thanks so much for all of your explanations - super nice of you to explain everything ☺️

Issue #321, opened 2021-06-23
Show probe locations count on a checks page


electron0zero:
**What would you like to be added**:

show probe location count on a checks page, like how we show frequency, and active series. 

we should add locations count so users can look at checks page and see how many probe locations a check is running at.

**Additional context**

![image](https://user-images.githubusercontent.com/9503187/123143619-69bba000-d478-11eb-9910-8248c2884f40.png)

electron0zero:
gonna take a stab at it.

PR #322, opened 2021-06-24
feat(ui): add locations count in checks list

electron0zero:
Fixes https://github.com/grafana/synthetic-monitoring-app/issues/321

## Screenshots

**Before:**
image
image

**After:**

for multiple locations:

image
image

for single location:

image
image

for disabled check:

image
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,9 +18,10 @@ We need to configure our local Grafana using provisioning. Provisioning does thr
 - Creates the necessary datasources
 - Provides some data to the plugin so it knows how to connect to the SM API and Cloud.
 
-Example provisioning files can be found in `/local-provisioning/datasources` and `/local-provisioning/plugins`
+Example provisioning files can be found in `scripts/local-provisioning/datasources` and `scripts/local-provisioning/plugins`
 
-- Create a new `yaml` file in each of `datasources` and `plugins` (they can be named anything). Copy the content from each example file and uncomment everything. These files will be mounted as volumes in the Grafana docker container.
+- Create new `yaml` files in `scripts/local-provisioning/datasources` and `scripts/local-provisioning/plugins` folders (they can be named anything).
+- Copy the content from each example file and uncomment everything. These files will be mounted as volumes in the Grafana docker container.
 - Fill in the values for the provisioning files.
   - Datasource info:
     - You'll need to input the basicAuthUser and password. You can find this information by going to your cloud portal, and copying the user info from the prometheus/loki datasource config UI
--- a/src/components/CheckListItem/CheckListItemDetails.tsx
+++ b/src/components/CheckListItem/CheckListItemDetails.tsx
@@ -12,7 +12,7 @@ const getStyles = (theme: GrafanaTheme) => ({
     white-space: nowrap;
     display: flex;
     align-items: center;
-    width: 284px;
+    width: 370px;
   `,
   labelWidth: css`
     max-width: 350px;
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -18,9 +18,10 @@ We need to configure our local Grafana using provisioning. Provisioning does thr
 - Creates the necessary datasources
 - Provides some data to the plugin so it knows how to connect to the SM API and Cloud.
 
-Example provisioning files can be found in `/local-provisioning/datasources` and `/local-provisioning/plugins`
+Example provisioning files can be found in `scripts/local-provisioning/datasources` and `scripts/local-provisioning/plugins`
 
-- Create a new `yaml` file in each of `datasources` and `plugins` (they can be named anything). Copy the content from each example file and uncomment everything. These files will be mounted as volumes in the Grafana docker container.
+- Create a new `yaml` files in `scripts/local-provisioning/datasources` and `scripts/local-provisioning/plugins` folders (they can be named anything).
```
electron0zero, line 23: I was confused by this, and thought I have to create `/local-provisioning` folder in root of repo.

updated it to make it clear that `/local-provisioning/` is inside scripts folder, and not root of repo.

rdubrock, line 23:
```suggestion
- Create new `yaml` files in `scripts/local-provisioning/datasources` and `scripts/local-provisioning/plugins` folders (they can be named anything). 
```

src/components/CheckListItem/CheckListItemDetails.tsx
```diff
@@ -22,18 +22,28 @@ const getStyles = (theme: GrafanaTheme) => ({
 interface Props {
   frequency: number;
   activeSeries?: number;
+  probeLocations?: number;
```
rdubrock, line 25: I don't think this value is possibly undefined. If `check.probes`  is undefined when it's passed as a prop to this component in `CheckListItem`, calling `.length` on it will cause the page to explode. So we probably need to make `probeLocations` behave more like the `frequency` prop instead of like the `activeSeries` prop.

rdubrock (changes requested):
Looks great! Good work getting through that.

src/components/CheckListItem/CheckListItemDetails.tsx
rdubrock: This was a really long winded way of saying I don't think you need the `?` and `undefined` checks in this component

src/components/CheckListItem/CheckListItemDetails.tsx
```diff
@@ -22,18 +22,28 @@ const getStyles = (theme: GrafanaTheme) => ({
...
+  frequency,
+  activeSeries,
+  probeLocations,
+  className,
+  labels,
+  onLabelClick,
+}: Props) => {
   const styles = useStyles(getStyles);
   const activeSeriesMessage = activeSeries !== undefined ? `${activeSeries} active series` : null;
+  const probeLocationsMessage = probeLocations !== undefined ? `${probeLocations} locations` : null;
```
rdubrock, line 42: You could check to see if there is only one location here to include/not include the `s` in location. So it would read `1 location` or `2 locations`

rdubrock: That's optional, though

src/components/CheckListItem/CheckListItemDetails.tsx
electron0zero: updated

src/components/CheckListItem/CheckListItemDetails.tsx
electron0zero: added, see screenshots in description

rdubrock (approved):
🦕

PR #327, opened 2021-07-08
fix: checking show advanced metrics updates usage calculation

jewbetcha:
Fixes usage calculation when creating a check and selecting "Publish advanced metrics". Also renamed form field to better reflect the value.

### DIFF
--- a/src/components/CheckEditor/CheckEditor.tsx
+++ b/src/components/CheckEditor/CheckEditor.tsx
@@ -151,11 +151,20 @@ export const CheckEditor = ({ check, onReturn }: Props) => {
             frequency={check?.frequency ?? fallbackCheck.frequency}
             probes={check?.probes ?? fallbackCheck.probes}
           />
-          <HorizontalCheckboxField
-            id="useFullMetrics"
-            name="useFullMetrics"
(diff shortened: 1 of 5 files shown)


src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -151,11 +151,20 @@ export const CheckEditor = ({ check, onReturn }: Props) => {
             frequency={check?.frequency ?? fallbackCheck.frequency}
             probes={check?.probes ?? fallbackCheck.probes}
           />
-          <HorizontalCheckboxField
-            id="useFullMetrics"
-            name="useFullMetrics"
-            label="Publish full set of metrics"
-            description={'Metrics are reduced by default'}
+          <Controller
```
rdubrock, line 154: At second glance, I'm not sure why/if this is necessary. The `HorizontalCheckboxField` component looks like it _should_ be registering with react-hook-form. I think it's OK because `HorizontalCheckboxField` will only try and register if it has a `name` prop (which is doesn't here), but now I'm concerned that maybe the checkbox field isn't working in the other places it's getting used

rdubrock (approved):
🦕

PR #329, opened 2021-07-09
fix: rename the success rate panels to reachability

rdubrock:
We need to align the naming of our measurements. The app was showing `Success Rate`, which is referred to as `Reachability` in the dashboards, so I changed the app to match the dashboards.

Screen Shot 2021-07-09 at 8 49 21 AM
Screen Shot 2021-07-09 at 8 49 58 AM


jewbetcha (approved):
Makes a lot more sense 👍

PR #336, opened 2021-07-27
Wrap probe token

### LINKED ISSUE #326: Probe Auth token text overflows the modal
Minor issue with how the Probe Authentication Token is displayed on the Probes page of the app, in my hosted-Grafana

**What happened**:
Text overflowed from the modal
image (3)

Text to wrap instead

Create a new probe

**Environment**:
- Grafana version: Current hosted-Grafana
- SM plugin version: Current hosted SM plugin
- User OS & Browser: MacOS BigSur 11.2.3, Chrome 91.0.4472.114 (Arm64)

rdubrock:
Fixes #326 
This fixes a layout issue with the token on probe creation, and adds a "copy to clipboard" button. I lifted the copy to clipboard from the onboarding plugin.

Before:
Screen Shot 2021-07-27 at 9 04 43 AM

After:
Screen Shot 2021-07-27 at 7 58 33 AM

### DIFF
--- a/src/components/ProbeEditor.tsx
+++ b/src/components/ProbeEditor.tsx
@@ -11,6 +11,7 @@ import {
   Switch,
   Legend,
   Alert,
+  useStyles2,
 } from '@grafana/ui';
 import { useForm, FormProvider } from 'react-hook-form';
 import { useAsyncCallback } from 'react-async-hook';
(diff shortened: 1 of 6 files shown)


jewbetcha (approved):
Nice, good call 👍

src/components/ProbeEditor.tsx
```diff
@@ -20,22 +21,33 @@ import { LabelField } from 'components/LabelField';
 import ProbeStatus from './ProbeStatus';
 import { InstanceContext } from 'contexts/InstanceContext';
 import { trackEvent, trackException } from 'analytics';
+import { GrafanaTheme2 } from '@grafana/data';
+import { Clipboard } from 'components/Clipboard';
```
electron0zero, line 25: Q: do we have clipboard in grafana ui, if not it think it makes sense for us to add it in grafana ui/toolkit.

we have other places(integrations) in Grafana where copy to clipboard is used.

electron0zero (approved):

src/components/ProbeEditor.tsx
rdubrock: Yeah, that's a good point. I adapted this one from how they were doing it in integrations, but it'd take some thought on how to make this more generic I think.

PR #338, opened 2021-07-28
Global thresholds

jewbetcha:
- Adds a form on the checks list to set threshold settings (for list view only)
- Adds uptime gauge to checks list
- Viz view reflects thresholds as well
- Works in tandem with new api routes (https://github.com/grafana/synthetic-monitoring-api/pull/327)


src/components/CheckList.tsx
```diff
@@ -648,6 +661,12 @@ export const CheckList = ({ instance, onAddNewClick, checks, onCheckUpdate }: Pr
           )}
         </div>
       )}
+      <ThresholdGlobalSettings
+        onDismiss={() => setShowThresholdModal(false)}
```
rdubrock, line 665: question: does onDismiss get called onSuccess? Or I guess the more relevant question is do we want the modal to close on successful select? I could go either way, just thinking aloud

src/components/ChecksVisualization/ChecksVisualization.tsx
```diff
@@ -132,13 +132,13 @@ export function ChecksVisualization({ checks, showIcons }: Props) {
         <div>{hoveredCheck?.target}</div>
         <div className={styles.successGaugeContainer}>
           <SuccessRateGauge
```
rdubrock, line 134: [random thinking for a future pr] we could put the other stats in this tooltip too. Maybe we could let folks choose which stat informs the color of the visualization too?

src/components/SuccessRateContextProvider.tsx
```diff
@@ -77,17 +100,16 @@ const parseProbeResults = (probes: Probe[] | undefined, data: any) => {
     return acc;
   }, {});
 
+  // TODO: Re-type latency check data
```
rdubrock, line 103: Is this still TODO? Fine if so and you want to do a follow up PR

src/components/Thresholds/ThresholdFormSection.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+const getSectionStyles = () => (theme: GrafanaTheme2) => ({
+  estimate: css`
+    margin-bottom: 0px;
+    font-style: italic;
+    align-self: center;
+    font-size: 0.7rem;
+  `,
+  disabled: css`
+    // This is a hack, can't seem to get the disabled prop on Input to work
```
rdubrock, line 50: I think you can get the disabled state to work by wrapping the `<Input />` in a `<Field  disabled={disabled} />` component. I don't _think_ that `Field` has to exist in a form, so you should be able to do it stand-alone. See [here](https://github.com/grafana/synthetic-monitoring-app/blob/efde3227caa8784f7e3677de1ef7a5eec57b3140/src/components/CheckTarget.tsx#L67)

src/components/Thresholds/ThresholdFormSection.tsx
```diff
@@ -0,0 +1,184 @@
...
+      const newThresholds = {
+        ...thresholds,
+        [key]: newValue,
+      };
+      setThresholds(newThresholds);
+    },
+    [thresholds, setThresholds]
+  );
+
+  const After = <div style={{ display: 'flex', alignItems: 'center', padding: '5px' }}>{unit}</div>;
```
rdubrock, line 106: [nit] consider adding these styles to your `getStyles` func and passing as a classname

src/components/Thresholds/ThresholdGlobalSettings.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  // Set thresholds in form when they are updated in context
+  useEffect(() => {
+    setUptimeThresholds(thresholds.uptime);
+    setReachabilityThresholds(thresholds.reachability);
+    setLatencyThresholds(thresholds.latency);
+  }, [thresholds]);
+
+  return (
+    <Modal title="Threshold Settings" isOpen={isOpen} onDismiss={onDismiss}>
```
rdubrock, line 84: I think I would prefer if this were a form so we could delegate all the state management shenanigans. Not a strong feeling, and it's up to you. Definitely don't need to do it in this PR.

src/components/Thresholds/ThresholdGlobalSettings.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  // Latency
+  expect(upperLimitInputs[2]).toHaveValue(5);
+  expect(lowerLimitInputs[2]).toHaveValue(10);
+});
+
+test('submits the form', async () => {
+  const instance = renderThresholdSettingsForm();
+  const saveButton = await screen.getByTestId('threshold-save');
+  userEvent.click(saveButton);
+  expect(instance.api?.updateTenantSettings).toHaveBeenCalled();
```
rdubrock, line 59: Maybe do a `toHaveBeenCalledWith({...})` so you can assert that the values are getting processed correctly.

src/contexts/SuccessRateContext.ts
```diff
@@ -30,35 +38,54 @@ export interface SuccessRates extends SuccessRatesByType {
...
+    uptimeValue: 0,
+    uptimeDisplayValue: 'N/A',
   },
 };
 
-const updateSuccessRate = (type: SuccessRateTypes, id: number, successRate: number | undefined) => {
-  const thresholdColor = getSuccessRateThresholdColor(successRate);
+export const defaultThresholds: ThresholdSettings = {
+  uptime: {
+    upperLimit: 90,
```
rdubrock, line 60: Are these the right defaults? I think the current defaults are 99.5, 99, and 75? Not sure. The latency defaults also seem a little small (if they are in `ms`)

src/utils.ts
```diff
@@ -276,28 +277,26 @@ export const fromBase64 = (value: string) => {
...
-  if (value >= 99) {
-    return config.theme2.colors.warning.main;
-  }
-  return config.theme2.colors.error.main;
+export const getSuccessRateThresholdColor = (
+  thresholds: ThresholdSettings,
+  key: 'reachability' | 'uptime' | 'latency',
+  compareValue: number
+) => {
+  return compareValue < thresholds[key].lowerLimit
```
rdubrock, line 285: Do you mind busting this out into `if` statements? Nested ternaries make my brain hurt

src/utils.ts
```diff
@@ -276,28 +277,26 @@ export const fromBase64 = (value: string) => {
...
-  if (value >= 99) {
-    return 'exclamation-triangle';
-  }
-  return 'times-square' as IconName;
+export const getSuccessRateIcon = (
+  thresholds: ThresholdSettings,
+  key: 'reachability' | 'uptime' | 'latency',
+  compareValue: number
+): IconName => {
+  return compareValue < thresholds[key].lowerLimit
```
rdubrock, line 297: Nested ternaries again. Really it's fine, just a personal preference thing.

rdubrock (approved):
Awesome work! I left some commentary, I don't think any of it is blocking necessarily, so I'm going ahead and approving.

src/components/CheckList.tsx
jewbetcha: Yep, upon successful update, `onSuccess` and `onDismiss` are called in succession.

https://github.com/grafana/synthetic-monitoring-app/pull/338/files#diff-e435217755667714e83a32a6ccbf2072b56705993b58c69d44fa7b78cef70efaR58

src/components/ChecksVisualization/ChecksVisualization.tsx
jewbetcha: Yes, absolutely 👍

src/components/SuccessRateContextProvider.tsx
jewbetcha: Still TODO, hoping for future self to come across it 😛

src/components/Thresholds/ThresholdFormSection.tsx
jewbetcha: Nice, will try

src/contexts/SuccessRateContext.ts
jewbetcha: Good catch 👍

PR #340, opened 2021-07-28
chore: silence inline svg errors

### LINKED ISSUE #339: Test output is full of errors
Something in the way grafana-ui is bundling icons doesn't play nice with our testing setup. I tracked down the error being thrown to a dependency of `react-inlinesvg`, which is what grafana/ui uses under the hood.

https://github.com/gilbarbara/react-from-dom/blob/cbca5083589fdb6810565309de59f52e3f5e61fc/src/index.ts#L227

rdubrock:
fixes #339

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,6 +8,10 @@ const standard = require('@grafana/toolkit/src/config/jest.plugin.config');
 const config = standard.jestConfig();
 // Adding jest-environment-jsdom-fifteen because the jsdom included with @grafana/toolkit doesn't play nice with the current version of @testing-library
 config.testEnvironment = 'jest-environment-jsdom-fifteen';
-config.setupFilesAfterEnv = ['@testing-library/jest-dom/extend-expect', ...(config.setupFilesAfterEnv || [])];
+config.setupFilesAfterEnv = [
+  '<rootDir>/src/test/setupTests.ts',
+  '@testing-library/jest-dom/extend-expect',
+  ...(config.setupFilesAfterEnv || []),
+];
 config.maxWorkers = 2;
 module.exports = config;
--- a/src/test/setupTests.ts
+++ b/src/test/setupTests.ts
@@ -1,0 +1,27 @@
+Object.defineProperty(window, 'matchMedia', {
+  writable: true,
+  value: jest.fn().mockImplementation((query) => ({
+    matches: false,
+    media: query,
+    onchange: null,
+    addListener: jest.fn(), // deprecated
+    removeListener: jest.fn(), // deprecated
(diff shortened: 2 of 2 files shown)


samjewell (approved):
I'd love a description on this PR personally - it wasn't clear to me where this improvement helps us.

First I tried to run the test suite locally, with `yarn build`. But I see plenty of test failures!
```
Test Suites: 14 failed, 5 passed, 19 total
Tests:       32 failed, 37 passed, 69 total
```
Should the tests fail locally, or not? I then ran this branch locally, but saw largely the same output.

Then I looked at the output on CircleCI for the "build and test frontend" step, and I could see that [here](https://app.circleci.com/pipelines/github/grafana/synthetic-monitoring-app/1468/workflows/f1207df5-79e4-466f-9ff9-d55b8f07dcc9/jobs/2736/parallel-runs/0/steps/0-104) we see _"Your output is too large to display in the browser. Only the last 400000 characters are displayed"_, while [here](https://app.circleci.com/pipelines/github/grafana/synthetic-monitoring-app/1469/workflows/24758e89-4fad-446e-bcde-1c13a2e86f65/jobs/2738/parallel-runs/0/steps/0-104) the whole output for that step is now visible, and not truncated.

That looks like your intended improvement, and look good, so I'm approving. I'm afraid I also don't really have much of a grasp on the code change you've made here - very happy to hear more about this if you're open to explaining it to me one day. Thanks!

samjewell:
Is there an alternative fix, which is to add the missing SVGs to the test setup / test suite? And would that be preferable here?

What if there's an SVG in future that we do expect to be in the app repo itself, but we don't spot that it's missing when the test suite runs, because we've silenced the warnings?

rdubrock:
Good points @samjewell !  So the problem we've been having is that the way `@grafana/ui` bundles icons doesn't play nice with our testing stack. It's a problem that all plugin developers have been seeing since v8. The tests still work and pass, but a library bundled with `@grafana/ui` will repeatedly throw an error and fill up the test output. The grafana core folks are working on a fix, but in order to make our test output readable, we can just ignore that specific error for now.

rdubrock:
Did you run `yarn` before running `yarn test` on your local? I think that might be why you were seeing your tests fail.

jewbetcha:
I'm trying this out also, the SVG warnings are gone but there seem to be a bunch of other new ones...

rdubrock:
I don't think the other ones are new, we just couldn't see them in the deluge

rdubrock:
Specifically I'm seeing a lot of stuff from `analytics.ts`

PR #352, opened 2021-08-17
Filter checks by probe

jewbetcha:
- adds a select to CheckList to filter by probes, with AND operator
- namespace CheckList with test file and some helpers

### DIFF
--- a/__mocks__/@grafana/ui.tsx
+++ b/__mocks__/@grafana/ui.tsx
@@ -40,6 +40,14 @@ const MultiSelect = forwardRef((props, ref) => <Select {...props} ref={ref} mult
 
 MultiSelect.displayName = 'MultiSelect';
 
+// Using this specifically for Probe filter at the moment
+const mockProbeOptions = [{ label: 'Chicago', value: 22 }];
+const AsyncMultiSelect = forwardRef((props, ref) => (
+  <Select {...props} options={mockProbeOptions} ref={ref} multiple={true} />
+));
+
+AsyncMultiSelect.displayName = 'AsyncMultiSelect';
+
 const Icon = (props) => <svg {...props} />;
 
 module.exports = {
@@ -47,4 +55,5 @@ module.exports = {
   MultiSelect,
   Select,
   Icon,
+  AsyncMultiSelect,
 };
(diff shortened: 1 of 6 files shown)


__mocks__/@grafana/ui.tsx
```diff
@@ -40,11 +40,20 @@ const MultiSelect = forwardRef((props, ref) => <Select {...props} ref={ref} mult
 
 MultiSelect.displayName = 'MultiSelect';
 
+// Using this specifically for Probe filter at the moment
```
rdubrock, line 43: So now that I think about it, if you're mocking the datasource in your tests (which I think you are), then we _could_ theoretically just call the `loadOptions` prop and pass that as options to the mocked select. Probably further down the rabbit hole than we need to go right now, just thinking out loud

src/components/CheckList/CheckList.tsx
```diff
@@ -540,6 +514,22 @@ export const CheckList = ({ instance, onAddNewClick, checks, onCheckUpdate }: Pr
             setCurrentPage(1);
           }}
           labelFilters={labelFilters}
+          className={styles.marginRightSmall}
+        />
+        <AsyncMultiSelect
+          data-testid="probe-filter"
+          prefix="Probes"
+          onChange={(v) => {
```
rdubrock, line 522: [nit] could this just be `onChange={setSelectedProbes}`?

src/page/ChecksPage.tsx
```diff
@@ -5,7 +5,7 @@ import React, { PureComponent } from 'react';
 import { Check } from 'types';
 import { getLocationSrv } from '@grafana/runtime';
 import { CheckEditor } from 'components/CheckEditor';
-import { CheckList } from 'components/CheckList';
+import { CheckList } from 'components/CheckList/CheckList';
```
rdubrock, line 8: You could put an `index.ts` that manages the exports from that directory if you like so the imports don't have to change

rdubrock (approved):
Looks great 🦏

src/components/CheckList/CheckList.tsx
jewbetcha: Will try!

PR #354, opened 2021-08-20
fix(ui): probe label display on probes page

electron0zero:
Fix display of labels on probes list page.

Before:
image

After:
image

### DIFF
--- a/src/components/ProbeList.tsx
+++ b/src/components/ProbeList.tsx
@@ -13,7 +13,7 @@ interface Props {
   onSelectProbe: (probeId: number) => void;
 }
 
-const labelsToString = (labels: Label[]) => labels.map(({ name, value }) => `${name}=${value}`);
+const labelsToString = (labels: Label[]) => labels.map(({ name, value }) => `${name}:${value}`).join(', ');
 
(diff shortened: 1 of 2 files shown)


samjewell (approved):
Nice, thanks for this!
Seems to be failing tests.

src/components/ProbeList.tsx
```diff
@@ -54,7 +54,7 @@ export const ProbeList = ({ probes, onAddNew, onSelectProbe }: Props) => {
                 <span className="add-data-source-item-text">{probe.name}</span>
                 <span className="add-data-source-item-desc">
                   <Badge color={color} icon={onlineIcon} text={onlineTxt} />
-                  <div>{labelsToString(probe.labels)}</div>
+                  <div>{labelsToString(probe.labels).join(', ')}</div>
```
jewbetcha, line 57: [nit] this `.join()` could be chained onto the end of `labelsToString` if you want

electron0zero: updated

jewbetcha: 🔥

PR #355, opened 2021-08-20
Checklist state refactor

jewbetcha:
A handful of changes to CheckList:
- all filters are stored in a central state object, the filtered list of checks responds to any changes to filter object
- moved all API actions into a separate file
- export a single filter matching function for readability

Next steps:
- the checks are ultimately rendered from `currentPageChecks` which isn't connected to our state, it feels like we could figure out a way to combine this with our filtered checks in an effort to hopefully avoid "derived state"
- `selectedChecks` and `selectAll` state can get out of sync if you manually select all checks, could be a low hanging follow-up to check if all are selected by comparing `selectedChecks` and the master list of checks, then we can remove another chunk of state


rdubrock (approved):
Looks like a big improvement!

Issue #375, opened 2021-11-04
Generate new SM auth tokens from the config page


rdubrock:
To help support terraform/cli users, there should be an easy way to generate an auth token from the UI that will allow them to interact with the SM api.

- API entrypoint is `/api/v1/token/create`
- Should be a button (probably on the config page?) that will make the call and present the token

PR #390, opened 2021-11-29
Gen terraform config

rdubrock:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/375

- Adds a button to generate a new SM access token on the plugin config page
- Adds a button to export current checks and private probes as terraform config in JSON

terraformconfig
.

### DIFF
--- a/src/components/AccessToken.tsx
+++ b/src/components/AccessToken.tsx
@@ -1,0 +1,47 @@
+import { Alert, Button, Link, Modal } from '@grafana/ui';
+import { InstanceContext } from 'contexts/InstanceContext';
+import React, { useContext, useState } from 'react';
+import { Clipboard } from 'components/Clipboard';
+
+export const AccessToken = () => {
+  const { instance } = useContext(InstanceContext);
+  const [showModal, setShowModal] = useState(false);
+  const [error, setError] = useState<string | undefined>();
+  const [token, setToken] = useState<string | undefined>();
+
+  const showTokenModal = async () => {
+    try {
+      const token = await instance.api?.createApiToken();
+      setToken(token);
+      setShowModal(true);
+    } catch (e) {
+      const cast = e as Error;
+      setError(cast.message ?? 'There was an error creating a new access token');
+    }
+  };
+  return (
+    <div>
+      <h5>Access tokens</h5>
+      <div>
+        <div>
+          You can use an SM access token to authenticate with the synthetic monitoring api. Check out the{' '}
+          <Link href="https://github.com/grafana/synthetic-monitoring-api-go-client">
+            synthetic monitoring go client
+          </Link>{' '}
+          or the{' '}
+          <Link href="https://registry.terraform.io/providers/grafana/grafana/latest/docs">
+            Grafana Terraform Provider
+          </Link>{' '}
+          to learn more about how to interact with the synthetic monitoring API.
+        </div>
+        <Button onClick={() => showTokenModal()}>Generate access token</Button>
+      </div>
+      <Modal title="Access Token" isOpen={showModal} onDismiss={() => setShowModal(false)}>
+        <>
+          {error && <Alert title={error} />}
+          {token && <Clipboard content={token} />}
+        </>
+      </Modal>
+    </div>
+  );
+};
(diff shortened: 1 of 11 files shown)


jewbetcha (approved):

src/components/TerraformConfig/TerraformConfig.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const probesConfig = probes?.reduce<TFProbeConfig>((acc, probe) => {
+      if (probe.public) {
+        return acc;
+      }
+      const probeConfig = probeToTF(probe);
+      const sanitizedName = sanitizeName(probe.name);
+      if (!acc[sanitizedName]) {
+        acc[sanitizedName] = probeConfig;
+      } else {
+        throw new Error(`Cannot generate TF config for probes with duplicate job names: ${probe.name}`);
```
mem, line 62: I'm going to guess that terraform names are case insensitve?

Because probe names cannot be duplicated, unless I'm missing something...

src/components/TerraformConfig/TerraformConfig.tsx
```diff
@@ -0,0 +1,136 @@
...
+      const type = checkType(check.settings);
+      if (type === CheckType.Traceroute) {
+        return acc;
+      }
+      const checkConfig = checkToTF(check);
+      const formattedJob = sanitizeName(check.job);
+      if (!acc[formattedJob]) {
+        acc[formattedJob] = checkConfig;
+      } else {
+        throw new Error(`Cannot generate TF config for checks with duplicate job names: ${check.job}`);
```
mem, line 48: what's unique is the job / target pair, job names can be duplicated, and in a sense they are expected to be.

src/components/AccessToken.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <div>
+          You can use an SM access token to authenticate with the synthetic monitoring api. Check out the{' '}
+          <Link href="https://github.com/grafana/synthetic-monitoring-api-go-client">
+            synthetic monitoring go client
+          </Link>{' '}
+          or the{' '}
+          <Link href="https://registry.terraform.io/providers/grafana/grafana/latest/docs">
+            Grafana Terraform Provider
+          </Link>{' '}
+          to learn more about how to interact with the synthetic monitoring API.
```
mem, line 35:
```suggestion
          documentation to learn more about how to interact with the synthetic monitoring API.
```

src/components/AccessToken.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }
+  };
+  return (
+    <div>
+      <h5>Access tokens</h5>
+      <div>
+        <div>
+          You can use an SM access token to authenticate with the synthetic monitoring api. Check out the{' '}
+          <Link href="https://github.com/grafana/synthetic-monitoring-api-go-client">
+            synthetic monitoring go client
```
mem, line 29:
```suggestion
            Synthetic Monitoring API Go client
```

src/datasource/DataSource.ts
```diff
@@ -427,6 +427,18 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
       });
   }
 
+  async createApiToken(): Promise<string> {
+    console.log('hello?');
```
mem, line 431: hello!

mem (commented):
Just a couple of minor corrections

src/components/TerraformConfig/TerraformConfig.tsx
rdubrock: 🤔 I might have to think about this. The problem is that in order to generate TF config, we have to use JSON, and the way the grafana terraform provider is structured expects checks to be identified by a single string key. I don't want to use the target as a json key because I think that will create it's own set of problems, so the job made the most sense. I could probably either generate a UUID, or just append something to the job name to guarantee uniqueness.

rdubrock: Just to illustrate the issue, the output config looks like this:
```JSON
    resource: {
      grafana_synthetic_monitoring_check: {
        a_jobname: {
          job: 'a jobname',
          target: 'example.com',
          enabled: true,
          probes: [1],
          labels: {},
          settings: { ping: { ip_version: 'V4', dont_fragment: false } },
        },
          some_other_name: {
          job: 'some other name',
          target: 'example.com',
          enabled: true,
          probes: [1],
          labels: {},
          settings: { ping: { ip_version: 'V4', dont_fragment: false } },
        },
      },
      }
      ```

src/components/TerraformConfig/TerraformConfig.tsx
rdubrock: good point, just being extra cautious

PR #402, opened 2022-01-06
fix: terraform import command for probes & link to docs

electron0zero:
current tf import command doesn't mention a way to import probe auth token, fix that and link to tf providor docs


jewbetcha (approved):
Approved, but there's a failure in the build just from a prettier error. Good to ship after fixing that :+1:

Issue #407, opened 2022-01-12
Usage calculation can vary between list/edit views


rdubrock:
The usage calculator in the edit view isn't taking into account series differences that can happen from check settings like SSL config in HTTP checks or custom labels. We should add those into the calculation so it more closely matches what we show in the list view.

PR #410, opened 2022-01-18
fix: [cr] check class correctly in CheckList

jewbetcha:
Partially fixes https://github.com/grafana/synthetic-monitoring-app/issues/407

CheckList was calculating all checks as `<check_type>_ssl`, since (I think) the API is returning an empty set of TLS settings for every non-TLS check. This PR makes sure that there are actually settings present inside the check, and both calculations are the same if it is non-tls.

There is however another issue that during new/edit a check, we aren't watching TLS settings in our calculations. This will take a bit more refactoring so I figured a follow-up PR would be best for that.


rdubrock (approved):
🦕

Issue #84, opened 2020-08-31
Provide a visual clue as to where a probe is located?


mem:
When creating a probe, this is what you see:

![new-probe](https://user-images.githubusercontent.com/189701/91743985-fb6ac180-eb75-11ea-8533-45ac3c1ebb17.png)

it would be nice to have some visual feedback as to what those numbers correspond to.

mem:
We already have the worldmap panel available and I wonder if it's possible to use it for this: simply show the location as the user is entering it.

I imagine this would save some users the uncertainty of whether they got lat / long right or flipped. It would also help with what's positive and what's negative.

mem:
@rdubrock any take on this?

jewbetcha:
Released with 1.9.7

PR #412, opened 2022-01-19
feat: [cr] adds a simple map with lat/long coordinates on probe edit page

jewbetcha:
Screen Shot 2022-01-19 at 1 41 32 PM

Preview for creating private probes as well as some visual cues on established public probes.

Addresses https://github.com/grafana/synthetic-monitoring-app/issues/84


src/components/SimpleMap/SimpleMap.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { ComposableMap, Geographies, Geography, Marker } from 'react-simple-maps';
+
+const geoUrl = 'https://raw.githubusercontent.com/zcreativelabs/react-simple-maps/master/topojson-maps/world-110m.json';
```
rdubrock, line 4: make me slightly nervous to depend on an external resource like this. Is it reasonable to copy the json into our repo, or is that a bad idea?

rdubrock (approved):
🦉

src/components/SimpleMap/SimpleMap.tsx
jewbetcha: Hmm yeah, let me see what I can do about that

electron0zero: have we looked into pulling in geomap panel? maybe we can use geomap with one layer and controls disabled 🤔

jewbetcha: We haven't, but in my past experience its not a trivial thing to use a panel outside of a dashboard

electron0zero (approved):

PR #418, opened 2022-01-31
Check page routing

jewbetcha:
Check edit/new form and probe edit/new form are controlled by nested routes and route params. Tests updated accordingly.


rdubrock (approved):
🦚

PR #421, opened 2022-02-02
chore: rework logic to extrac values from alert rules

mem:
This is trying to make the extraction logic more generic, so that it
works with other alerts similar to the existing ones, but not exactly
like them.

### DIFF
--- a/src/components/alertingTransformations.ts
+++ b/src/components/alertingTransformations.ts
@@ -1,5 +1,5 @@
-import { AlertFormValues, AlertRule, Label } from 'types';
-import { ALERT_PROBE_SUCCESS_RECORDING_METRIC } from 'components/constants';
+import { AlertDescription, AlertFormValues, AlertRule, AlertSensitivity, Label } from 'types';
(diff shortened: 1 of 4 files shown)


src/components/alertingTransformations.ts
```diff
@@ -21,3 +21,29 @@ export const transformAlertFormValues = (alertValues: AlertFormValues | undefine
...
   };
 };
+
+export const alertDescriptionFromRule = (rule: AlertRule): AlertDescription | undefined => {
+  const result = ALERT_RULE_EXPR_REGEX.exec(rule.expr);
+  if (!result) {
+    return undefined;
+  }
+
+  const strToEnum = (value: string): AlertSensitivity | undefined => {
```
rdubrock, line 31: nit: I'd say thins function could move outside the scope of `alertDescriptionFromRule` and be named something like `isAlertSensitivity`

rdubrock (approved):

src/components/alertingTransformations.ts
mem: thanks!

PR #422, opened 2022-02-03
feat: [cr] refactor of filtering UX, updates tests

jewbetcha:
Screen Shot 2022-02-03 at 9 37 34 AM
Screen Shot 2022-02-03 at 9 38 28 AM
Screen Shot 2022-02-03 at 9 38 38 AM

Re-arranged filters under a central dropdown, with some state to show when active/inactive. Also cut down on vertical rows after talking with Teddy about it.

Also changed the empty check list alert to use `<Alert />` since `<InfoBox />` is marked as deprecated.

### DIFF
--- a/src/components/CheckList/CheckFilterGroup.tsx
+++ b/src/components/CheckList/CheckFilterGroup.tsx
@@ -1,0 +1,94 @@
+import React, { useCallback, useEffect, useState } from 'react';
+
+import { Button, Icon, useStyles } from '@grafana/ui';
+import { css } from '@emotion/css';
+import { GrafanaTheme } from '@grafana/data';
+import { defaultFilters, CheckFilters } from './CheckList';
+
+const groupStyles = (theme: GrafanaTheme) => ({
+  container: css`
+    position: relative;
+  `,
+  dropdown: css`
+    position: absolute;
+    background-color: rgb(24, 27, 31);
+    border: 1px solid rgba(204, 204, 220, 0.15);
+    border-radius: 2px;
+    z-index: 100;
+    padding: 20px;
+    margin-top: 5px;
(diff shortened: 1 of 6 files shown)


src/components/CheckList/CheckFilterGroup.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const styles = useStyles(groupStyles);
+
+  const handleFilterOpen = useCallback(() => {
+    setOpenFilters(!openFilters);
+  }, [openFilters]);
+
+  useEffect(() => {
+    let active = 0;
+    // Count which filters have been applied
+    Object.keys(filters).map((key) => {
```
rdubrock, line 58: philisophical nit: this could be a forEach or a reduce since you don't need the array produced by mapping

src/components/CheckList/CheckList.test.tsx
```diff
@@ -214,6 +214,8 @@ test('search matches label name', async () => {
 
 test('clicking label value adds to label filter', async () => {
   renderCheckList();
+  const additionalFilters = await screen.findByText('Additional Filters');
```
rdubrock, line 217: nit: using `findByText` is fine, but I think `findByRole` is the more encouraged query:

```
findByRole('button', {name: 'Additional Filters', exact: false})
```

rdubrock (approved):
🪲

src/components/CheckList/CheckFilterGroup.tsx
jewbetcha: Yes, yes it could

src/components/CheckList/CheckList.test.tsx
jewbetcha: 👍

PR #423, opened 2022-02-03
feat: add probe duration and SSL expiry alerts

mem:

### DIFF
--- a/src/components/AlertRuleForm.tsx
+++ b/src/components/AlertRuleForm.tsx
@@ -2,7 +2,14 @@ import { AppEvents, GrafanaTheme, SelectableValue } from '@grafana/data';
 import { Alert, Button, Field, HorizontalGroup, Icon, Input, Label, Select, useStyles } from '@grafana/ui';
 import React, { useState, useContext } from 'react';
 import { Controller, FormProvider, useForm } from 'react-hook-form';
-import { AlertRule, AlertSensitivity, Label as LabelType, TimeUnits } from 'types';
+import {
+  AlertFormExpressionContent,
+  AlertFormValidations,
+  AlertRule,
+  AlertSensitivity,
+  Label as LabelType,
+  TimeUnits,
+} from 'types';
 import { ALERT_SENSITIVITY_OPTIONS, TIME_UNIT_OPTIONS } from './constants';
 import { css } from '@emotion/css';
 import { AlertLabels } from './AlertLabels';
@@ -24,10 +31,11 @@ export enum AlertTimeUnits {
   Years = 'y',
 }
 
-export interface AlertFormValues {
-  expression?: string;
+interface AlertFormValues {
   name: string;
-  probePercentage: number;
+  metric: string;
+  threshold: number;
+  operator: string;
   timeCount: number;
   sensitivity: SelectableValue<AlertSensitivity>;
   timeUnit: SelectableValue<TimeUnits>;
@@ -57,15 +65,15 @@ const getAlertFormValues = (rule: AlertRule): AlertFormValues | undefined => {
     value,
   }));
 
-  const probePercentage = desc.threshold;
-
-  if (!timeOption || !probePercentage || !sensitivityOption) {
+  if (!timeOption || !desc.threshold || !sensitivityOption) {
     return undefined;
   }
 
   return {
     name: rule.alert,
-    probePercentage,
+    metric: desc.metric,
+    threshold: desc.threshold,
+    operator: desc.operator,
     timeCount: parseInt(timeCount, 10),
     timeUnit: timeOption ?? {},
     sensitivity: sensitivityOption,
@@ -138,10 +146,12 @@ const getStyles = (theme: GrafanaTheme) => ({
 
 type Props = {
   rule: AlertRule;
+  expressionContent: AlertFormExpressionContent;
+  validations: AlertFormValidations;
   onSubmit: (alertValues: AlertFormValues) => Promise<FetchResponse<unknown> | undefined>;
 };
 
-export const AlertRuleForm = ({ rule, onSubmit }: Props) => {
+export const AlertRuleForm = ({ rule, onSubmit, expressionContent, validations }: Props) => {
   const defaultValues = getAlertFormValues(rule);
   const { instance } = useContext(InstanceContext);
   const styles = useStyles(getStyles);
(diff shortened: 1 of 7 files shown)


src/components/AlertRuleForm.tsx
```diff
@@ -141,7 +142,11 @@ type Props = {
   onSubmit: (alertValues: AlertFormValues) => Promise<FetchResponse<unknown> | undefined>;
 };
 
-export const AlertRuleForm = ({ rule, onSubmit }: Props) => {
+const CommonAlertForm = (
+  rule: AlertRule,
+  onSubmit: (alertValues: AlertFormValues) => Promise<FetchResponse<unknown> | undefined>,
+  expressionContent: (styles: any, control: any, errors: any, register: any) => any
```
mem, line 148: I have no idea what these types are supposed to be...

src/components/Alerting.test.tsx
```diff
@@ -55,6 +60,7 @@ const toggleSection = async (sectionName: string): Promise<HTMLElement> => {
 };
 
 it('adds default alerts and edits alerts', async () => {
+  return;
```
mem, line 63: oops...

src/components/Alerting.tsx
```diff
@@ -156,9 +156,20 @@ export const Alerting: FC = () => {
...
+
+          case AlertFamily.ProbeSuccess:
+            return <AlertRuleForm key={`${rule.alert}-${index}`} rule={rule} onSubmit={getUpdateRules(index)} />;
+
+          case AlertFamily.SSLCertExpiry:
+            return <SSLCertExpiryAlertForm rule={rule} onSubmit={getUpdateRules(index)} />;
+        }
+
+        return undefined;
+      })}
```
mem, line 172: I'm sure there's a better way...

src/components/alertingTransformations.ts
```diff
@@ -13,8 +13,8 @@ export const labelToProm = (labels?: Label[]) => {
 export const transformAlertFormValues = (alertValues: AlertFormValues | undefined): AlertRule => {
   return {
     alert: alertValues?.name ?? '',
-    expr: `${ALERT_PROBE_SUCCESS_RECORDING_METRIC}{alert_sensitivity="${alertValues?.sensitivity?.value}"} < ${
-      alertValues?.probePercentage ?? ''
+    expr: `${alertValues?.metric}{alert_sensitivity="${alertValues?.sensitivity?.value}"} ${alertValues?.operator} ${
```
mem, line 16: `metric` is not part of the editable values, but I need to access the metric name here...

src/components/AlertRuleForm.tsx
```diff
@@ -141,7 +142,11 @@ type Props = {
   onSubmit: (alertValues: AlertFormValues) => Promise<FetchResponse<unknown> | undefined>;
 };
 
-export const AlertRuleForm = ({ rule, onSubmit }: Props) => {
+const CommonAlertForm = (
```
rdubrock, line 145: I think there might be a couple ways of doing this that are a bit more React-y. 

CommonAlertForm should be a component (it pretty much already is, it just needs to take in a single object as it's arguments). Something like this
```
const CommonAlertForm = ({ rule, onSubmit, subForm }: Props)
```

The `subForm` prob can be the sub section of form inputs you want to render. That way you could just render `CommonAlertForm` like this:
```
<CommonAlertForm rule={alertRule} onSubmit={handleOnSubmit} subForm={ProbeDurationAlertForm} />
```

Alternatively you could leverage React.Children:
```
const CommonAlertForm = ({ rule, onSubmit, children }: Props)
```
And that would lead to 
```
<CommonAlertForm rule={alertRule} onSubmit={handleOnSubmit}>
    <ProbeDurationAlertForm rule={alertRule} ... />
</CommonAlertForm>
```

src/components/AlertRuleForm.tsx
```diff
@@ -320,3 +283,168 @@ export const AlertRuleForm = ({ rule, onSubmit }: Props) => {
     </>
   );
 };
+
+export const ProbeDurationAlertForm = ({ rule, onSubmit }: Props) => {
```
rdubrock, line 287: I think all three alert forms could move into their own file. They could be grouped in a directory like we do with `CheckListItem` where we only expose what needs to be consumed externally from an `index.ts`

src/components/AlertRuleForm.tsx
```diff
@@ -212,49 +217,7 @@ export const AlertRuleForm = ({ rule, onSubmit }: Props) => {
...
-                  />
-                </Field>
-                <div className={styles.selectInput}>
-                  <Controller
-                    render={({ field }) => <Select {...field} options={TIME_UNIT_OPTIONS} />}
-                    control={control}
-                    name="timeUnit"
-                  />
-                </div>
+                {expressionContent(styles, control, errors, register)}
```
rdubrock, line 220: If you end up switching to one of the other approaches I mentioned above, you could change this to be just 
```
{children}
```
or 
```
<SubForm ... />
```

rdubrock (approved):
Thanks for doing this!

rdubrock (changes requested):
Clicked the wrong thing the first time!

src/components/AlertRuleForm.tsx
rdubrock: Any alternatives/opinions @jewbetcha?

jewbetcha: Agree. I think the `children` method sounds good for this, and can wrap all the other forms.

src/components/Alerting.tsx
jewbetcha: Not crazy about the need for a `switch` here but it seems like the logic is pretty intense

src/components/Alerting.tsx
```diff
@@ -156,9 +156,20 @@ export const Alerting: FC = () => {
           </Button>
         </div>
       )}
-      {alertingRules.map((alertRule, index) => (
-        <AlertRuleForm key={`${alertRule.alert}-${index}`} rule={alertRule} onSubmit={getUpdateRules(index)} />
-      ))}
+      {alertingRules.map((rule, index) => {
+        switch (alertFamilyFromRule(rule)) {
+          case AlertFamily.ProbeDuration:
+            return <ProbeDurationAlertForm rule={rule} onSubmit={getUpdateRules(index)} />;
```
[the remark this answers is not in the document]
mem: thanks for the pointer!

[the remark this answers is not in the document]
mem: Fixed

src/components/Alerting.tsx
```diff
@@ -156,9 +156,20 @@ export const Alerting: FC = () => {
...
+
+          case AlertFamily.ProbeSuccess:
+            return <AlertRuleForm key={`${rule.alert}-${index}`} rule={rule} onSubmit={getUpdateRules(index)} />;
+
+          case AlertFamily.SSLCertExpiry:
+            return <SSLCertExpiryAlertForm rule={rule} onSubmit={getUpdateRules(index)} />;
+        }
+
+        return undefined;
```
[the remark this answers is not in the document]
mem: Fixed

src/components/AlertRuleForm.tsx
mem: Ok, I ... made changes 😄

src/components/AlertRuleForm.tsx
mem: Fixed

rdubrock:
So I dove in and did a take on how I'd solve this problem: https://github.com/grafana/synthetic-monitoring-app/pull/426

The main thing is that I feel like the different forms are similar enough that we can just abstract the differences and pass those as props. What changes is some of the language in the UI and threshold values on a couple inputs, and the form itself doesn't really have to know about those things. 

I also think the approach going on here works fine! It just needs a few small tweaks to pass the tests I believe.  I'll leave it up to you to compare and see which one appeals more.

src/components/AlertRuleForm.tsx
```diff
@@ -139,9 +140,10 @@ const getStyles = (theme: GrafanaTheme) => ({
 type Props = {
   rule: AlertRule;
   onSubmit: (alertValues: AlertFormValues) => Promise<FetchResponse<unknown> | undefined>;
+  children: React.ReactNode;
 };
 
-export const AlertRuleForm = ({ rule, onSubmit }: Props) => {
+export const CommonAlertForm = ({ rule, onSubmit, children }: Props) => {
```
rdubrock, line 146: You might bump into some weird behavior here. You are creating a form and all it's state management and context and whatnot here in CommonAlertForm on line 151, but doing it again inside the child component (like on line 287). I think one or the other is fine, but doing both might result in something unexpected happening

rdubrock: I think leaving it up to the children is probably best?

mem:
> So I dove in and did a take on how I'd solve this problem: #426
> 
> The main thing is that I feel like the different forms are similar enough that we can just abstract the differences and pass those as props. What changes is some of the language in the UI and threshold values on a couple inputs, and the form itself doesn't really have to know about those things.
> 
> I also think the approach going on here works fine! It just needs a few small tweaks to pass the tests I believe. I'll leave it up to you to compare and see which one appeals more.

Thanks, what you are doing makes sense to me!

The only reason I didn't start by doing that is because I wasn't sure how different the different alerts would be, but you are right, they are still very similar, and the differences are better captured in the way you are doing it.

src/components/Alerting.tsx
mem: the switch is gone now...

src/components/AlertRuleForm.tsx
mem: they are now gone

rdubrock (approved):

rdubrock:
Closing because it will be covered by our future plans for handling alerting

Issue #419, opened 2022-01-31
Provision the plugin with datasource UIDs instead of names


rdubrock:
Renaming a stack in Cloud results in the SM plugin losing track of which prometheus and loki datasources it should be talking to. This happens because prom/loki datasources get renamed when the stack gets renames, but the provisioning that informs the plugin doesn't have the new names. The plugin identifies datasources by name, so things stop working. We should look into updating the plugin provisioning to use UIDs instead (if possible), it will be more stable and the SM plugin will continue to work without interruption across renames.

stephaniehingtgen:
Thanks for creating this, @rdubrock! Some background from the HG side:

- Currently in production the `grafanaName` in the json data isn't changed for SM, but I'm working on changing the datasources names now as the datasources are renamed [issue here](https://github.com/grafana/hosted-grafana/issues/1835). This will still require an uninstall & reinstall, though, to get SM up and running again.
- HG names datasources as `grafanacloud-<slug>-logs` but keeps the UIDs consistent across all instances so that we are able to key off of those (they will look like `grafanacloud-logs` or `grafanacloud-prom`).
- HG will provision the datasources via a yaml file, with the UID specified, so the UID will exist from the start (in cloud at least).

Also, in case it's helpful to reference, the k8s integration uses UID of the prometheus datasource in order to grab the scrape interval from that config.

rdubrock:
Adding some thoughts after a bit of looking:

- We have to support looking up the datasources by name AND uid. Since the SM plugin can be installed anywhere we aren't guaranteed to have provisioning or the consistent cloud uids
- We can use UIDs instead of names for generating the SM dashboards
- We're going to need to do some sort of migration for existing tenants to move them from name => uid in cloud

mem:
Since this is merged, can this issue be closed or is there missing work?

rdubrock:
We got a report that renaming stacks will break SM until the datasource is deleted, so there is still an issue here.

rdubrock:
This should be fixed with version 1.11.0

Issue #274, opened 2021-05-27
Changes in provisioning file are not updated in Datasource


electron0zero:
**What happened**:
I updated apiHost in Provisioning file, and restarted grafana. After restarted plugin was still making api calls to old apiHost.

synthetic-monitoring settings (`<grafana-url>/api/plugins/grafana-synthetic-monitoring-app/settings`) was showing updated value of apiHost

after some investigation, I found that Synthetic Monitoring Datasource still has old value (check it at `<grafana-url>/api/datasources`)

**What you expected to happen**:
after updating provisioning file and restarting grafana, plugin should have updated the value apiHost

**How to reproduce it (as minimally and precisely as possible)**:
- install synthetic-monitoring
- setup and use it
- change provisioning file
- restart grafana
- changes in provisioning file should be reflated in in data-source.

**Workaround:**
- disable plugin
- delete Synthetic Monitoring Datasource
- enabled plugin
- go to home screen and click monitor your systems

now datasource should get values from provisioning file, you can verify it at `<grafana-url>/api/datasources`

peterschretlen:
No plans to address this - it does surface occassionaly, the listed workaround of re-initializing the plugin works

PR #424, opened 2022-02-03
fix: try and use UIDs for datasources if they are available

rdubrock:
This would be a first step towards fixing https://github.com/grafana/synthetic-monitoring-app/issues/419

I think ultimately to resolve this issue we're gonna need to do a few things that might be worth breaking into a few different PRs:

- Auto-update dashboards on plugin load (we can look at GEM for an example of how to do this)
- Auto-update the SM datasource JSON on load ([also reported as an issue for other reasons](https://github.com/grafana/synthetic-monitoring-app/issues/274))
- Update the cloud provisioning to include the uid

### DIFF
--- a/src/components/TenantView/LinkedDatasourceView.tsx
+++ b/src/components/TenantView/LinkedDatasourceView.tsx
@@ -1,18 +1,19 @@
 import React from 'react';
-import { LinkedDatsourceInfo } from 'datasource/types';
-import { config } from '@grafana/runtime';
+import { LinkedDatasourceInfo } from 'datasource/types';
 import { Spinner } from '@grafana/ui';
 import { useNavigation } from 'hooks/useNavigation';
 import { ROUTES } from 'types';
+import { findLinkedDatasource } from 'utils';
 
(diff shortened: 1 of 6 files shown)


src/components/TenantView/LinkedDatasourceView.tsx
```diff
@@ -1,18 +1,19 @@
...
-  info: LinkedDatsourceInfo;
+  info: LinkedDatasourceInfo;
 }
 
 const LinkedDatasourceView = ({ info }: Props) => {
   const navigate = useNavigation();
-  const datasource = config.datasources[info.grafanaName];
+  const datasource = findLinkedDatasource(info);
 
+  console.log({ datasource, info });
```
jewbetcha, line 16: 🤖

jewbetcha (approved):

src/components/TenantView/LinkedDatasourceView.tsx
rdubrock: oopsy doopsy

PR #425, opened 2022-02-04
Sync ds json data

### LINKED ISSUE #274: Changes in provisioning file are not updated in Datasource

### LINKED ISSUE #419: Provision the plugin with datasource UIDs instead of names

rdubrock:
Closes #419 
Closes #274 

This will leverage the `preload: true` plugin setting to update to sync provisioned data with the SM datasource whether a use navigates to the plugin or not. Currently if things change users have to delete the SM datasource and reinitialize the plugin.

This also checks to see whether dashboards have changed and updates them.

TODO: tests!

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -1,7 +1,11 @@
-import { AppPlugin } from '@grafana/data';
+import { AppPlugin, PluginMeta } from '@grafana/data';
 import { GlobalSettings } from './types';
 import { App } from 'components/App';
 import { ConfigPage } from 'page/ConfigPage';
+import { findSMDataSources } from 'utils';
+import { getBackendSrv } from '@grafana/runtime';
+import { updateSMDatasource } from 'initialization-utils';
+import { dashboardsNeedingUpdate, importAllDashboards, importDashboard } from 'dashboards/loader';
 
 export const plugin = new AppPlugin<GlobalSettings>().setRootPage(App).addConfigPage({
   title: 'Config',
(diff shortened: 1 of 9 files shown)


src/module.ts
```diff
@@ -1,11 +1,61 @@
...
+  ) {
+    return true;
+  }
+  return false;
+};
+
+const preloadInit = async () => {
+  const smDatasources = findSMDataSources();
+  const pluginSettings = await getPluginSettings();
+  // if there is no SM datasource, we have nothing to do. If there is more than 1 we don't know what to do.
```
jewbetcha, line 42: Could maybe short circuit this function earlier based on what comes back in `smDatasources`

jewbetcha (commented):
Any luck showing feedback to user if an update is happening?

rdubrock:
> Any luck showing feedback to user if an update is happening?

I haven't experimented with that yet, will try next week I think. They do get a `Datasource updated` toast, which isn't super helpful. There's nothing for when dashboards update.

src/module.ts
```diff
@@ -1,11 +1,61 @@
...
+import { getDashboardsNeedingUpdate, importAllDashboards } from 'dashboards/loader';
 
 export const plugin = new AppPlugin<GlobalSettings>().setRootPage(App).addConfigPage({
   title: 'Config',
   icon: 'cog',
   body: ConfigPage,
   id: 'config',
 });
+
+const getPluginSettings = async () =>
```
rdubrock, line 17: We have to fetch this from the backend because `plugin.meta` is undefined at this point in the execution

rdubrock:
So we _can_ give a toast notifying the user that the SM dashboards were updated. I skipped adding one for when the datasource is updated as that's an implementation detail and should be transparent to users. I think this is a pretty high risk change, if folks are willing to give this a careful 👀 and maybe even test locally I'd appreciate it.

PR #438, opened 2022-03-02
fix: handle dashboard redirects

### LINKED ISSUE #432: Frontend routing isn't correctly routing dashboard redirects
To reproduce:
- Go to the summary dashboard
- Click on a link in the ping checks table
- The link will go to something like `.../a/grafana-synthetic-monitoring-app/?page=redirect&dashboard=ping&var-probe=All&var-instance=google.com&var-job...`

What happens:
- This links back to the plugin which routes this to the home page

What should happen:
- This should redirect to the ping dashboard

rdubrock:
Fixes #432 

~I'm not really sure why, but react-router won't pick up the route change unless the browser refreshes. I feel like I've missed something here.~

Should be good to go

### DIFF
--- a/src/components/DashboardUpdateModal.tsx
+++ b/src/components/DashboardUpdateModal.tsx
@@ -4,13 +4,11 @@ import { InstanceContext } from 'contexts/InstanceContext';
 import { importAllDashboards, listAppDashboards } from 'dashboards/loader';
 import { Button, HorizontalGroup, Modal } from '@grafana/ui';
 import { hasDismissedDashboardUpdateModal, persistDashboardModalDismiss } from 'sessionStorage';
-import { useNavigation } from 'hooks/useNavigation';
 
 export const DashboardUpdateModal = () => {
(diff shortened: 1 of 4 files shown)


jewbetcha:
I think I figured it out. The table with check links in the summary dashboard is building the links as:
`/a/grafana-synthetic-monitoring-app/?page=redirect...` with the `page` query param being an old vestige before we migrated to React Router, but when we're looking at dashboards, technically we're in Grafana router land. If you change the links in the table to be `/a/grafana-synthetic-monitoring-app/redirect?...` it seems to redirect correctly since it goes through the `<DashboardRedirector />` route component.

I believe you can take away the whole chunk of logic in regards to the `page` query param, as long as we make sure we aren't using it anywhere anymore.

rdubrock:
Yessss! OK that's a much better way to go. That page logic was there to handle bookmarks or things like that, but apparently it hasn't been working anyways.

jewbetcha (approved):

PR #451, opened 2022-04-07
chore: remove deprecated metrics OR queries from dashboards

### LINKED ISSUE #401: Remove the OR clauses in the dashboards
In order to navigate a metric name change (from `probe_success` to `probe_all_success`), we added an `OR` clause to many of the queries in our dashboards. I think that the time window of needing to worry about that metric naming has passed, so we could maybe get rid of all those OR clauses now.

electron0zero:
Removed old deprecated metrics from dashboards.

Fixes #401

### DIFF
--- a/src/dashboards/sm-dns.json
+++ b/src/dashboards/sm-dns.json
@@ -1089,5 +1089,5 @@
   "timezone": "",
   "title": "Synthetic Monitoring DNS",
   "uid": "lgL6odgGz",
-  "version": 20
+  "version": 21
 }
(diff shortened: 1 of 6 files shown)


rdubrock (changes requested):
Waaay down at the bottom of every dashboard file there is a `version` key. That version value needs to be manually incremented every time we make a change to dashboards. It's how the plugin detects there has been updates and prompts users to re-import.

electron0zero:
> Waaay down at the bottom of every dashboard file there is a `version` key. That version value needs to be manually incremented every time we make a change to dashboards. It's how the plugin detects there has been updates and prompts users to re-import.

thanks, updated now 👍🏼

PR #461, opened 2022-04-26
Docker: add a basic docker setup for development

leventebalogh:
### What changed?
Introduced a basic Docker setup (`Dockerfile` and `docker-compose.yml`) to make it easier to develop the plugin against a certain Grafana version. (It can be that I have missed something and it already exists in the repo somewhere, in that case please let me know).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+ARG grafana_version=latest
+
+FROM grafana/grafana:${grafana_version}
+
+# Make it as simple as possible to access the grafana instance for development purposes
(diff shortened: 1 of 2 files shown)


electron0zero:
we have https://github.com/grafana/synthetic-monitoring-app/blob/main/scripts/run-local.sh to run it locally, @rdubrock has more context on it.

leventebalogh:
Aaa, thanks @electron0zero, I will close the PR then, doesn't make sense to have two solutions for the same problem.

PR #471, opened 2022-06-22
Fix duplicate panel ID in this dashboard

ethirolle:
In support ticket, a customer discovered that they are unable to edit the "Average latency" panel in the "Synthetic Monitoring Ping" dashboard provided by the SM App. The problem (as described in https://community.grafana.com/t/selecting-edit-on-panel-opens-edit-window-for-a-different-panel/5435/2) is that two panels have this in their JSON:

```
      "id": 34,
```

By changing one of the panels to have id=33, that fixes the problem

n/a


ethirolle:
@rdubrock, thanks for the review. Done!

rdubrock (approved):

PR #472, opened 2022-07-11
fix: change the way we compute uptime to avoid going over 100%

mem:
Tweak the current expression to avoid situations where a 1 is becoming a
2 because of rounding. Also tweak the expression to make it a little
simpler.

Fixes #


jewbetcha:
LGTM 👍

PR #478, opened 2022-09-29
Upgrade to 9

rdubrock:


mem (approved):
My only question is whether this is going to be compatible with the current version (non-beta, 9.1?)

rdubrock:
> My only question is whether this is going to be compatible with the current version (non-beta, 9.1?)

It's a good question, and I need to do some testing to make sure (with v8 as well)

PR #480, opened 2022-10-04
feat: minor ux tweaks to setup page

torkelo:
Noticed this strange orange link that goes to grafana.com, never seen an orange link in Grafana so did some minor tweaking to this link. Made links in the text instead and an extra one in the field description. 

Before:
Screenshot from 2022-10-04 08-22-18

After:
Screenshot from 2022-10-04 08-21-38

### DIFF
--- a/src/components/TenantAPISetupForm.tsx
+++ b/src/components/TenantAPISetupForm.tsx
@@ -25,6 +25,14 @@ const TenantAPISetupForm = ({ onSubmit, submissionError }: Props) => {
     defaultValues: { apiHost: DEFAULT_API_HOST, adminApiToken: '' },
   });
 
+  const renderAPIKeyLink = (text: string) => (<a
+    className="external-link"
+    href="//grafana.com/profile/api-keys"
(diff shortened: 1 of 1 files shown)


src/components/TenantAPISetupForm.tsx
```diff
@@ -33,23 +41,11 @@ const TenantAPISetupForm = ({ onSubmit, submissionError }: Props) => {
...
             url={'https://grafana.com/grafana/plugins/grafana-synthetic-monitoring-app/'}
           >
-            <p>
-              To initialize the App and connect it to your Grafana Cloud service you will need a Admin API key for you
-              Grafana.com account. The <b>API key</b> is only needed for the initialization process and will not be
-              stored. Once the initialization is complete you can safely delete the key.
-              <br />
-              <br />
-              <a
-                className="highlight-word"
```
torkelo, line 43: there are more instances of this className used for links that would be good to replace with external-link (if they are links)

rdubrock (approved):

PR #479, opened 2022-10-03
Handle topnav

rdubrock:
This adds some new routing logic to handle the new top navigation with breadcrumbs. Some examples of how this looks:

Screen Shot 2022-10-03 at 11 40 47
Screen Shot 2022-10-03 at 11 43 44
Screen Shot 2022-10-03 at 11 43 36
Screen Shot 2022-10-03 at 11 41 52

PR #481, opened 2022-10-04
More topnav fixes

torkelo:
based on https://github.com/grafana/synthetic-monitoring-app/pull/479
* Fixes some instances of flashing empty page with loading spinner (more of those to fix)
* Removes empty div wrapping Switch (caused issues for Page height) 
* Fixes some double headings (maybe there is a cleaner way to reduce duplication of those)
* Fixes max heigh 100% on HorizontalGroup causing all the card items to be pushed outside page  (new issue with HorizontalGroup when a direct child of PluginPage, the height 100% is now actually having an affect)

### DIFF
--- a/src/components/Routing.tsx
+++ b/src/components/Routing.tsx
@@ -65,37 +65,35 @@ export const Routing = ({ onNavChanged, meta, ...rest }: AppRootProps) => {
   }, [page, navigate, queryParams]);
 
   return (
-    <div>
-      <Switch>
-        <Route exact path={`${PLUGIN_URL_PATH}${ROUTES.Redirect}`}>
-          <DashboardRedirecter />
-        </Route>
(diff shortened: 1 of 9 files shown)


src/components/Routing.tsx
```diff
@@ -65,37 +65,35 @@ export const Routing = ({ onNavChanged, meta, ...rest }: AppRootProps) => {
   }, [page, navigate, queryParams]);
 
   return (
-    <div>
```
torkelo, line 68: this change is just removing this div

src/plugin.json
```diff
@@ -13,9 +12,18 @@
...
+      },
+      {
+        "name": "HTTP Dashboard",
+        "path": "img/screenshot-dash-http.png"
+      },
+      {
+        "name": "Check List",
+        "path": "img/screenshot-check-list.png"
+      }
     ],
```
torkelo, line 27: just prettier doing it's thing

Issue #486, opened 2022-10-20
The check timeout validation error message is sometimes incorrect


rdubrock:
**What happened**:
The hard limit for a timeout in probe options are 10 seconds. However when adding a value greater than 30 seconds it displays an error saying that the value cant be higher than 30 seconds even though the limit is 10. Adding any value here higher than 10 seconds automatically sets it back to 10.

PR #488, opened 2022-11-07
Fix: fix error for non-traceroutes; clear error if check type changes

CatPerry:
1) Traceroute checks have a default and immuatble timeout of 30 seconds; but all other checks should have an adjustable timeout from 1 to 10 seconds. Previously our error message misled users with non-traceroute checks to believe that they could adjust up to 30 seconds, which is inaccurate. So this PR fixes that to reflect the correct params.

2) Also fixed a bug that didnt clear the error messages from the slider input, even if the user changed their `checkType`

Fixes https://github.com/grafana/support-escalations/issues/4312
Fixes https://github.com/grafana/synthetic-monitoring-app/issues/486

Fixes #

Screenshot 2022-11-07 at 11 25 37 AM

Screenshot 2022-11-07 at 11 25 52 AM

### DIFF
--- a/src/components/CheckEditor/CheckEditor.tsx
+++ b/src/components/CheckEditor/CheckEditor.tsx
@@ -217,8 +217,9 @@ export const CheckEditor = ({ checks, onReturn }: Props) => {
           {submissionError && (
             <div className={styles.submissionError}>
               <Alert title="Save failed" severity="error">
-                {`${submissionError.status}: ${submissionError.data?.msg?.concat(', ', submissionError.data?.err ?? '') ?? 'Something went wrong'
-                  }`}
+                {`${submissionError.status}: ${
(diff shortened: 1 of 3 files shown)


src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -217,8 +217,9 @@ export const CheckEditor = ({ checks, onReturn }: Props) => {
           {submissionError && (
             <div className={styles.submissionError}>
               <Alert title="Save failed" severity="error">
-                {`${submissionError.status}: ${submissionError.data?.msg?.concat(', ', submissionError.data?.err ?? '') ?? 'Something went wrong'
-                  }`}
+                {`${submissionError.status}: ${
```
CatPerry, line 220: Just a linter fix

rdubrock (approved):
🦙

PR #501, opened 2022-12-22
chore: take an initial stab at the data shape of multihttp settings

rdubrock:
This is the data model for multihttp settings at the moment. I'm wondering if it might be better to use the types from `har-to-k6` directly, because they are almost the same. The only problem being that har-to-k6 is handling a lot more than our use case so it's not a perfect match.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,6 +1,7 @@
 import { DataSourceSettings, OrgRole, SelectableValue } from '@grafana/data';
 import { LinkedDatasourceInfo } from './datasource/types';
 import { SMDataSource } from 'datasource/DataSource';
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -546,3 +547,92 @@ export interface AdHocCheckResponse {
...
+  value?: string;
+  expression?: string;
+}
+
+export interface MultiHttpEntryRequest {
+  method: string;
+  url: string;
+  httpVersion?: string;
+  headers?: MultiHttpRequestHeader[];
+  queryString?: MultiHttpRequestQueryString[];
```
rdubrock, line 611: @CatPerry  check out the query string location here

PR #507, opened 2023-01-11
Adds multi-http via k6 api to Synthetics UI

CatPerry:

This PR adds multi-http checks to the Synthetics UI. 

It does so via an entirely different form because the K6 API that this new check uses is unrelated to blackbox exporter, which several of the other checks use in the CheckEditor form.
Screenshot 2023-01-30 at 4 11 35 PM

The user can add multiple http requests to a single check
Screenshot 2023-01-30 at 4 18 37 PM

 There is some basic form validation, but we may need to add more robust validation, depending on how K6 performs in the wild.
Screenshot 2023-01-30 at 4 12 16 PM

Fixes #


src/components/MultiHttp/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,308 @@
...
+                      <HorizontalGroup key={field.id} spacing="md" align="center" className={styles.headersQueryInputs}>
+                        {/* TODO: MAKE A SELECT COMPONENT INSTEAD */}
+                        <Input
+                          {...register(`${headersNamePrefix}.name` as const, {
+                            required: true,
+                          })}
+                          type="text"
+                          placeholder="name"
+                          disabled={!isEditor}
+                          onMouseLeave={() =>
```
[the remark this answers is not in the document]
rdubrock: I'm guessing it relates to validation? I think maybe we're fighting the form a bit

src/components/MultiHttp/consts.ts
```diff
@@ -0,0 +1,63 @@
...
+  alertSensitivity: AlertSensitivity.None,
+  settings: {
+    multihttp: fallbackSettings(CheckType.MULTI_HTTP) as MultiHttpSettings,
+  },
+  basicMetricsOnly: true,
+} as Check;
+
+const removeEmptyFields = (getValues: any) => {
+  const entries = getValues().settings.multihttp;
+  getValues().settings.multihttp.entries.forEach((ent: MultiHttpEntry) => {
```
rdubrock, line 24: I feel like it's worth digging into how we're using the form to make sure this is necessary. If it is, then 🤷‍♂️

src/components/constants.ts
```diff
@@ -373,3 +396,38 @@ export function fallbackSettings(t: CheckType): Settings {
 }
 
 export const PLUGIN_URL_PATH = '/a/grafana-synthetic-monitoring-app/';
+
+export const methodOptions = [
```
rdubrock, line 400: SInce this is a constant now let's change it to SCREAMING_SNAKE_CASE

src/validation.ts
```diff
@@ -207,11 +211,14 @@ export function validateSettings(settings: Settings): string | undefined {
   if (!settings[checkT]) {
     return 'Settings values required';
   }
-
+  // WHAT IS THIS FUNCTION FOR??
```
rdubrock, line 214: VALIDATING 😆

rdubrock: As you can see, there's a lot of things we _don't_ validate on the frontend (but we probably _should_)

rdubrock (commented):
So this is looking pretty good. My takeaways are that: 
- I think we're fighting how the form wants to work, and I think that's because we're sort of stuck between completely separating multi-http from the existing check editor or completely integrating with it
- I think it's worth buzzing this by the UX team to see how they feel about it.

CatPerry:
> So this is looking pretty good. My takeaways are that:
> 
> * I think we're fighting how the form wants to work, and I think that's because we're sort of stuck between completely separating multi-http from the existing check editor or completely integrating with it
> * I think it's worth buzzing this by the UX team to see how they feel about it.

I talked about UX early on in the process and was told we should just use the current UI for multi-http, and make UX adjustments down the road. So for the MVP, a month into building, it doesnt feel like the right time to talk to UX? Plus a buzz by seems unlikely; it'll be more like an avalanche, as there are lots of things to refresh in the SM UI.  We can do an iteration of the UI after the MVP, at the beginning of the process, no?

As for forms inside of forms, I did that to keep the user's selection of checktype and "Enable metrics and logs". And the Multi-http form doesnt seem to affect the Check Editor parent. Is this a dealbreaker for MVP? @rdubrock

src/page/CheckRouter.tsx
```diff
@@ -50,6 +52,12 @@ export function CheckRouter() {
         <Route path={`${path}/edit/:id`} exact>
           <CheckEditor onReturn={returnToList} checks={checks} />
         </Route>
+        <Route path={`${path}/type`} exact>
+          <ChooseCheckType onReturn={returnToList} checks={checks} />
+        </Route>
+        <Route path={`${path}/new-multi`} exact>
```
rdubrock, line 58: Might make sense to use a param variable here. So like `${path}/:checkType`. Then in the form you can grab the `useParams` hook and pull the value of the check type out of the url

CatPerry: So to thought if that second part, but the reason I didnt pass the checkType as a query param is because the current value stored in `CheckFormValues['checkType']` looks like this `{label: 'MULTI-HTTP', value: 'multihttp'}`. So the current function `selectableValueFrom` wouldn't work right, and I could write another onemaking an exception for the label `MULTI-HTTP'` because it's CheckType is snake case, not kebab. Anywho, that didnt feel particularly perfect in comparison , since the state exists on `location` so you can pass and access these sorts of things. 

But the first part your comment to adjust the routing altogether to `${path}/:checkType` I'm fine with doing, but I didnt want to be presumptuous. We'll have to adjust all event tracking I think as a result.

rdubrock: Yeah maybe let's pair on it for a minute today. I think there might be an easy way to reconcile all that

CatPerry:
> I tested locally and I can't find anything leaking past the feature flag, so I think it's all good! We can fix any follow ups in their own PRs.
> 
> _Really_ good job on this. I know it's been a slog but folks are going to be stoked on this

Eeeeeeee! Yayayayaya!

PR #509, opened 2023-01-12
fix: check test runs should be limited to 5 probes

rdubrock:
Screenshot 2023-01-12 at 09 12 59

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -305,5 +305,5 @@ export function getRandomProbes(probes: number[], quantity: number): number[] {
     const index = Math.floor(Math.random() * probes.length);
     randomProbes.add(probes[index]);
   }
-  return Array.from(randomProbes).sort();
+  return Array.from(randomProbes).sort((a, b) => a - b);
 }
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -295,3 +295,15 @@ export const getSuccessRateIcon = (
...
+export function getRandomProbes(probes: number[], quantity: number): number[] {
+  if (quantity >= probes.length) {
+    return probes;
+  }
+  const randomProbes = new Set([] as number[]);
+  while (randomProbes.size < quantity) {
+    const index = Math.floor(Math.random() * probes.length);
+    randomProbes.add(probes[index]);
+  }
+  return Array.from(randomProbes).sort();
```
CatPerry, line 308: What kind of sorting are you expecting here? I assume it's not a real sort. So why call sort then?

CatPerry (approved):
a test! great!

src/utils.ts
rdubrock: It's an array of integers, so I'm just expecting the default

PR #516, opened 2023-02-06
chore: remove worldmap dependency

rdubrock:
We haven't used the worldmap panel for over a year, but we had to keep the dependency around for users on older grafana instances (I think? The reason has been lost to the sands of time). The time window for supporting that switch has long passed, I think, so it's time to remove it completely.


rdubrock:
Let's not merge this for now, there's a larger conversation around how to handle getting rid of worldmap panel

rdubrock:
We actually can't do this yet

PR #518, opened 2023-02-07
Various multi http cleanup

rdubrock:
Fixed up the broken tests, did some type cleanup, streamlined the routing a touch. The form was getting double instantiated, and that was creating a bit of weirdness so I went after that as well. 

I also moved the validation triggers into the form itself instead of triggering with mouse events. I _think_ it's all working as intended, but double check me!


rdubrock:
I also synced with main so the build could pass

CatPerry (approved):

PR #530, opened 2023-03-30
Add synthetics scenes

rdubrock:
Experimenting with scenes. It's mostly complete, the traceroute node graph isn't quite working yet. 

Everything is gated behind a feature flag, and on a route not present in the navbar, so I think it's safe to merge if we want. Turn on `synthetics-scenes` and go to <plugin url>/scene to test.


package.json
```diff
@@ -95,9 +95,11 @@
   },
   "dependencies": {
     "@emotion/css": "^11.1.3",
-    "@grafana/data": "9.2.1",
-    "@grafana/runtime": "9.2.1",
-    "@grafana/ui": "9.2.1",
+    "@grafana/data": "canary",
```
rdubrock, line 98: Hmm I need to think about how to deal with this for now.  I think maybe we don't update the dependencies so we definitely aren't breaking anything, and if we need to do dev on scenes we modify the package.json each time

CatPerry (approved):
Wowza 👍🏾  ... 
just a nit or two, nothing blocking; lgtm

PR #531, opened 2023-04-03
Bump SM app to v1.9.40

CatPerry:
Bump Synthetic Monitoring to v1.9.4 since multi-http feature was merged to main

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "synthetic-monitoring-app",
-  "version": "1.9.36",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "synthetic-monitoring-app",
-  "version": "1.9.36",
+  "version": "1.9.40",
```
rdubrock, line 3: I think you could go `1.10.0` if you wanted to count this as a feature instead of a patch

PR #536, opened 2023-04-12
Revamp dev setup

rdubrock:
Migrates to latest create-plugin version and gets rid of our bash scripts

- Hot reloading now, yay!
- _Hopefully_ simplified and standardized dev setup


jewbetcha (approved):
LGTM. Happy live reloading 🥵

PR #548, opened 2023-05-15
Viz view as scene

rdubrock:
Replacing the viz view of the check list with a scene. The check list code is getting a little creaky and needs a refactor, but that's for another day.

I had to hide all the filtering/sorting and bulk actions. It doesn't super make sense for this view, and I could add some of that functionality back through dashboard controls but this seemed like a good first step. 

I did check and the upper limit on number of stats to display is 5000, so I think this will support lots of checks better than our current situation

I don't _love_ that the check job is basically illegible, but I think the threshold color and sorting is what users want from this view. We originally made it so people could get a really quick "up/down" view of everything without having to scroll

Screenshot 2023-05-15 at 10 14 16


jewbetcha:
Nicely done 💪 I'm seeing a bunch of filters and stuff in the code but not in the screenshot, are they only conditional if scenes aren't enabled?

rdubrock:
> I'm seeing a bunch of filters and stuff in the code but not in the screenshot, are they only conditional if scenes aren't enabled?

Yeah, the scene fetches it's own data, and all the filters and whatnot are operating on data passed down through a context that the rest of the app uses. So the solution for now is to just hide all the filtering when looking at the scene

jewbetcha (approved):

torkelo:
> I don't love that the check job is basically illegible, but I think the threshold color and sorting is what users want from this view. We originally made it so people could get a really quick "up/down" view of everything without having to scroll

you can set a fixed font size for title

torkelo:
don't merge this yet, I have another layout / idea proposal here

torkelo:
It feels like this is not a good replacement for the checks view (As it looks now)

* Repeat a layout 
I would try to create a query/data frame response that results in a DataFrame per check. Then use the [ByFrameRepeater to repeat a layout](https://github.com/grafana/grafana/blob/main/public/app/features/scenes/apps/scenes.tsx#L145) of 2-3 panels (for each check). 

The VizPanel supports headerActions which you can use for links / drilldown buttons in the panel frame top right corner. 

VizPanel also supports a custom Viz panel, which you could use to show the information that the current one is showing (active series, frequency etc). 

Example (But here we have a graph and 1 stat panel layout being repeated for each frame). 
image

The layout toggle could we can add back later to change how the layout is built / options for the panels.

torkelo:
ah, so this is only repacing this 
image

torkelo:
There is also an option panel option `"textMode": "name",`  which makes the stat panel only show the name, the percent value will be shown in a tooltip 

image

src/components/CheckList/CheckListScene.tsx
```diff
@@ -0,0 +1,196 @@
...
+    to: 'now',
+    from: 'now - 6h',
+  });
+
+  return new EmbeddedScene({
+    $timeRange: timeRange,
+    controls: [
+      new SceneReactObject({
+        //@ts-ignore not sure what's going on here, but the types are unhappy for some reason
+        component: CheckListViewSwitcher,
```
dprokop, line 83: What's the TS error here?

src/components/CheckList/CheckListScene.tsx
```diff
@@ -0,0 +1,196 @@
...
+    ],
+    body: new SceneFlexLayout({
+      children: [
+        new SceneFlexItem({
+          width: '100%',
+          height: '100%',
+          body: new VizPanel({
+            pluginId: 'stat',
+            title: '',
+            $data: transformed,
```
dprokop, line 112: Agree with @torkelo that using the repeater here and showing it in a different way than stat tiles may be better experience. Also this will allow you to control `minHeight` of each individual viz making it behave better on mobile devices (not rish of an element getting completely squshed)

torkelo: I think this stat panel view is only aiming to replace this honeycomb view. But for the main checks list, a repeater that repeats a layout is definitely the solution.

image

rdubrock:
> ah, so this is only replacing this

Yeah, at the moment this was just to replace the honeycomb up/down view. I like the idea of the repeater, I'll see if it makes more sense to use that to replace the other views or whether it will work for this case as well. 

I'll play around with just showing the name to make things more legible. Thanks for the reviews!

rdubrock:
Screenshot 2023-05-17 at 10 05 42

I think this is the final answer for now

PR #584, opened 2023-07-26
fix: error rate maps not filtering by probes

rdubrock:


src/scenes/Common/errorRateMap.ts
```diff
@@ -17,7 +17,7 @@ function getErrorMapQuery() {
         )
       /
         sum by (probe, geohash) (
-          rate(probe_all_success_count{instance="$instance", job="$job"}[$__range])
+          rate(probe_all_success_count{instance="$instance", job="$job", probe=~"$probe"}[$__range])
```
samjewell, line 20: Why not also make the same change to the `sm_check_info` selector on lines 15 (above) and 25 (below)?

To: `sm_check_info{instance="$instance", job="$job"}`

I think this would be more consistent, and also have better performance (although the performance gain is probably negligible - more just nice to follow best practices maybe?).

rdubrock: Good call, I'll add that in another PR

PR #595, opened 2023-08-28
chore: removed grafana-worldmap-panel dependency in plugin.json

xnyo:

Removes `grafana-worldmap-panel` from the list of dependant plugins in plugin.json.

`grafana-worldmap-panel` is deprecated, and it was replaced with `geomap` in #398

However, the dependency in `plugin.json` was not removed. This will cause Grafana to attempt to install `grafana-worldmap-panel` when installing `synthetic-monitoring-app`, even if it's not required anymore.

This is a problem because `grafana-worldmap-panel` uses Angular, which is being removed and will causes issues when installing `synthetic-monitoring-app` if [Angular is disabled](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#angular_support_enabled) in Grafana via config (which will also be the default in the future)

Fixes #

### DIFF
--- a/dev/provisioning/plugins/example.yaml
+++ b/dev/provisioning/plugins/example.yaml
@@ -1,8 +1,5 @@
 # apiVersion: 1
 
-# panel:
-#   - type: grafana-worldmap-panel
-#     disabled: false
 # apps:
 #   - type: grafana-synthetic-monitoring-app
 #     name: grafana-synthetic-monitoring-app
--- a/src/plugin.json
+++ b/src/plugin.json
@@ -83,14 +83,6 @@
   ],
   "dependencies": {
     "grafanaDependency": ">=8.3.3",
-    "grafanaVersion": "8.3",
-    "plugins": [
-      {
-        "type": "panel",
-        "name": "Worldmap Panel",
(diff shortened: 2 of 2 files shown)


rdubrock:
Hey, thanks for the PR! We tried to remove the dependency before, but it ended up causing problems. The reason goes like this:

- When users activate synthetic monitoring, some dashboards are generated
- Ever since geomap came out, those dashboards use geomap panel, however, every stack that activate _before_ geomap existed has dashboards that use worldmap panel. 
- Users that have old dashboards can upgrade with a click, but in practice they rarely do

I can't remember whether this resulted in a broken dashboard (which is fine, and we can merge this change), or whether it actually crashed grafana itself (which is less fine, and will take some other approach). I'll look around

sympatheticmoose:
Thanks for taking a look @rdubrock - as giuseppe says we're looking to disable angular by default for new stacks, so we need to make this change as soon as possible so that it doesn't cause issues.

rdubrock:
Sure, in my tests it wasn't crashing Grafana so I'm happy to merge, but @xnyo and @sympatheticmoose please do keep an eye out

rdubrock (approved):

PR #597, opened 2023-08-29
fix: update URL for worldping project

mattsimonsen:
Updated URL to point the worldping project, rather than a 404 at grafana.net


rdubrock (approved):
Thanks for the PR!

PR #609, opened 2023-09-14
Add oauth2 fields behind a feature flag

rdubrock:


src/components/http/HttpSettings.tsx
```diff
@@ -318,7 +334,127 @@ export const HttpSettingsForm = ({ isEditor }: Props) => {
...
+                    })}
+
+                    <Button
+                      type="button"
+                      icon="plus"
+                      variant="secondary"
+                      size="sm"
+                      disabled={!isEditor}
+                      onClick={() => {
+                        console.log('appending scope');
```
ckbedwell, line 444: Left in by mistake? 😃

ckbedwell (changes requested):
I assume the backend isn't set up for this yet as I could see it in the request but didn't return any of the options I had added.

I left a comment about what my expectation of the "Use Oauth2" checkbox should do after you untick it but when it is ticked my other expectation is that some of the fields should be required at that point but they can all be sent empty?
 
Screenshot of Synthetic Monitoring plugin authentication accordion for http checks. The inputs are very  close to each other with not enough margin.

The UI is pretty cramped, could do with a bit of margin-bottom to separate the forms from the tickboxes.

rdubrock:
Yeah this one is in limbo and I've been meaning to come back to it when I carve out some time. The backend supports it but might require some changes, and I need a good test case

rdubrock:
Closing this for now, this got a little too stale

PR #625, opened 2023-10-12
Fix: Update Grafana runtime dependencies

ckbedwell:
# Problem
Grafana as a user experiences it in the browser !== how our development environments experience it, which doesn't build the best trust when relying on your IDE and tests for accurate results.

# Solution

Updating all the Grafana packages to the latest (or the latest publicly available via npm...) so it is more representative of the browser experience and updating any of the new type errors.

## Small housekeeping
- Deleted `yarn.lock` and did a fresh install to regenerate it. It resolved quite a few dependency issues for me so I could remove the previously introduced `resolutions` object in `package.json`.
- I know the grafana plugin devs don't want us to touch `./config/webpack/webpack.config.ts` and do our changes in the plugin webpack file but it would be such an over the top and verbose change for simply correcting the `baseUrl` to an absolute path that it seemed the best option to edit it directly.
- Reconciled `.eslintrc` and `.eslintrc.js` so local dev environment stops complaining. Removed the duplicated rules and configs already provided by `@grafana/eslint-config`.
- Updated node version to 18 in `.nvmrc` and `.github/workflows/is-compatible.yml` as one of the newly introduced dependencies from the grafana packages requires it.

## New icon button tooltips
[Grafana 10.1.0 introduced a breaking change for `IconButton`](https://github.com/grafana/grafana/blob/main/CHANGELOG.md#1010-2023-08-22) so that it either needs a tooltip or aria-label to be considered correctly implemented. Where they were missing I've added `tooltip`s and changed a few `aria-labels` to tooltips, too. Not all users may be familiar with what particular icons do, so it helps with accessibility that they can check before choosing to make an interaction.

Screenshot of Synthetic Monitoring plugin with new View Dashboard tooltip for appropriate icon.

Screenshot of Synthetic Monitoring plugin with new Delete tooltip for appropriate icon.


.config/webpack/webpack.config.ts
```diff
@@ -81,7 +81,7 @@ const config = async (env): Promise<Configuration> => ({
           loader: 'swc-loader',
           options: {
             jsc: {
-              baseUrl: './src',
+              baseUrl: path.resolve(process.cwd(), SOURCE_DIR),
```
rdubrock, line 84: The reason they don't want changes here is because these files get autogenerated and overwritten by the update command: https://www.npmjs.com/package/@grafana/create-plugin#update-your-plugin-build-config

So even if it's a pain, I think it's probably worth putting the customization in the other webpack.config file

.eslintrc.js
```diff
@@ -1,16 +1,9 @@
...
-    'plugin:import/warnings',
-    'plugin:import/typescript',
-  ],
+  extends: ['./.config/.eslintrc', 'plugin:import/errors', 'plugin:import/warnings', 'plugin:import/typescript'],
   rules: {
     // Since TS shows errors for unresolved imports anyway,
     // we can safely turn this off. (Plus, eslint doesn't play
     // nice with TS' baseUrl out of the box.)
     'import/no-unresolved': 'off',
-    'react-hooks/exhaustive-deps': 'error',
```
rdubrock, line 14: Not a fan of this rule?

src/components/NameValueInput.tsx
```diff
@@ -50,7 +50,7 @@ export const NameValueInput = ({ name, disabled, limit, label, validateName, val
             `}
           >
             <Input
-              {...register(`${name}.${index}.value` as const, { required: true, validate: validateValue })}
```
rdubrock, line 53: Did the validation logic move somewhere else? I don't think we want to delete it if not

rdubrock (changes requested):
Looks good! I think some extra things got deleted, and moving the config to the plugin webpack config is important (unless the baseUrl change is specific to this version of `create-plugin`)

.eslintrc.js
ckbedwell: It comes with [@grafana/eslint-config by default](https://github.com/grafana/eslint-config-grafana/blob/master/index.js#L84) :)

src/components/NameValueInput.tsx
ckbedwell: Oh, oops! Good catch! I meant to just remove the `as const` assertion. I couldn't see the need for it?

***
_Tangential rant..._

I hate the 120 character line limit we have across Grafana repos and if we used the default [80 as Prettier recommends](https://prettier.io/docs/en/options.html) I probably would have noticed everything after the assertion.

It [wasn't as chaotic](https://github.com/grafana/grafana-k6-app/pull/672) as we worried it was going to be in the k6 plugin 😅

rdubrock: Yeah I remember that being annoying to me too. It's been so long I've completely forgotten. If you want to change it to 80, sally forth.

rdubrock (approved):
Nice!

PR #631, opened 2023-10-19
Add multihttp to terraform

rdubrock:
Makes the "Generate Config" button in the Terraform section of the plugin config page handle MultiHTTP checks

### DIFF
--- a/src/hooks/useTerraformConfig.test.tsx
+++ b/src/hooks/useTerraformConfig.test.tsx
@@ -244,7 +244,6 @@ describe('terraform config generation', () => {
     await waitFor(() => {
       expect(result.current.config).not.toBeUndefined();
     });
-    console.log(result.current.config);
     expect(result.current.config).toEqual({
       provider: {
         grafana: {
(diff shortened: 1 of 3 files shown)


src/hooks/useTerraformConfig.test.tsx
```diff
@@ -193,4 +193,159 @@ describe('terraform config generation', () => {
...
+      };
+      const meta = {} as AppPluginMeta<GlobalSettings>;
+
+      return <InstanceContext.Provider value={{ instance, loading: false, meta }}>{children}</InstanceContext.Provider>;
+    };
+    const { result } = renderHook(() => useTerraformConfig(), { wrapper });
+    await waitFor(() => {
+      expect(result.current.config).not.toBeUndefined();
+    });
+    console.log(result.current.config);
```
ckbedwell, line 247: Accidentally left in? 😃

ckbedwell (approved):
Does any of this only want to be enabled when the multi-http feature flag is on?

Otherwise after the console.log is removed from the test LGTM 💪

src/hooks/useTerraformConfig.test.tsx
rdubrock: oopsy doopsy...

rdubrock:
> Does any of this only want to be enabled when the multi-http feature flag is on?

I think not, just because there could be a situation where the plugin has the flag off, but they've already got a multihttp check (like, maybe it got added by the api directly or something). In that case it makes sense to just give them the config for the checks they already have created

PR #636, opened 2023-10-31
Fix: Multihttp tabs styling and open first request tab with error

ckbedwell:
**Before**

- Field / Error misalignments.

image

- Request header non-keyboard accessible

- Difficult to spot errors on save

https://github.com/grafana/synthetic-monitoring-app/assets/6906380/7a9dfbf4-0d17-4bcf-a76a-b304497cc03e

**After**

- Aligned fields / errors.

image

- Keyboard-accessible request header

image

- Automatically opens the first collapsed request header and goes to the first tab with errors. One known issue with this implementation is that @grafana/ui does not forward the ref to the Select in a react-hook-form Controller, so it doesn't automatically focus some fields. I'll put another PR up tomorrow to address this as I know another workaround that is a little bit involved.

https://github.com/grafana/synthetic-monitoring-app/assets/6906380/3abe4097-955d-4651-a7f1-40b51dd79ebd

### DIFF
--- a/src/components/MultiHttp/MultiHttpCollapse.tsx
+++ b/src/components/MultiHttp/MultiHttpCollapse.tsx
@@ -30,11 +30,9 @@ export const MultiHttpCollapse = ({
           onToggle();
         }}
       >
-        <>
-          <Icon name={isOpen ? 'angle-down' : 'angle-right'} className={styles.headerIcon} />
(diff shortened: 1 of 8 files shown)


rdubrock (commented):
Looks great!

rdubrock (approved):

PR #639, opened 2023-11-02
Feat/scroll and focus multihttp errors

ckbedwell:
This is to make the user experience a little bit nicer when dealing with the MultiHttp form. When setting up multiple URLs it could be easy to miss some of the validation as you are creating multiple requests, which each have several tabs, which each have several fields.

This change makes it so the viewport scrolls to the first URL panel with an error and automatically focuses the first error field.

https://github.com/grafana/synthetic-monitoring-app/assets/6906380/10c20c93-7add-4af5-b102-b19b453e4625

**Notes to the reviewer:**

There's a few undesirable solutions going on to address this behaviour correctly but I think they are worthwhile for this functionality:

1. Because we use custom Selects and Inputs from @grafana/ui, often the refs created by react-hook-form don't get passed down correctly to the input we would like focussed. I had a look to see if there was anyway of addressing this but it was too involved and overly complicated.

2. Because of the above I've had to remove the automatic error focussing that `react-hook-form` provides. On the plus side it has given us the opportunity of adding a nicer scroll to error rather than the default 'snapping' which can disorientate users.

3. In the `focusField` function we have to check what type of input it is to know whether we are on the right input or we have to do a bit of dom traversal to find the user-facing input.

_I've targeted the import order linting PR to minimise diff noise and mitigate any potential merge conflicts. If that PR doesn't go in I can easily recreate this._


rdubrock (approved):
Nice! It looked good enough yesterday to merge, but this is even snazzier.

rdubrock:
FWIW the Select in `@grafana/ui` has been a source of _much_ annoyance over the years. It's `react-select` under the hood, and it's totally opaque for testing as well. I've ended up just mocking it for all the tests.

PR #647, opened 2023-11-13
Add scripted checks

rdubrock:

Fixes #


rdubrock:
@ckbedwell I'm interested in your thoughts on how to proceed here. The feature is obviously very much a WIP, but as you can see the PR is already getting too big for it's britches. I'm inclined to merge it since it's behind a feature flag, and then bite of smaller pieces in more manageable PRs

mem (approved):

ckbedwell (approved):
@rdubrock I agree, let's get this in and then we can review it in full when it is nearing release-ready. There's a lot to digest in this PR and I've looked through most of it but I don't think there is any need to hold this up 😃

PR #648, opened 2023-11-15
chore: add beginnings of a scripted checks scene

rdubrock:
Adds a scripted checks scene, and a new check list. I think what I'm angling for here is actually to subsume the entire check list instead of making a separate page for scripted checks.

Beginning of the list:

Screenshot 2023-11-20 at 14 03 54

Beginnings of the data viz:
Screenshot 2023-11-20 at 14 04 01


ckbedwell (commented):
I'll do a more in-depth code review when you are nearing the end of this and the UI/UX is finalised. It's looking good! I've got some thoughts in general that are worth thinking about:

## Checks overview

Synthetic Monitoring plugin on the checks page with a new table as per this PR.

I don't love the positioning of the bulk actions, especially as a keyboard user (and the edit actions in the dropdown aren't keyboard accessible). It might be worth having a look how we do it in k6 for bulk selecting multiple test runs. The only bulk action we have is mass deletion but you could move all of the actions to the dropdown, I actually find it a bit confusing that some actions are in the dropdown and some aren't. I'd argue that enable / disabling is still editing the check. I'd also like to see individual actions in the last column (the same as k6) for at least editing and deleting.

I won't take up this fight here but there are also some considerations for adding the primary action as an obvious button choice as well as allowing row on click as you have it now. I'm not too familiar with the table you've used so not sure what limitations there are. We had a lot of conversations in k6 FE about table row clicks being difficult to decipher at times. Some users expect them, some don't and get frustrated looking for a button to click to perform their desired action.

I'd definitely like a confirmation modal for enable, disable and deletion actions, too! One misclick away from a very bad day on the bulk deletion side!

I think the table is missing some key information, enabled / disabled and frequency In particular.

## Checks dashboard

I don't know if there is a Scenes bug or there's some unstable reference with the way filters get passed down that is forcing full page re-renders but the render flickering when you use any of the filters is pretty disorientating! It's not a major issue but it is pretty noticeable.

I am in two minds about the edit drawer here:
1. I'm not sure how common a pattern this is across Grafana? Not that I think that should necessarily stop us but would be interested in if any other apps do something similar. I get the convenience of having this for doing quick edits, however...
2. My big gripe is that you could spend 5 minutes filling that form in to edit it and if you hit escape (which for keyboard users is super easy to do -- especially because of a bug in behaviour for dropdowns inside drawers / modals in `grafana/ui`) or accidentally click out of the drawer, all of that information is lost. In the least I'd look at a way to preserve it so a user could pick up where they left off (maybe some tricksy sessionStorage save onDismiss?).

I also noticed a bug that when you open the drawer the breadcrumbs change and that multihttp values don't populate.

I would be tempted to keep the edit on its own page, I'm thinking it's likely a user might want to go directly to edit from the checks table and it could be confusing either having two different places / patterns how you edit checks or being taken to the overview and the drawer opens? You could go the other way and bring in some of the overview information to the edit page! If it followed the probes layout, it would go where the 'supporting content' is rendered.

Let me know your thoughts!

## Scripted checks

I thought it was interesting you use `k6` in the route. Makes me think if we should make it clearer to the user (with the k6 logo or some such) that these use k6 under the hood or are k6-powered in the ui more? Not sure if product have provided much direction on that yet and what documentation we can provide users?

Synthetic Monitoring plugin on the new scripted checks page two uneven cards as per this PR.

These cards are uneven sizes on my ultra-wide, would be nice to keep them equal size and maybe a bit more centred with a maximum width.

Synthetic Monitoring plugin on the script editor page with very wide inputs strecthing across an ultrawide monitor as per this PR.

Similar on the scripts pages themselves. We have a max width for the probes forms which works nicely, maybe put something similar in?

I'm also not sure about the floating save button as this brings in a new UI pattern which isn't used elsewhere in the app. I can see why you've done it because of the huge height of the Monaco editor but I think most users would know to scroll to the bottom to finalize and save a form. Also, where the save button is now in the DOM is very unpredictable with keyboard focus. If you are a keyboard user, you go through all of the form, then have to shift-tab right to the beginning to save it (or just _know_ to hit enter on any of the forms outside the Monaco editor to save). 

Also, on my ultrawide monitor, it isn't very noticeable.

Synthetic Monitoring plugin on the new scripted check page on an ultrawide monitor as per this PR.

## Routing

I think there's a good opportunity to rewire the routing for checks as you are doing this overhaul. It's a bit unpredictable as it stands and when using the breadcrumbs to navigate you don't really know where you are going to end back up at. I see the breadcrumbs as an alternative to using the back button and that's not what most of the behaviour does at the moment.

- When you are on `/checks/new/k6/script-editor` if you click on the breadcrumb "New check" you end up at creating a new ping check (I'd remove this default route, I find it confusing). My expectation would be to go back to the "choose scripted check" screen.

- If you are on any of the other checks there is no option to go to the choose scripted check screen from the breadcrumbs. You have to go back to checks then click through again.

I think `/checks/choose-type` should become `checks/new` and then every check type should come under this, e.g. `checks/new/http`, `checks/new/traceroute`, etc.

In short, I think the URL should reflect the breadcrumb structure. I'd expect to be able to both navigate predictably by clicking on breadcrumbs or removing parts of the URL and know where I'd end up.

Okay, hope this helps! I am more than happy to jump on a call and go through it together if that's easier or if you want to bounce some ideas off me. Let me know 😄

PR #663, opened 2024-01-03
feat: add an errors only toggle to logs panel

rdubrock:
Adds a toggle that filters multihttp/scripted log panels by errors:

Screenshot 2024-01-03 at 08 46 33
Screenshot 2024-01-03 at 08 46 21


mem (approved):
🥳

PR #678, opened 2024-02-09
Feat/react query

ckbedwell:
## Summary of changes:

- Remove all of the global contexts providing data in the app wrapper.
- Replaced all data-fetching for probes, checks and thresholds with react-query. This creates a deep coupling so that any mutations automatically invalidate the cache for the listings. It should be impossible for the listed state to get out of sync with the mutated state now. No more code smells of pausing and resuming data fetching and manually refetching stale lists after mutations.
- If a component needs check, probe or threshold data it should call the data-fetching hook inside it. `react-query` handles duplicated fetches so there isn't a worry about making redundant network calls. With these changes we are actually making **significantly less** network requests as components now have a global cache and know when data is fresh vs stale. 
- All of the toasts, error messaging and telemetry data (e.g. Faro) has been standardised so that it is handled at the querying level rather than rewriting the boilerplate each time it is used in a component.
- Hasn't been fully implemented in this PR but I've set-up a design pattern for handling network failures in the UI using suspense queries and error boundaries.

### Things that shouldn't be in this PR but came along for the ride

- I've refactored a large chunk of the `CheckList` page. I had to break it down into smaller components because at one point I was having some rendering issues and it was _very_ difficult to pinpoint where the problem lay in such large components.
- During that refactor I removed lots of unnecessary `useState`s and `useEffect`s as their values could be derived from other state and didn't need to be managed independently.
- I've simplified a lot of components and dumbed them down so they are more re-usable.
- I added extra sort options to the CheckList that I thought made sense. I removed sort by success because as a user I had no idea what that meant.
- Refactored the gauges so they are simpler and dumber and could more tightly call the hooks for the data they want to use rather than relying on it being provided from above.

I still think the CheckList page could do with a lot more love but I think these changes make it easier to dive in and start making changes.

### What definitely shouldn't be in this PR

- I noticed the Terraform config doesn't have an entry for converting k6 checks so was erroring out. This should be addressed in a separate PR but I had to validate my changes to fetching checks were working correctly so am returning nothing for k6 checks.

_Lots of failing tests. Creating a separate PR to add msw so there can be a long-term testing strategy for the api implemented._


src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -205,24 +189,22 @@ export const CheckEditor = ({ onReturn, checks }: Props) => {
...
           </form>
         </FormProvider>
       </>
+      {error && (
+        <div className={styles.submissionError}>
+          <Alert title="Save failed" severity="error">
+            {error.message ?? 'Something went wrong'}
+          </Alert>
+        </div>
+      )}
```
ckbedwell, line 201: I thought some of the alerts were displaying in strange places so I am trying to standardise them so they appear after all the submit buttons.

src/components/CheckEditor/CheckEditor.tsx
```diff
@@ -205,24 +189,22 @@ export const CheckEditor = ({ onReturn, checks }: Props) => {
...
       <ConfirmModal
         isOpen={showDeleteModal}
         title="Delete check"
         body="Are you sure you want to delete this check?"
         confirmText="Delete check"
-        onConfirm={onRemoveCheck}
+        onConfirm={onDelete}
```
ckbedwell, line 207: Small thing but I thought I'd start and unify some of the naming so we reference CRUD verbs.

src/components/CheckEditor/testHelpers.ts
```diff
@@ -18,11 +18,10 @@ export const toggleSection = async (sectionName: string, user: UserEvent): Promi
   return sectionHeader.parentElement?.parentElement ?? new HTMLElement();
 };
 
-export const submitForm = async (onReturn: (arg0: Boolean) => void, user: UserEvent) => {
+export const submitForm = async (user: UserEvent) => {
```
ckbedwell, line 21: I thought what `onReturn` was doing was a little odd (you'll see I've co-located it to the onSuccess methods and using the locationService.goBack method -- as there are multiple places you can edit a check from now) and I didn't like the prop-drilling where it was being continuously passed down.

src/components/CheckFilters.tsx
```diff
@@ -17,18 +16,14 @@ const getStyles = (theme: GrafanaTheme2) => ({
...
   verticalSpace: css`
     margin-top: 10px;
     margin-bottom: 10px;
-    margin-right: ${theme.spacing(2)};
   `,
 });
 
 interface Props {
-  handleResetFilters: () => void;
+  onReset: () => void;
```
ckbedwell, line 26: Also thought I'd try and standardise so that props have `on${Whatever}` and `handle${Whatever}` functions are defined within a component that need to add extra steps to `on${Whatever}`.

src/components/CheckHealth.test.tsx
ckbedwell, line 1: Wasn't used anywhere.

src/components/CheckListItem/CheckListItem.tsx
```diff
@@ -3,26 +3,23 @@ import { GrafanaTheme2 } from '@grafana/data';
...
 
 import { CheckCardLabel } from '../CheckCardLabel';
-import { LatencyGauge } from '../LatencyGauge';
 import { CheckItemActionButtons } from './CheckItemActionButtons';
 import { CheckListItemDetails } from './CheckListItemDetails';
 import { CheckStatusType } from './CheckStatusType';
 
 interface Props {
-  check: FilteredCheck;
+  check: Check;
```
ckbedwell, line 17: In another PR I want to fix the type for `Check`. It's _so_ confusing that the ID is optional and I am assume this is because we are lacking a `DraftCheck` type for check editors?

I feel like we had / have a bandaid of FilteredCheck is the real Check whilst Check is proxying as the non-existent DraftCheck?

src/components/Gauges/Gauge.tsx
ckbedwell, line 1: I've refactored the gauges so they all use a common component. I've also refactored the display logic because it was very confusing following the chain.

src/components/TerraformConfig/terraformConfigUtils.ts
```diff
@@ -142,6 +142,13 @@ const settingsToTF = (check: Check): TFCheckSettings => {
           }),
         },
       };
+    case CheckType.K6: {
```
ckbedwell, line 145: This needs a separate PR to be filled out correctly.

src/components/UsageStats/UsageStats.tsx
ckbedwell, line 1: Pulled this out into its own component from the homepage.

src/components/constants.ts
```diff
@@ -316,8 +316,20 @@ export const CHECK_LIST_SORT_OPTIONS = [
     value: CheckSort.ZToA,
   },
   {
-    label: 'Success',
-    value: CheckSort.SuccessRate,
+    label: 'Asc. Reachability ',
```
ckbedwell, line 319: I added these whilst refactoring as I thought they made a lot of sense. Should I add latency Asc / Desc, too?

src/datasource/DataSource.ts
ckbedwell, line 1: Typed this all out correctly 💪

src/hooks/useAlerts.ts
ckbedwell, line 1: Going to move this to `react-query` in a separate PR when I get to Alerting.

src/page/HomePage.tsx
```diff
@@ -211,10 +81,7 @@ export const HomePage = () => {
   }, [instance.api, scenesEnabled, multiHttpEnabled, scriptedEnabled]);
 
   if (perCheckDashboardsEnabled) {
-    if (!scene || loading) {
-      return <LoadingPlaceholder text="Loading..." />;
-    }
-    return <scene.Component model={scene} />;
+    return <SceneHomepage />;
```
ckbedwell, line 84: Pulled this out to its own component.

ckbedwell (commented):
Self-review.

rdubrock (commented):
Nice, looks very slick! There's enough here I don't think I can really meaningfully critique it. I think we can just roll with it and see how good our guardrails are

PR #686, opened 2024-02-15
Refactor and unify test data

ckbedwell:
## Problem

Synthetic Monitoring is growing in complexity and a lot of information displayed in the browser is based on compounded api calls and calculations from multiple end points (e.g. checks often need to know about probes to display correctly; checks and probes need to query metrics to display correctly).

Modern, durable and lasting testing practices are based on the idea we are asserting on the way users interact with our application rather than doing smaller and less manageable unit tests -- unit tests are focussed on implementation details and not on the end goal of if users can complete core user journeys and create a barrier to refactoring and paying down tech debt.

I recently did a large refactor to unify the data-fetching in the application by using React-query and I broke a lot of tests in the process. The problem is the tests were written in a semi-integration / semi-unit testing stance so the data they expected was hard-coded into the test and didn't account for what happens when the data changes.

## Solution

Create a universal set of fixtures that **_all_** tests utilise. These fixtures are built from each other, which is realistic of our real development environment rather than the tests being expected to use universal data in some situations and local data in others.

For tests which are based on mutations, I have removed virtually all of the hard-coded aspects of them in their assertions and built the assertions from the data they were first fed. I've also extracted the mutated strings so they are referenced in the mutation and the assertion so they are tightly coupled.


package.json
```diff
@@ -72,7 +73,7 @@
     "identity-obj-proxy": "3.0.0",
     "jest": "^29.5.0",
     "jest-environment-jsdom": "^29.5.0",
-    "jest-environment-jsdom-fifteen": "^1.0.2",
+    "msw": "1.3.2",
```
ckbedwell, line 76: I'm going to leave this here as I am about to put another PR up which adds MSW. I wanted to do them in separate PRs rather than make a giant one. It also is useful to validate all the tests work correctly as they are now before utilizing MSW.

package.json
```diff
@@ -86,7 +87,7 @@
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^4.2.0",
     "tslib": "2.5.3",
-    "typescript": "4.8.4",
+    "typescript": "4.9.4",
```
ckbedwell, line 90: I also bumped Typescript out but I think I am going to bump it further. I wanted to utilize the `satisfies` keyword introduced in `4.9`.

src/validation.test.ts
```diff
@@ -17,100 +10,14 @@ import {
...
-          tls: false,
-        },
-      },
-    };
-
-    expect(validateCheck(check)).toBe(false);
+  BASIC_CHECK_LIST.forEach((check) => {
+    test(`should validate ${Object.keys(check.settings)[0]} check`, () => {
+      expect(validateCheck(check)).toBe(false);
+    });
```
ckbedwell, line 20: I just want to pick on this test in particular because I think this really shows the power of having unified data and more closely mocking a real world database for our integration tests.

As all tests are now based on the tests within `BASIC_CHECK_LIST`, you should be able to update any value in those tests and no longer break any other test -- unless you do something which would be invalid, so if for any reason our type checker doesn't pick it up, then this test should as a fallback.

On that note, I have another tech debt PR planned to make the Check typing _much_ tighter.

ckbedwell (commented):
Self-review.

rdubrock (approved):
LGTM!

PR #688, opened 2024-02-16
Feat/introduce msw

ckbedwell:
## Problem

For our current testing suite, all of our datasource instances are mocked and injected into our tests which means we have zero test coverage for any changes we do to the datasource. Due to the nature of our tests moving towards fully embracing integration testing practices this gives us a a false confidence that any changes we make in the datasource will be caught by our tests, which is not the case.

## Solution

Utilise [MSW](https://mswjs.io/) to act as our server for our integration tests. This means we move the responsibility for what our server layer is in our tests to something which we have no control over in our code as it is completely abstracted away -- this gives us a huge confidence boost in our tests and also allows us to make larger changes with no concerns about breaking tests due to them being tightly coupled with implementation details.

It also means we can easily mock different server responses per test so we can make sure we are testing various unhappy paths in the future.

It also has a huge added benefit that all of our tests api data is typed to help speed up development:

A screenshot of VSCode autocomplete filling in the available list of api routes in the apiRoute function.

A screenshot of VSCode type checker recognizing that the wrong shape of data has been provided to the `listChecks` api route.


__mocks__/@grafana/runtime.tsx
ckbedwell, line 1: I had to move this to being imported in set-up tests -- it was having issues being mocked when I left it here for some reason. To be honest, I don't think it is a bad thing as it consolidates finding mocked dependencies.

__mocks__/@grafana/ui.tsx
```diff
@@ -47,7 +47,7 @@ const MultiSelect = forwardRef((props, ref) => <Select {...props} ref={ref} mult
 MultiSelect.displayName = 'MultiSelect';
 
 // Using this specifically for Probe filter at the moment
-const mockProbeOptions = [{ label: 'Chicago', value: 22 }];
+const mockProbeOptions = [{ label: `tacos`, value: 1 }];
```
ckbedwell, line 50: This should reference PRIVATE_PROBE.name but it can't compile correctly. This will be removed and mocked correctly once the react-query PR goes in.

jest.config.js
```diff
@@ -7,7 +7,7 @@ const config = require('./.config/jest.config');
 module.exports = {
   // Jest configuration provided by Grafana scaffolding
   ...config,
-  setupFilesAfterEnv: ['@testing-library/jest-dom/extend-expect', ...(config.setupFilesAfterEnv || [])],
+  setupFilesAfterEnv: ['<rootDir>/src/test/jest-setup.tsx'],
```
ckbedwell, line 10: I wanted to simplify the jest set-up as it was quite confusing all the places you have to look to find out what our jest dependencies and mocks are. I'm starting to consolidate all the ones we manage to the jest-setup file, rather than having them spread between here, the various mocks files, and the jest setup file.

src/components/BulkEditModal/BulkEditModal.test.tsx
```diff
@@ -57,15 +53,20 @@ test('successfully adds probes', async () => {
     },
   ];
 
-  const { instance, user } = renderBulkEditModal('add', checksWithASingleProbe);
+  const { record, read } = getServerRequests();
+  server.use(apiRoute(`bulkUpdateChecks`, {}, record));
```
ckbedwell, line 57: I'll document how to use these correctly in the dev readme once I've finished the full implemenation.

src/components/CheckHealth.test.tsx
ckbedwell, line 1: Wasn't used so I deleted it.

ckbedwell (commented):
Self-review.

rdubrock (approved):

PR #692, opened 2024-02-20
Update usage types and fix usage tests utilizing SSL.

ckbedwell:
This is Part 1 of a types refactor to tighten up the Checks type.

I needed to split this particular work out because the usage calculations are comparatively complicated and I believe some of the tests are wrong so want to bring special attention to them rather than this getting buried in a larger PR.

# Problem

The usage calculations are typed too loosely referencing `Check`s when that isn't _really_ what they need to calculate correctly. It appears the tests that accompany the usage checks are also wrong.

There were also multiple bugs that have been addressed:

## The homepage usage stats now include multihttp check calculations correctly:

Screenshot of the usage stats on the Synthetic Monitoring homepage.

**After**

Screenshot of the usage stats on the Synthetic Monitoring homepage.

## SSL not accounted for
**Before**

https://github.com/grafana/synthetic-monitoring-app/assets/6906380/2840bf78-fcd8-4a72-bca3-c776eb729fe2

**After**

https://github.com/grafana/synthetic-monitoring-app/assets/6906380/adbec3ea-5baf-4415-aa95-5a7d7a9736f4

# Solution

Re-type what is expected to be provided to the usage calculation functions and update the tests. Associate the tests with the checkinfo provided by the endpoint so the intent is clearer and that future developers can understand the relationship between the checkinfo endpoint and how they correlate to usage stats.


src/hooks/useUsageCalc.ts
```diff
@@ -1,92 +1,56 @@
...
-  const withSSL = addSSL(check, baseClass);
-
-  const withBasic = check.basicMetricsOnly ? `${withSSL}_basic` : withSSL;
+  return AccountingClassNames[accountClass];
+};
 
-  return AccountingClassNames[withBasic as keyof typeof AccountingClassNames];
+const calcFunctionMap = {
+  multi: calculateMultiHTTPUsage,
+  other: calculateUsage,
```
ckbedwell, line 24: It would be useful to document how these are being calculated. I was considering if you could merge these two functions as at first glance it appears all the multi-http entry is doing is checking how many requests are being made -- then I noticed it had hard-coded values that differ from the other usage calculations but I didn't understand why.

src/components/CheckEditor/CheckSettings.tsx
```diff
@@ -34,5 +34,8 @@ export const CheckSettings: FC<Props> = ({ isEditor, typeOfCheck }) => {
     case CheckType.K6: {
       throw new Error('Invalid check type for this location');
     }
+    case CheckType.GRPC: {
+      throw new Error('Invalid check type for this location');
+    }
```
ckbedwell, line 39: To be as accurate as possible I've decided not to shy away from the GRPC type checking but make sure it is valid and leave it in as it is something that will be worked on soon.

src/hooks/useUsageCalc.test.tsx
```diff
@@ -67,16 +84,18 @@ describe('http usage', () => {
...
       },
       frequency: 10000,
-    });
+    } as Check);
     expect(withSSL.current).toStrictEqual({
       checksPerMonth: 262800,
-      activeSeries: 118,
+      activeSeries: http_ssl.Series,
       logsGbPerMonth: 0.21,
-      dpm: 708,
```
ckbedwell, line 79: In the test all of the references to SSL assertions had identical values to the non-SSL entries which seems to be wrong?

src/types.ts
```diff
@@ -624,3 +626,12 @@ export interface ThresholdSettings {
...
 }
+
+export interface CalculateUsageValues {
+  assertionCount: number;
+  basicMetricsOnly: boolean;
+  checkType: CheckType;
+  frequencySeconds: number;
+  isSSL: boolean;
+  probeCount: number;
+}
```
ckbedwell, line 637: This is the core part of this PR. Will all be explained in the next PR.

PR #697, opened 2024-02-22
Allow unlimited headers

rdubrock:
Allow unlimited headers in http requests

### DIFF
--- a/src/components/NameValueInput/NameValueInput.tsx
+++ b/src/components/NameValueInput/NameValueInput.tsx
@@ -3,7 +3,8 @@ import { FieldErrorsImpl, useFieldArray, useFormContext } from 'react-hook-form'
 import { Button, Field, HorizontalGroup, Icon, IconButton, Input, useTheme2, VerticalGroup } from '@grafana/ui';
 import { css } from '@emotion/css';
 
-import { CheckFormValues, CheckFormValuesHttp } from 'types';
+import { CheckFormValues, Probe } from 'types';
+import { isHttpErrors } from 'utils.types';
 
 type NameValueName = 'settings.http.headers' | 'settings.http.proxyConnectHeaders' | 'labels';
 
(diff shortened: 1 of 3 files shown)


src/components/NameValueInput/NameValueInput.tsx
```diff
@@ -1,39 +1,54 @@
...
   disabled?: boolean;
   label: string;
   validateName?: (name: string) => string | undefined;
   validateValue?: (value: string) => string | undefined;
 }
 
+function getErrors(errors: FieldErrorsImpl<CheckFormValues>, name: NameValueName) {
+  switch (name) {
+    case 'settings.http.headers': {
+      const cast = errors as FieldErrorsImpl<CheckFormValuesHttp>;
```
rdubrock, line 22: I was fighting with this for a long time. I don't get why the new union types won't allow me to say `CheckFormValues` _might_ have HTTP on it. Is there a way to say "I have a form that might be of any check type, but if it's HTTP I would like to use the HTTP key"? Casting always feels icky.

@ckbedwell

src/components/http/HttpSettings.tsx
```diff
@@ -211,7 +211,6 @@ export const HttpSettingsForm = ({ isEditor }: Props) => {
               name="settings.http.headers"
               disabled={!isEditor}
               label="header"
-              limit={10}
               validateName={validateHTTPHeaderName}
               validateValue={validateHTTPHeaderValue}
```
peterschretlen, line 215: There is also a proxy header limit of 10 below - should we remove that one too?

rdubrock: Yeah good call

src/components/NameValueInput/NameValueInput.tsx
```diff
@@ -1,39 +1,54 @@
...
+  }
+}
+
 export const NameValueInput = ({ name, disabled, limit, label, validateName, validateValue }: Props) => {
   const {
     register,
     control,
     formState: { errors },
-  } = useFormContext(); // todo: type correctly
+  } = useFormContext<CheckFormValues>();
```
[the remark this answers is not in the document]
rdubrock: Any reason not to just union the two options? I feel like this component doesn't need to care, it just needs to know there's a `label` field

ckbedwell (approved):

PR #699, opened 2024-02-22
Add a new drone config

rdubrock:
Always feels a bit risky to cut over CI, so I'm going to wait until next week to merge. There are a number of improvements:
- PR builds will no longer overwrite main, which in turn will stop breaking dev unless we meant to
- Every merge to main will create a PR in deployment_tools with a hashed version. Merging that will cause it to roll out to the dev environment. 
- Releasing to staging and production can be accomplished just with promotion. Selecting Promote => "staging" will open a PR in deployment_tools for staging. Promote => "production" will put a build in the production GCS bucket and open a PR...you guessed it...in deployment_tools

Things left to do
- I haven't tested the gcom-release part of this CI. I don't want to run it until we're actually ready to deploy
- We need to point drone at the .drone.yml config file in the UI in order for it to start using this configuration instead of the shared config


mem (approved):

Makefile
```diff
@@ -0,0 +1,96 @@
...
+  	else \
+		yq -Poy rules.json > rules.yml; \
+	fi
+	rm rules.json
+
+test-promql: rules.yml
+	promtool check rules --lint=all --lint-fatal rules.yml
+
+.PHONY: install-promql-deps
+install-promql-deps:
```
rdubrock, line 64: All the promql rules are for something that will come later. The k8s crew has a really interesting system of extracting promql queries and validating them as part of CI

PR #708, opened 2024-02-27
chore: add a publish to gcom promotion step

rdubrock:
This is just so I can test promoting to gcom without running the full deployment flow


rdubrock:
This is mostly a failure, but getting closer

PR #711, opened 2024-02-28
chore: clean up test output

rdubrock:
The MultiHTTP tests were spewing lots of act errors. I think the `waitFor` was originally put in there when render behaved differently?

I was also getting tests timing out a lot, so I turned the global timeout back on

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -15,7 +15,7 @@ module.exports = {
       '<rootDir>/src/test/fileMock.js',
     '^!raw-loader!*': '<rootDir>/src/test/rawLoaderMock.js',
   },
-  // testTimeout: 30000,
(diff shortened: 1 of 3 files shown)


ckbedwell (approved):
LGTM! 💪 

I'm going to make a ticket to take a better look at the tests. I think we have some really great and comprehensive tests but some of them are just a bit too big so become difficult to debug. In [this last PR](https://github.com/grafana/synthetic-monitoring-app/pull/710/files#diff-e747fc1117a0159e4dfd2139446c20810dfb9425ef70567d488005d338812ca5R123-R348) I broke down the huge HTTP check into checking the different sections as it was becoming really difficult to debug and reason what was going on.

PR #741, opened 2024-04-04
chore: add fs data attributes to the check form

rdubrock:


src/components/TimeSlider/TimeSlider.tsx
```diff
@@ -139,6 +140,7 @@ export function TimeSlider({
           onChange={onMinutesInputChange}
           onBlur={onMinutesInputBlur}
           id={minutesId}
+          data-fs-element="Timeslider minutes input"
```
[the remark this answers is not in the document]
rdubrock: I don't know. I'll swap around how this is happening to make them unique

ckbedwell (changes requested):
I found quite a useful way to check for how the experience might look in FullStory is opening up the console and adding (or you could just add it in the App root whilst you are developing temporarily):

```
document.addEventListener(`focus`, (e) => { console.log(e.target.getAttribute(`data-fs-element`)) }, true)
```

Then this yields these logs so you can see what interactive elements are missing the `data-fs-element` if you tab through with your keyboard.

A screenshot of Chrome's console with a list of data-fs-attributes alongside some bugs and null results.

src/components/Collapse.tsx
```diff
@@ -24,6 +24,7 @@ export const Collapse = ({ className, isOpen, label, children, onClick, ...props
           onClick?.(!isOpenState);
         }}
         type="button"
+        data-fs-element={`Collapse header ${label}`}
```
ckbedwell, line 27: This is a React.Node so you can't really use this one. Might be worth adding it higher up and letting this accept `...rest` props so it is added automatically. (Meant to leave this comment with the review -- also shows how I discovered it quickly as it logged [Object object]).

rdubrock: Hmmm yeah. I've updated it to come from above, but since we're already using the `...rest` pattern here with a destination that isn't the button, the syntax gets a little gross. Sort of hinges on how accurate the `document.addEventListener...` simulation is, because putting the attribute on the wrapping div might or might not do the thing we want. I can't find anything more specific in the docs than "We look at the DOM"

src/components/Collapse.tsx
```diff
@@ -8,12 +8,15 @@ interface Props {
...
   onClick?: (open: boolean) => void;
+  'data-fs-element'?: string;
 }
 
 export const Collapse = ({ className, isOpen, label, children, onClick, ...props }: PropsWithChildren<Props>) => {
   const [isOpenState, setIsOpenState] = React.useState(isOpen);
   const styles = useStyles2(getStyles);
 
+  const analyticsName = props['data-fs-element'];
+  delete props['data-fs-element'];
```
ckbedwell, line 19: You can do this a cleaner way when you have kebab-case props by using the renaming function when destructuring:

```
{ className, isOpen, label, children, onClick, 'data-fs-element': dataFsElement, ...props }
...
<button data-fs-element={dataFsElement} ... />
```

Then you get rid of all the horrible props mutation 😃

src/components/Collapse.tsx
ckbedwell: >  Sort of hinges on how accurate the `document.addEventListener...` simulation is, because putting the attribute on the wrapping div might or might not do the thing we want. I can't find anything more specific in the docs than "We look at the DOM"

I rewrote the simulation slightly so it recursively checks parent elements for the `data-fs-element` attribute because I figure they must be doing something like that in the worst case:

```
document.addEventListener(
  'focus',
  (e) => {
    if (e.target instanceof HTMLElement) {
      logDomChain({ node: e.target });
    }
  },
  true
);

type logDomChainProps = {
  parentChain?: string[];
  node: HTMLElement | null;
};

function logDomChain({ parentChain = [], node }: logDomChainProps) {
  if (node === null) {
    return console.log(constructChain(parentChain));
  }

  const dataFsElement = node.getAttribute(`data-fs-element`);

  if (dataFsElement) {
    return console.log(constructChain([...parentChain, dataFsElement]));
  }

  const newChain = [...parentChain, node.tagName.toLowerCase()];
  return logDomChain({ node: node.parentElement, parentChain: newChain });
}

function constructChain(chain: string[]) {
  return [...chain].reverse().join(` > `);
}
```

This is showing me that none of the Selects or Sliders are receiving the attribute and is also showing me some other fields we haven't tagged up. (I figure we might as well try and remove as much noise as possible, right?).

In those cases where we can't inject it we could either write some 'fun-but-not-really'™️ React code to target the underlying elements or just try adding it to the wrapping `<Field />` component or such?

A screenshot of Chrome's console dev tools with a list of selectors. Some of them are showing the values for data-fs-element whilst others are showing a very long html tag chain up to the document body.

rdubrock: Nice! Looking at the fullstory results, it looks like it will report the `id` in a readable way if it's present. Maybe that's a way out for select/slider that won't let us pass arbitrary attributes down to the dom?

src/components/Collapse.tsx
rdubrock: haha all the cool kids are doing prop mutation

ckbedwell (approved):
LGTM!

PR #748, opened 2024-04-11
Refactor: check transformations split

ckbedwell:
## Problem

I was thinking after [the last PR](https://github.com/grafana/synthetic-monitoring-app/pull/745) about how sometimes the UI form state can incorrectly pollute the payload state we should do a more thorough check of sanitising the payload. The last PR was a quick fix but I wasn't particularly happy about how maintainable it would be if we needed to do more sanitisation.

## Solution

This PR is to address a few points:

- Break up the `checkFormTransformation` monolith into 18 separate files to make navigation and maintainability easier:
  - Eight `toPayload` files for each check type (e.g. `toPayload.dns.ts`, `toPayload.http.ts`, etc.)
  - a payload utils file
  - Eight `toFormValues` files for each check type (e.g. `toFormValues.dns.ts`, `toFormValues.http.ts`, etc.
  - a formValues util file
  
- I also took this opportunity to remove all of the `SelectableValues` that need to be filtered out of the form payloads to make the transformations much simpler.
- This led to create some more type-safe functions, such as `getDnsValidations` in `toFormValues.dns.ts` and I uncovered a typo that has existed for quite a long time so it wasn't getting added correctly to DNS checks 🙈  (`dnsSettings.validateAditionalRRS` -> `dnsSettings.validateAdditionalRRS`)
- As well another typo in `DNSCheckResponseMatches` where the wrong type was being appended `matchType` -> `responseType`


rdubrock (approved):
LGTM

PR #753, opened 2024-04-16
scripted check latency: use avg of http requests duration

peterschretlen:
The latency column for scripted check isn't really latency right now - it's the average run duration of the k6 script, which can include lots of non-request overhead such as k6 startup time. 

For multihttp we take the average the http requests rather than the average execution duration of the test. 
This PR applies the same to scripted checks.  

The only downside is that scripted checks can include other protocols like websockets or grpc,  so we will not accurately capture average latency when those are used.   But in the interim, using http request duration average is better than the alternative of average execution duration.

### DIFF
--- a/src/data/useLatency.ts
+++ b/src/data/useLatency.ts
@@ -33,6 +33,8 @@ export function useLatency({ job, target, settings }: Check) {
 }
 
 function getQuery(job: Check['job'], target: Check['target'], type: CheckType) {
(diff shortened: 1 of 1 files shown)


ckbedwell (approved):
I just pushed up a comment to make this assumption explicitly clear rather than having to check this PR in the git blame.

Otherwise LGTM! 🚀

PR #755, opened 2024-04-19
fix: query logs by probe success instead of error level

rdubrock:
Untoggled:

Screenshot 2024-04-19 at 11 23 32

Now toggled:

Screenshot 2024-04-19 at 11 23 37

Hopefully this makes it easier to see _why_ a check failed. The one in this screenshot timed out. Before this change the logs were getting filtered by error level, and all the lines that preceded the actual error got filtered out.


peterschretlen (approved):
Nice! LGTM

PR #757, opened 2024-04-22
feat: add toggle to prevent scripted checks from running on probes

rdubrock:
Probes now have a `capabilities` key that allows them to communicate their preferences to the frontend. The only one that exists so far is whether to disable scripted checks or not, that allows us to prevent users from added scripted checks to non-AWS probes, and allows users to decide whether to run scripted checks on private probes


ckbedwell (approved):
LGTM 🚀

The-9880 (approved):
Thank you! 😄

PR #759, opened 2024-04-23
Form as wizard

rdubrock:
Screenshot 2024-04-23 at 10 46 56

Initial exploration for turning the form into a wizard. There's a couple dicey things I chucked in here that I'll highlight in comments to get opinions

### DIFF
--- a/src/components/CheckForm/CheckForm.tsx
+++ b/src/components/CheckForm/CheckForm.tsx
@@ -70,6 +70,7 @@ const CheckFormContent = ({ check, checkType, overCheckLimit, overScriptedLimit 
   const formMethods = useForm<CheckFormValues>({
     defaultValues: initialValues,
     shouldFocusError: false, // we manage focus manually
+    mode: `onBlur`,
   });
 
   const { updateCheck, createCheck, deleteCheck, error, submitting } = useCUDChecks({ eventInfo: { checkType } });
--- a/src/components/CheckForm/CheckForm.tsx
+++ b/src/components/CheckForm/CheckForm.tsx
@@ -1,12 +1,12 @@
 import React, { BaseSyntheticEvent, useMemo, useRef, useState } from 'react';
 import { FormProvider, SubmitErrorHandler, SubmitHandler, useForm } from 'react-hook-form';
 import { useParams } from 'react-router-dom';
-import { GrafanaTheme2, OrgRole } from '@grafana/data';
+import { GrafanaTheme2 } from '@grafana/data';
 import { Alert, Button, ConfirmModal, LinkButton, useStyles2 } from '@grafana/ui';
 import { css } from '@emotion/css';
 
 import { Check, CheckFormValues, CheckPageParams, CheckType, ROUTES } from 'types';
-import { hasRole, isOverCheckLimit, isOverScriptedLimit } from 'utils';
+import { isOverCheckLimit, isOverScriptedLimit } from 'utils';
 import { useChecks, useCUDChecks } from 'data/useChecks';
 import { useTenantLimits } from 'data/useTenantLimits';
 import { useNavigation } from 'hooks/useNavigation';
@@ -74,7 +74,6 @@ const CheckFormContent = ({ check, checkType, overCheckLimit, overScriptedLimit 
 
   const { updateCheck, createCheck, deleteCheck, error, submitting } = useCUDChecks({ eventInfo: { checkType } });
 
-  const isEditor = hasRole(OrgRole.Editor);
   const navigate = useNavigation();
   const navigateBack = () => navigate(ROUTES.Checks);
   const onSuccess = () => navigateBack();
@@ -138,20 +137,6 @@ const CheckFormContent = ({ check, checkType, overCheckLimit, overScriptedLimit 
         </Button>
       );
     }
-    if (check.id) {
-      actions.push(
-        <Button
-          variant="destructive"
-          key="delete"
-          data-fs-element="Delete check button"
-          onClick={() => setShowDeleteModal(true)}
-          disabled={!isEditor}
-          type="button"
-        >
-          Delete Check
-        </Button>
-      );
-    }
     actions.push(
       <Button
         type="submit"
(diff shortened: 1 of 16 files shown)


src/components/CheckForm/CheckForm.tsx
```diff
@@ -123,6 +124,70 @@ const CheckFormContent = ({ check, checkType, overCheckLimit, overScriptedLimit
     deleteCheck(check, { onSuccess });
   };
 
+  const actions = useMemo(() => {
+    const actions = [
```
rdubrock, line 128: This is probably the code smelliest thing. I'm kind of unsure about delegating the responsibility for rendering actions to FormLayout. It's not super ergonomic and requires some prop drilling. It was the quickest way I could see to get the desired layout, though

src/components/CheckForm/FormLayout/FormLayout.test.tsx
```diff
@@ -77,7 +77,7 @@ describe(`FormLayout`, () => {
         <FormLayout.Section label="First section">
           <div>{firstSectionText}</div>
         </FormLayout.Section>
-        <FormLayout.Section label="Second section" isOpen>
+        <FormLayout.Section label="Second section">
```
rdubrock, line 80: Tests still need to be updated, didn't want to fiddle with them until I know we've got consensus on the direction though

src/components/CheckForm/FormLayout/FormLayout.tsx
```diff
@@ -1,42 +1,117 @@
...
+                <Button onClick={() => setActiveIndex(activeIndex - 1)} icon="arrow-left" variant="secondary">
+                  {sectionHeaders[activeIndex - 1].label}
+                </Button>
+              )}
+              {activeIndex !== sectionHeaders.length - 1 && (
+                <Button
+                  onClick={() => {
+                    trigger(sections?.[activeIndex].props.fields).then((valid) => {
+                      if (valid) {
+                        setActiveIndex(activeIndex + 1);
```
rdubrock, line 91: The state of the wizard is internal to the FormLayout component at the moment. It could also be via URL/routing, I don't have a strong preference, opinions welcome

ckbedwell: My preference is to be able to replicate state via the URL but I don't think it is a requirement for this PR.

src/components/CheckForm/FormLayout/FormSidebar.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {sections.map(({ label: sectionTitle, required, hasErrors }, sectionIndex) => {
+        const prefix = hasErrors ? (
+          <Icon name={`exclamation-triangle`} color={config.theme2.colors.error.main} />
+        ) : (
+          <span className={css({ width: '16px', display: 'inline-block' })}>{sectionIndex + 1}</span>
+        );
+        return (
+          <li
+            key={sectionTitle}
+            onClick={() => onSectionSelect(sectionIndex)}
```
ckbedwell, line 28: Nest a button inside the `li` and here and move the onClick to that so we get lots of a11y niceties 😃

src/components/CheckForm/FormLayout/FormSidebar.tsx
```diff
@@ -0,0 +1,66 @@
...
+            className={cx(styles.listItem, { [styles.active]: activeIndex === sectionIndex })}
+          >
+            <div className={css({ cursor: 'pointer' })}>
+              {prefix}&nbsp;{sectionTitle}
+              {required && ' *'}
+            </div>
```
ckbedwell, line 34: The spacing and alignment looks a little off here. I'd drop the `&nbsp` character and flex the container with a gap and `align-items: center`. I think as a rule spacing should be handled by the theme as much as possible.

src/components/CheckForm/FormLayout/FormLayout.tsx
```diff
@@ -1,42 +1,117 @@
...
+            index !== sectionCount - 1
+              ? () => {
+                  trigger(child.props.fields).then((valid) => {
+                    if (valid) {
+                      setActiveIndex(activeIndex + 1);
+                    }
+                  });
+                }
+              : undefined
+          }
```
ckbedwell, line 57: I don't think this is used?

src/components/CheckForm/FormLayout/FormLayout.tsx
```diff
@@ -1,42 +1,117 @@
...
+              {activeIndex !== 0 && (
+                <Button onClick={() => setActiveIndex(activeIndex - 1)} icon="arrow-left" variant="secondary">
+                  {sectionHeaders[activeIndex - 1].label}
+                </Button>
+              )}
+              {activeIndex !== sectionHeaders.length - 1 && (
+                <Button
+                  onClick={() => {
+                    trigger(sections?.[activeIndex].props.fields).then((valid) => {
+                      if (valid) {
```
ckbedwell, line 90: I don't think we should block advancing in the form if the section isn't valid. We don't block it when they are using the sidebar and some people might just want to explore the form options before they fill it all in.

src/components/CheckForm/FormLayout/FormLayout.tsx
```diff
@@ -1,42 +1,117 @@
...
+      <FormSidebar sections={sectionHeaders} onSectionSelect={setActiveIndex} activeIndex={activeIndex} />
+      <div
+        className={css({
+          display: 'flex',
+          flexDirection: 'column',
+          flexGrow: '1',
+          justifyContent: 'space-between',
+        })}
+      >
+        <div className={css({ paddingLeft: '24px' })}>{sections}</div>
```
ckbedwell, line 76: I think we should set the `maxWidth: 800px` here. On my ultrawide the primary action buttons are a mile away from everything else.

Screenshot of Synthetic Monitoring on a 3440px ultra wide on the check creation page in the new wizard form layout. The primary action buttons are very far to the right at the bottom of the screen. The rest of the content is in the first third of the screen on the left.

ckbedwell (changes requested):
This is awesome that you got this out so quickly!

Two big things for me that I think are missing:
1. The validation _confirmation_ for the sections (so a green tick when it is valid). I think that is just as important as the the error state because it creates a quick feedback loop for the user that they haven't missed anything.
2. When submitting, going back to the first invalid section and scrolling to the error. It's easy to miss a required section, click submit and not notice that a section's icon has gone invalid as that is the only feedback you get.

The trigger / blur validation works okay. I don't think I'd block the PR if we kept this but I am certainly interested in exploring having an outside schema which gets resolved. If we took this approach we could remove the dependency for `react-hook-form` to trigger validation for each section, as we could do it ourselves by referencing the schema.

This is how they do it in the [Wizard Form Example](https://grafana.com/developers/saga/templates/multistep-form-page) on Saga.

rdubrock:
Thanks for the review! I think it's in a mostly acceptable state now. 

1. I find validation confirmation to be very tricky. Like, do we show it if the form is in a valid state? Or only if the fields haven't been touched yet? Or maybe only if the form section hasn't been viewed yet? And then what about editing, when there is a value that is valid already, but the user hasn't opened the section yet? I opted for the simplest one: if the user leaves a section that has required fields and they are valid, it turns green. 

2. I added the focusing/scrolling on error submit back into the form, unfortunately the way I saw to get there sort of further wrecked the previously very pleasant API design of `FormLayout`

Issue #766, opened 2024-04-25
TCP Check creation: Query params form is shown when target is a hostname (not IP)


w1kman:
![image](https://github.com/grafana/synthetic-monitoring-app/assets/1382208/664922f1-7cfe-41a7-951e-6e52789148d1)

**What happened**:
Query params is shown when target is a hostname (not when it is an IP). 

**What you expected to happen**:
Query params wasnt configurable since they cannot be used with this check type (TCP)

**How to reproduce it (as minimally and precisely as possible)**:
- Create new TCP check
- Enter `grafana:` as target
- See how the "Query Params" form is displayed (whilst not usable)

**Anything else we need to know?**:
The underlying issue is that when ever `CheckTarget` is used, it will show "Query Params" based on input value, not taking check type in consideration.

**Environment**:
- Grafana version (if running on prem): `v11.x`
- SM plugin version: `v1.13.4`
- OS Grafana is installed on: Official docker container
- User OS & Browser: macOS Sanoma 14.4.1 (23E224) / Chrome Version 124.0.6367.80 (Official Build) (arm64)

PR #768, opened 2024-04-25
Fix: only show query params for `CheckType.HTTP`

w1kman:
Every check that was using `CheckTarget` component could have the "Query Params" inputs rendered based on the `target` value. AFAIK, only `CheckType.HTTP` is using the "Query Params" part of `RequestTargetInput`.

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/766

image
image
image
image


ckbedwell (approved):
LGTM 🚀

PR #769, opened 2024-04-25
chore: add check usage to the scripted check types

rdubrock:


peterschretlen (approved):
Looks good to me - thank you for squeezing this in 🙏

PR #776, opened 2024-04-29
fix: set fixed overflow widgets on monaco editor

rdubrock:
This fixes the autosuggest window wandering away from the cursor for the check editor. The actual fix is adding `          fixedOverflowWidgets: false,`, the rest is just reverting back the responsive layout. 

I'm not sure _why_. It has something to do with how Monaco handles overflows, and how `container` is sizing things, but I'm unsure what's going on there

### DIFF
--- a/src/components/CheckForm/FormLayout/FormLayout.tsx
+++ b/src/components/CheckForm/FormLayout/FormLayout.tsx
@@ -198,36 +198,34 @@ function checkForErrors({
 
 const getStyles = (theme: GrafanaTheme2) => {
   const containerName = `checkForm`;
-  // const breakpoint = theme.breakpoints.values.xs;
-  // const query = `(min-width: ${breakpoint + 1}px)`;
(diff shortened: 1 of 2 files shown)


ckbedwell (approved):
LGTM 🚀 

As discussed on a call just now, the one downside of this solution is the responsiveness of the script editor degrades slightly.

Just for future lookups of this PR, will leave a link to this issue for Monaco as it looks like they are related: https://github.com/microsoft/monaco-editor/issues/2793

[This comment on the issue](https://github.com/microsoft/monaco-editor/issues/2793#issuecomment-1966864571) makes a reference to somebody solved it by removing all properties that lead to a new stacking context.

[MDN's definition of a stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) includes scenarios where an element with a `container-type` value of `size` or `inline-size` which is what is going on here.

PR #780, opened 2024-05-02
fix: traceroute check default timeout

ka3de:

Traceroute checkk only accepts 30s as timeout. Since this was not specified in its default definition, it was falling back to the base check timeout, which specifies a TO of 3s.


rdubrock (approved):
Confirmed this will fix the problem. The other option would be to handle it in the submit (in `toPayload.traceroute.ts`), I don't have a preference really

PR #781, opened 2024-05-03
Tests/traceroute tests

ckbedwell:
To expand upon https://github.com/grafana/synthetic-monitoring-app/pull/780. I wanted to make sure it doesn't regress and add tests to ensure so. I also wanted to make sure the Traceroute timeout input has a single source of truth rather than being defined within the component.

### DIFF
--- a/src/components/CheckEditor/ProbeOptions.tsx
+++ b/src/components/CheckEditor/ProbeOptions.tsx
@@ -96,7 +96,7 @@ export const ProbeOptions = ({ checkType }: Props) => {
       <Field
         label="Frequency"
         description="How frequently the check should run."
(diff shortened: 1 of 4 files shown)


src/components/CheckEditor/ProbeOptions.tsx
```diff
@@ -95,7 +96,7 @@ export const ProbeOptions = ({ checkType }: Props) => {
       <Field
         label="Frequency"
         description="How frequently the check should run."
-        disabled={!isEditor || isTraceroute}
+        disabled={!isEditor || isTraceroute} // is this correct?
```
ckbedwell, line 99: I noticed the field is disabled but the input isn't, should frequency be disabled?

PR #782, opened 2024-05-03
feat: zod validation set-up

ckbedwell:
# Problem

For the relaunch of Synthetics we were desperate to get the wizard form out so we set the expectation with customers  what the experience of SM 2.0 is moving forwards. All of our forms are powered by `react-hook-form` and it doesn't work so well out of the box when utilizing a wizard form pattern.

In particular one workaround we had to implement to hit the deadline is triggering the form validation when moving from one section to the next, combined with setting the form validation `mode` to `onBlur` which creates a less than ideal check creation and editing experience. [Details of the problems introduced with the current pattern are discussed in the initial PR](https://github.com/grafana/synthetic-monitoring-app/pull/759) but include:

- Errors not clearing until you focus out of the field
- Errors not clearing when utilizing auto-complete / the probe 'All' button (or other buttons which populate values)
- Errors appearing in later sections as soon as you tab in / out of them

# Solution

Remove all of the inline validation for each field and provide a [zod schema](https://zod.dev/) to the check form to resolve. This has multiple benefits -- it doesn't only fix all of the issues listed above -- but it also makes a tight coupling of our user-facing check validation with our developer-defined Typescript definitions of the checks. We are now able to validate any values against all or part of a check's schema as needed.

The other big benefit is that a check's schema is now defined in a single place within the plugin's code-base rather than scattered throughout and is no longer tied to validating against what is rendered to the DOM. It gives a huge confidence boost to how we manage our forms.

The process of tying the zod schemas to our Typescript definitions uncovered several areas in the UI which weren't providing the user with error feedback.

A potential long-term goal of moving to zod schemas is utilizing the check's protobuf definitions in the backend repo and generating schemas automatically so we can ensure the validation is 100% aligned between services.


ckbedwell (commented):
Self-review.

src/schemas/forms/ScriptedCheckSchema.ts
```diff
@@ -0,0 +1,19 @@
+import { z, ZodType } from 'zod';
+
+import { CheckFormValuesScripted, CheckType, ScriptedSettings } from 'types';
+
+import { BaseCheckSchema } from './BaseCheckSchema';
+
+const ScriptedSettingsSchema: ZodType<ScriptedSettings> = z.object({
+  script: z.string().min(1, `Script is required.`),
```
VikaCep, line 8: Looks like there's an [API validation](https://github.com/grafana/synthetic-monitoring-api/pull/846) for `128kb` on size, do you think we should enforce some sort of limit validation on the frontend?

src/components/CheckEditor/FormComponents/HttpCheckRegExValidation.tsx
```diff
@@ -48,44 +68,41 @@ export const HttpCheckRegExValidation = () => {
...
                   {isHeaderMatch && (
                     <div className={styles.validationHeaderName}>
-                      <Input
-                        {...register(`${REGEX_FIELD_NAME}.${index}.header`)}
-                        placeholder="Header name"
-                        data-fs-element={`Regex header name ${index}`}
-                      />
+                      <Field
+                        className={styles.field}
+                        invalid={isHttpRegexHeaderError(baseErrorPath) && Boolean(baseErrorPath?.header) ? true : false}
```
VikaCep, line 84:
```suggestion
                        invalid={isHttpRegexHeaderError(baseErrorPath) && Boolean(baseErrorPath?.header)}
```

src/components/CheckEditor/FormComponents/HttpCheckRegExValidation.tsx
```diff
@@ -48,44 +68,41 @@ export const HttpCheckRegExValidation = () => {
...
                     </div>
                   )}
-                  <Input
-                    {...register(`${REGEX_FIELD_NAME}.${index}.expression`)}
-                    placeholder="Regex"
-                    data-fs-element={`Regex expression ${index}`}
-                  />
+                  <Field
+                    className={cx(styles.field, styles.grow)}
+                    invalid={Boolean(baseErrorPath?.expression) ? true : false}
```
VikaCep, line 97:
```suggestion
                    invalid={Boolean(baseErrorPath?.expression)}
```

VikaCep: Feel free to disregard though if you think the other option is better for legibility.

src/schemas/general/Header.ts
```diff
@@ -0,0 +1,29 @@
...
+    .min(1, { message: NAME_REQUIRED_ERROR }),
+  value: z
+    .string({
+      required_error: VALUE_REQUIRED_ERROR,
+    })
+    .min(1, { message: VALUE_REQUIRED_ERROR }),
+});
+
+export const HeadersSchema = z.array(HeaderSchema).superRefine((headers, ctx) => {
+  const headerNames = headers.map((label) => label.name);
```
VikaCep, line 20:
```suggestion
  const headerNames = headers.map((header) => header.name);
```

src/components/MultiHttp/Tabs/AssertionsTab.tsx
```diff
@@ -143,82 +167,94 @@ function AssertionFields(props: AssertionProps) {
 
 function AssertionSubjectField({ entryIndex, assertionIndex }: AssertionProps) {
   const { formState } = useFormContext<CheckFormValuesMultiHttp>();
-  const error = formState.errors.settings?.multihttp?.entries?.[entryIndex]?.checks?.[assertionIndex]?.subject;
-  const errMessage = error?.message;
+  const error = formState.errors.settings?.multihttp?.entries?.[entryIndex]?.checks?.[assertionIndex];
+  const err = errorHasSubject(error) ? error?.subject : undefined;
```
VikaCep, line 171: Can we rename `err` to `errorSubject` or similar? I find it a bit confusing to see both `error` and `err`

VikaCep (commented):
Amazing work! 👏👏👏
I still have to test the app itself as I have only looked at the code so far, but it certainly seems like a great improvement. I really like how clean and easy it seems to be to add and modify both simple and complex validations to the different scenarios now 🎉

src/components/CheckEditor/FormComponents/RequestTargetInput.tsx
```diff
@@ -59,6 +52,7 @@ export const RequestTargetInput = ({
               placeholder={targetHelp?.example}
               data-fs-element="Target input"
               {...field}
+              value={typeof field.value === `string` ? field.value : ''}
```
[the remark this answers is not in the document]
ckbedwell: Yeah, this is one of those annoying ones where even though we know it can only be a string with the way we are using it we are adding verbosity to make the Typescript compiler stop complaining.

The problem lies with the loose type definition we ask for the `name` prop. Really we are only allowing very specific name fields here but it is a pain in the ass to write them so I'm going cheap and loose and paying for it this way. We can fix this at a later date if it is proving annoying 😃

src/components/MultiHttp/Tabs/AssertionsTab.tsx
```diff
@@ -57,11 +74,13 @@ export function AssertionsTab({ index, active }: MultiHttpTabProps) {
                           {...rest}
                           options={MULTI_HTTP_ASSERTION_TYPE_OPTIONS}
                           menuPlacement="bottom"
+                          onChange={(e) => {
+                            field.onChange(e.value);
```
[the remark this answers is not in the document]
ckbedwell: Actually, this is a bug that is getting fixed from a previous change and should have been done separately. 😅 

We used to have tons of references of `SelectValue` getting transformed from the form when going into the payload. I removed a ton of them in this [PR](https://github.com/grafana/synthetic-monitoring-app/pull/748/files) in favour of keeping it simple but unfortunately our testing coverage isn't 100% and I missed updating the onChange handler here so it it only writes the value (it used to write the whole option object as the value, e.g. `text` would be `{ label: "Text", value: "text" }.

src/schemas/general/Probes.ts
```diff
@@ -0,0 +1,17 @@
+import { z } from 'zod';
+
+const MAX_PROBES = 64;
+
+export const ProbesSchema = z
```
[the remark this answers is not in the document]
ckbedwell: In theory the UI shouldn't allow you to do that so I'm going to leave it out... also it shouldn't matter if they send duplicates (I think...).

src/schemas/forms/ScriptedCheckSchema.ts
ckbedwell: Screenshot of the Syntehtic Monitoring script editor showing a validation error beneath it as the script is too large.

I have no idea if the way I've implemented this is too naive and we have to account for including imported modules in the limit (which we won't be able to do)? cc @mem

src/schemas/forms/ScriptedCheckSchema.ts
```diff
@@ -4,8 +4,24 @@ import { CheckFormValuesScripted, CheckType, ScriptedSettings } from 'types';
...
+      const sizeInBytes = textBlob.size;
+      const sizeInKb = sizeInBytes / 1024;
+
+      if (sizeInKb > MAX_SCRIPT_IN_KB) {
+        return ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: `Script is too big (${sizeInKb.toFixed(2)}kb). Maximum size is ${MAX_SCRIPT_IN_KB}kb.`,
+        });
+      }
+    }),
```
VikaCep, line 24: Nice!!

VikaCep (approved):
Great work!!! 👏👏👏👏👏

Issue #788, opened 2024-05-22
Actual value of  Validations > SSL options is not sent to the backend when editing an HTTP check


roobre:
**What happened**:

I am able to reproduce the issue reported by a customer [here](https://github.com/grafana/support-escalations/issues/10651): It does not seem to be possible to edit the SSL validation options from the app.

It seems that `settings.http.failIfSSL` and `settings.http.failIfNotSSL` are always sent as `false` in the request to `/sm/check/update`.

Clicking save with the following selection:

![image](https://github.com/grafana/synthetic-monitoring-app/assets/969721/0797feaa-1e99-4c96-af52-989ffd045f8b)

Causes the following payload to be sent:

```json
{
  "id": 11035,
  "tenantId": 501,
  "frequency": 60000,
  "offset": 0,
  "timeout": 10000,
  "enabled": true,
  "labels": [],
  "settings": {
    "http": {
      "ipVersion": "V4",
      "method": "GET",
      "noFollowRedirects": false,
      "tlsConfig": {},
      "failIfSSL": false,
      "failIfNotSSL": false
    }
  },
  "probes": [
    2,
    70
  ],
  "target": "https://expired.badssl.com/",
  "job": "SSLTest",
  "basicMetricsOnly": true,
  "alertSensitivity": "none",
  "created": 1716395175.8532348,
  "modified": 1716395242.9186902
}
```

**What you expected to happen**: They should represent the selected value.

**How to reproduce it (as minimally and precisely as possible)**:
1. Edit an HTTP check
2. Change the "Validation > SSL Options" dropdown to any value other than the default
3. Save changes
4. Observe the outgoing request.

**Anything else we need to know?**:

**Environment**:
- Grafana version (if running on prem):
- SM plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

PR #790, opened 2024-05-23
fix: fix onChange for ssl select and add tests

ckbedwell:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/788

There was a regression when updating the payloads recently where we missed updating the onChange handler for the select in `<HttpCheckSSLOptions />`.

This fixes the regression and adds tests to ensure it doesn't break again in the future.


VikaCep (approved):
💯💯💯

PR #791, opened 2024-05-23
chore: Remove noise which has crept into passing tests

ckbedwell:
# Problem

Some noise has crept into the tests which can get in the way when debugging why tests failed in the CI step.

# Solution

Fix all the areas which are causing noise, resulting in a nicer logging experience 😄 

A tidy list of the 39 past tests, indicating they have passed one after the other on separate lines with no other logging in-between.


src/test/silenceErrors.ts
```diff
@@ -4,6 +4,7 @@ const originalConsoleError = console.error;
 const ignoreErrorsList = [
   `Could not parse CSS stylesheet`,
   `Using kebab-case for css properties in objects is not supported.`,
+  `Warning: Received \`%s\` for a non-boolean attribute \`%s\``, //  See if updating @grafana/ui fixes this one. It is related to the invalid prop on <Field />
```
VikaCep, line 7: Linking the [issue](https://github.com/grafana/synthetic-monitoring-app/issues/799) so we don't forget to check this once it's upgraded.

VikaCep (approved):
👏👏👏

PR #793, opened 2024-05-23
Remove public preview alert from multihttp form

peterschretlen:
Multihttp is now GA, but we missed removing this alert


VikaCep (approved):
👍

Issue #762, opened 2024-04-24
Convert checks page to utilize URL parameters


ckbedwell:
**What would you like to be added**:
Currently the checks page utilizes localStorage as a way to replicate state. This is problematic for a number of reasons and should be removed. It should be updated by utilizing URL query parameters instead. 

**Why is this needed**:
The localStorage state can be a huge hinderance and unwanted behaviour in certain user workflows. Changing the state to url query parameters means we can persist check page state when navigating the app via back and forward buttons, and it also enables the ability to deep-link and share the checks list page from other applications as well as in peer-to-peer sharing.

Please ensure the solution is easily extensible as we add more filters in the future.

**Additional context**
[Product board link.](https://grafana.productboard.com/entity-detail/features/26915022)

rdubrock:
Fixing this would close #607

PR #797, opened 2024-05-28
Feat: Use url params to store state in Checks list page

VikaCep:
This PR implements applying list filters (such as search term, status, labels, probes), sorting selection and view type using the url query parameters instead of local storage. This brings a number of benefits such as being able to share the page maintaining the filters selection, being able to use the back/forth buttons, and also preventing unwanted behavior in certain scenarios. 

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/762


ckbedwell (changes requested):
I am super excited that this works and you got it up so quickly!

Just some feedback to address to make this as durable as we can and make it something that endures and is re-usable in the long-term.

If you want some reference material, [this is how it's been added in the k6-app](https://github.com/grafana/grafana-k6-app/blob/main/src/hooks/useQueryParameter.ts). By no means have to use it, just something to consider 😄

src/components/CheckList/CheckList.tsx
```diff
@@ -50,10 +44,14 @@ type CheckListContentProps = {
 const CheckListContent = ({ onChangeViewType, viewType }: CheckListContentProps) => {
   const { data: checks } = useSuspenseChecks();
   const { data: reachabilitySuccessRates = [] } = useChecksReachabilitySuccessRate();
-  const [checkFilters, setCheckFilters] = useState<CheckFiltersType>(getDefaultFilters());
+  const [checkFilters, setCheckFilters] = useQueryParametersState<CheckFiltersType>(
+    'checkFilters',
+    getDefaultFilters()
```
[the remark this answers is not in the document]
VikaCep: Great feedback Chris! I'm working on a refactor in order to address all your points. So far I removed the `useEffect` to set state which caused several problems as you mentioned, and added some flexibility to the hook in order to configure whether we want to push/replace on the history and provide encode/decode custom functions for each parameter, similar to the k6 implementation you shared. 
I'm still working on making a better implementation on the [filters param](https://github.com/grafana/synthetic-monitoring-app/pull/797/files#diff-2b2763ea08047fdb8d46e4a3ab0de7c316ed557ae5a0bd3b7ed232683bedbbadR52-R55) in order to improve the way they're stored, making it more human-readable. I'll also address preventing the page from crashing if parameters are wrong 💥, great catch!
I'll ask for re-review once I'm done with these changes 👍

src/components/CheckList/CheckList.tsx
```diff
@@ -64,13 +81,28 @@ const CheckListContent = ({ onChangeViewType, viewType }: CheckListContentProps)
   const isAllSelected = selectedCheckIds.size === filteredChecks.length;
   const totalPages = Math.ceil(filteredChecks.length / CHECKS_PER_PAGE);
 
-  const handleFilterChange = (filters: CheckFiltersType) => {
-    setCheckFilters((cf) => ({
-      ...cf,
-      ...filters,
-    }));
+  const handleFilterChange = (filters: CheckFiltersType, type?: FilterType) => {
```
ckbedwell, line 84: I think type isn't optional here, right? If you didn't provide a `type` nothing would happen?

src/hooks/useQueryParametersState.ts
```diff
@@ -0,0 +1,69 @@
...
+  decode?: (value: string) => ValueType;
+  strategy?: HistoryStrategy;
+}
+
+const useQueryParametersState = <ValueType>({
+  key,
+  initialValue,
+  encode = JSON.stringify,
+  decode = JSON.parse,
+  strategy = HistoryStrategy.Push,
```
ckbedwell, line 24: I think this should be `HistoryStrategy.Replace` by default. I love that we have the option to put it in the push history but I think my expectation as a user would be if I had gone to the checks page, done some filtering then hit back I'd go back to the page I've come from.

I could easily be convinced the other way, though. I know often in Grafana I use browser back after using the time range filter and I love that functionality.

What do you think?

ckbedwell (dismissed):
Great work! Pure 🔥🔥🔥🔥🔥🔥🔥

I'm approving even though I've left two minor comments, feel free to merge after you've either addressed or dismissed them= 😄

src/hooks/useQueryParametersState.ts
VikaCep: Sure thing! I missed this when changing the implementation. I agree with you. Fixed!

src/components/CheckList/CheckList.tsx
VikaCep: You're right, it shouldn't be optional. Changed it 👍

ckbedwell (approved):
🔥🔥🔥🔥🔥🔥

Issue #761, opened 2024-04-24
Change authentication options for HTTP checks to a radio button pattern


ckbedwell:
**What would you like to be added**:
During HTTP check creation in the authentication section there are checkboxes to enable / disable bearer authorization and basic authorization -- the implication being that you can send both in one request if you tick both boxes. This is not possible as they share the same header.

It should be converted to a radio button pattern with three options 'None', 'Bearer' and 'Basic'.

**Why is this needed**:
To improve the UX for creating HTTP checks.

**Additional context**
If a user fills in one of the inputs and then changes their option, the UI state should remain (e.g. if they filled in the input for username and password and then clicked on bearer auth, then back to basic, the username and password they typed should remain) but it should not be sent in the payload if they have selected none or a different option.

ckbedwell:
Spoken with @majavojnoska. Please check with her what the design of this should be (currently researching existing patterns).

PR #800, opened 2024-05-28
feat: use radio button group for http auth options

VikaCep:
Changes the way the HTTP Auth settings for checks work by only allowing to select either the "None", "Bearer" or "Basic" option, disallowing to select two at the same time by using a radio button group component. It also adds validation.

image

image

image

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/761


VikaCep:
@majavojnoska I made a first implementation but I'd really appreciate your feedback on it 🙏

majavojnoska:
> @majavojnoska I made a first implementation but I'd really appreciate your feedback on it 🙏

I like this approach, great job Vika!

src/components/CheckEditor/FormComponents/HttpCheckAuthentication.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (isBasicAuth) {
+      setValue('settings.http.authType', 'basic');
+      return;
+    }
+    if (isBearerAuth) {
+      setValue('settings.http.authType', 'bearer');
+      return;
+    }
+    setValue('settings.http.authType', 'none');
+  }, [isBasicAuth, isBearerAuth, setValue]);
```
ckbedwell, line 35: I don't think this needs a `useEffect`. It's effectively only used on initial load as the state it is handling is taken over by `handleChangeOption` so could be derived from the existing state instead.

ckbedwell (commented):
Great work, just a few places to tighten the bolts 🔩🔧   

I think you can either drop `settings.http.authType` completely and handle all of that behaviour in either local state with a ref to act as a memory cache for the unselected option (I can expand on that if you want) or move it to a UIState object on either  `CheckFormValuesBase` or `CheckFormValuesHttp` so it doesn't pollute the definition we have that is synced with backend 😃

src/types.ts
```diff
@@ -178,6 +180,7 @@ export interface HttpSettings {
   proxyConnectHeaders?: string[];
 
   // Authentication
+  authType?: HttpAuthType;
```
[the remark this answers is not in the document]
VikaCep: Great feedback! I changed the implementation to use refs instead of polluting these type definitions with something that's just frontend related as you suggested 👍

src/components/CheckEditor/FormComponents/HttpCheckAuthentication.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  const isBasicAuth = Boolean(
+    getValues(`settings.http.basicAuth.username`) || Boolean(getValues(`settings.http.basicAuth.password`))
+  );
+  const isBearerAuth = Boolean(getValues(`settings.http.bearerToken`));
+
+  const [authType, setAuthType] = useState<HttpAuthType>(isBasicAuth ? 'basic' : isBearerAuth ? 'bearer' : 'none');
+
+  const bearerTokenRef: MutableRefObject<string | undefined> = useRef(undefined);
+  const basicAuthRef: MutableRefObject<BasicAuth | undefined> = useRef(undefined);
```
ckbedwell, line 27: This is super nitty but just highlighting if you weren't aware that useRef can accept a generic 😄 

```suggestion
  const bearerTokenRef = useRef<string | undefined>(undefined);
  const basicAuthRef = useRef<BasicAuth | undefined>(undefined);
```

src/components/CheckEditor/FormComponents/HttpCheckAuthentication.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { MutableRefObject, useCallback, useRef, useState } from 'react';
```
ckbedwell, line 1:
```suggestion
import React, { useCallback, useRef, useState } from 'react';
```

Just in case you want to accept the suggestion below without checking out.

ckbedwell (approved):
Great job! 💪 ✅

ckbedwell (approved):
LGTM 🌟

PR #801, opened 2024-05-29
Homepage summary table: join on compound key job + instance

### LINKED ISSUE #798: Synthetic check page duplicated results
[_Full escalation ticket here._](https://github.com/grafana/support-escalations/issues/10708)

**Summary of Issue**:   
Customer is seeing their checks duplicated multiple times: 
image

Even though the check is configured only once
   
This is happening for all of their checks - above is an example of just one.   
   
I believe this is a frontend issue since metric queries seem to be working fine for what I could tell. 
   
Noticed that when I apply a filter, for instance `environment = devNet` then the checks are listed only once as expected.
   
Also noticed that for brand new checks this is not happening, I created a new check in their instance and could see it listed only once - it's been now deleted to not contribute to their billing.

peterschretlen:
If job name is reused across more than one check,   the join transformation in the homepage summary table does not have a key unique to each check. This is creating duplicate entries in the summary table.  This fixed the problem by concatenating `job` and `instance` into a key, and we do the inner join using that key. 

Fixes #798

Before:
image

After:
image


ckbedwell (approved):
LGTM

PR #803, opened 2024-05-30
Feat/probe form zod

ckbedwell:
# Problem

With moving over to zod schemas for our check forms it makes sense to create a universal experience for all of our forms for consistency.

# Solution

Add a ProbeSchema that the ProbeEditor component utilizes. It has the added benefit to make some of the error messaging a little clearer without adding any additional noise to the ProbeEditor component.


src/types.ts
```diff
@@ -87,11 +87,11 @@ export interface TCPQueryResponse {
...
 
 export interface BaseObject {
+  created?: number; // seconds
   id?: number;
+  modified?: number; // seconds
   tenantId?: number;
-  created?: number; // seconds
   updated?: number; // seconds
-  modified?: number; // seconds
 }
```
ckbedwell, line 95: If you are wondering what this change is about, I often select objects and alphabetise them because I get tired of looking up and down for keys I know exist 😅

Todo: add eslint rule that does it automatically.

VikaCep (approved):
Great work! 

Just one detail, it would be good to improve the message when we enter non-numeric values in the longitude field
image

It looks good for the Latitude one

image

VikaCep (approved):
✅ ✅ ✅

PR #811, opened 2024-06-05
fix: bring back id in checks link from homepage dashboard

VikaCep:
Fixes a regression from https://github.com/grafana/synthetic-monitoring-app/pull/801 where the check ID was missing from the path in the checks list, making it unable to load each check's individual dashboard.

**Before**
image

**After**
image

image

Reported in Slack https://raintank-corp.slack.com/archives/C0162C1K9E1/p1717612374226399

### DIFF
--- a/src/scenes/Summary/summaryTable.ts
+++ b/src/scenes/Summary/summaryTable.ts
@@ -117,6 +117,13 @@ function getSummaryTableQueryRunner(metrics: DataSourceRef, sm: DataSourceRef) {
     $data: queryRunner,
     transformations: [
       {
+        id: 'renameByRegex',
+        options: {
+          regex: 'target',
+          renamePattern: 'instance',
+        },
+      },
+      {
         id: 'groupBy',
         options: {
           fields: {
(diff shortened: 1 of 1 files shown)


src/scenes/Summary/summaryTable.ts
```diff
@@ -116,6 +116,13 @@ function getSummaryTableQueryRunner(metrics: DataSourceRef, sm: DataSourceRef) {
   const transformed = new SceneDataTransformer({
     $data: queryRunner,
     transformations: [
+      {
+        id: 'renameByRegex',
+        options: {
+          regex: 'target',
+          renamePattern: 'instance',
+        },
+      },
```
VikaCep, line 125: This is the only actual change, as @peterschretlen pointed out in Slack:
The checks data has the `id` we need for the path. But since it doesn't have a value for `instance`, (the field is instead called `target`) then it's not in the final results after going through all those transformations https://github.com/grafana/synthetic-monitoring-app/blob/main/src/scenes/Summary/summaryTable.ts

So, adding this transformation at the top will fix it. Otherwise, by the time the data reaches that last transformation, `target` is no longer there (it gets lost after the join)

ckbedwell:
Unfortunately this solution brings back _another_ regression. I'm going to add a custom transform to deal with this and we can revisit how we want to make Scenes dashboards more maintainable in the future. 


Notice in the table above that job and instance columns are being duplicated.

This inspired me to [create a utility function to help with debugging transformations in Scenes](https://github.com/grafana/synthetic-monitoring-app/pull/813) because without using something like this it is painful to work out what is going on.

ckbedwell:
image

I've added a custom transform to deal with this so it solves all the issues:
- 1. Originally reported duplication of job / instance rows
- 2. Linking to the job dashboard
- 3. Only rendering the `instance`, `job`, `check type`, `state`, `reachability` and `latency` columns.

Ideally I'd like to start adding some tests for our Scenes dashboards, so will add a ticket to do a POC. I have a feeling the main issue for this will lie in replicating and maintaining the database queries which msw would be providing.

On a separate note, I think moving forwards we should use custom transformations more liberally when maintaining Scenes dashboards. Frankly, the summary table transformation chain is still a huge maintainability problem.

ckbedwell (approved):

PR #813, opened 2024-06-06
feat: add scenes debugTransform function

ckbedwell:
# Problem

We've had some problems with the Scenes dashboards [1](https://github.com/grafana/synthetic-monitoring-app/pull/801), [2](https://github.com/grafana/synthetic-monitoring-app/pull/811) recently and the debugging process is very difficult due to the [extensive transforms](https://github.com/search?q=repo%3Agrafana%2Fsynthetic-monitoring-app%20SceneDataTransformer&type=code) that get added throughout the panels.

Slack message from Peter Schretlen saying 'I can see why people struggle with scenes.. this is really hard to debug.'

# Solution

I've added a simple utility function called `debugTransform` in the scenes folder which you can inject at any point during the transformation process to view what the DataFrames appear like at that point in the transform.

The above is an example of where I've added `debugTransform` three times in separate locations during the summaryTable transformation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,6 +127,7 @@
     "react-hook-form": "7.50.1",
     "react-popper": "^2.2.5",
     "react-router-dom": "^5.2.0",
+    "rxjs": "7.8.1",
     "sortablejs": "^1.15.0",
     "valid-url": "^1.0.9",
     "yaml": "^2.2.2",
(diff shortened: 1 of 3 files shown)


VikaCep (approved):
This is super helpful! 💯💯💯

Issue #805, opened 2024-05-31
Check list view: display and sorting check by test execution rate


peterschretlen:
**What would you like to be added**:

In addition to frequency, active series, and locations,  I would like to see the rate of test execution so I have an idea of how costly each a test is.

When trying to understand and manage costs, it would be helpful to sort the list by rate.  

**Why is this needed**:

When you have a large number of synthetic monitors,  it's hard to identify the ones that are running the most tests.  In the account shown in the screenshot I have 272 checks.  Which ones are generating the most volume? Which are the best candidates to scale back to save costs?  Being able to display and sort by execution rate helps answer these questions. 

<img width="1308" alt="image" src="https://github.com/grafana/synthetic-monitoring-app/assets/7903188/32e1a90d-d2fc-4c70-9d22-6366f3074522">

PR #814, opened 2024-06-06
Feat: Display test execution rate in check's list

VikaCep:
This PR adds a new indicator for each check card: the **number of test executions per month**.
Besides indicating the number of executions, it's also possible to sort by that value either in ascending or descending order. 

Why is it needed?
When trying to understand and manage costs, it would be helpful to sort the list by rate.

image

Note that it is the same number that we display in the "Probes" section when creating/editing a check. This provides consistency and makes it easy to translate that into a monthly cost.

image

Closes https://github.com/grafana/synthetic-monitoring-app/issues/805

### DIFF
--- a/src/checkUsageCalc.ts
+++ b/src/checkUsageCalc.ts
@@ -14,7 +14,7 @@ const getChecksPerMonth = (frequencySeconds: number) => {
   return Math.round(checksPerMonth);
 };
 
-const getTotalChecksPerMonth = (probeCount: number, frequencySeconds: number) => {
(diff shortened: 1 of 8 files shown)


ckbedwell (approved):
LGTM 💪 

I don't want to address this in this PR but this does highlight how much I want to re-design the checks list page and also make it responsive. Definitely something that will be coming up soon.

cc @majavojnoska

The check list page in the Synthetic Monitoring plugin. The new executions metric is cut off by the larger metrics to the right of it and unreadable at a screen size that would be similar to a 13

Issue #799, opened 2024-05-28
Update grafana/ui dependency 


VikaCep:
We should update the `grafana/ui` dependency to the latest version. Doing this will allow us to replace all `HorizontalGroup` and `VerticalGroup` components (which are getting deprecated soon https://github.com/grafana/grafana/issues/85510) in favor of `Stack` (https://grafana.com/developers/saga/components/primitives/stack/#:~:text=The%20Stack%20component%20is%20a,container%20either%20horizontally%20or%20vertically.)

We currently use `v10.1.4` and the latest version is `v11.0.0`.

PR #815, opened 2024-06-07
Fix: Update Grafana runtime dependencies

VikaCep:
Updating Grafana packages to their latest versions:

- `@grafana/ui` to version `11.0.0`
- `@grafana/data` to version `11.0.0`
- `@grafana/runtime` to version `11.0.0`
- `@grafana/schema` to version `11.0.0`

After upgrading, making sure that:

- [x] There are no unexpected errors
- [x] All tests pass
- [x] Replace `HorizontalGroup` and `VerticalGroup` components, which are deprecated, to use `Stack` instead.
- [x] See if we can remove this [noise](https://github.com/grafana/synthetic-monitoring-app/pull/791#discussion_r1629708272) in tests.

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/799


jest.config.js
```diff
@@ -7,7 +7,7 @@ const config = require('./.config/jest.config');
 module.exports = {
   // Jest configuration provided by Grafana scaffolding
   ...config,
-  setupFilesAfterEnv: ['<rootDir>/src/test/jest-setup.tsx'],
+  setupFilesAfterEnv: ['<rootDir>/src/test/encoder-workaround.tsx', '<rootDir>/src/test/jest-setup.tsx'],
```
ckbedwell, line 10: I'd just include the encoder-workaround in `jest-setup.tsx`.

Because there are so _many_ different ways to set-up a jest environment I've been working towards condensing as much as we can so the `src/test/jest-setup.tsx` file is the singular point of reference so it's easier to reason with rather than going:

1. Let me check what's in package.json
2. Let me check what's in `jest.config.js`
3. Let me check what folders exist named `__mocks__` and what's in them
4. Let me check what environment variables are
5. Let me check what's imported / mocked in individual tests

etc. 😄

package.json
```diff
@@ -98,13 +98,13 @@
...
+    "@grafana/data": "11.0.0",
     "@grafana/faro-web-sdk": "1.3.6",
     "@grafana/k6-test-builder": "^0.8.6",
-    "@grafana/runtime": "10.1.4",
+    "@grafana/runtime": "11.0.0",
     "@grafana/scenes": "3.5.0",
-    "@grafana/schema": "10.1.4",
-    "@grafana/ui": "10.1.4",
+    "@grafana/schema": "11.0.0",
+    "@grafana/ui": "11.0.0",
```
ckbedwell, line 107: 💪 💪 💪

src/components/CheckEditor/CheckEditorNew.test.tsx
```diff
@@ -22,6 +23,11 @@ jest.mock('hooks/useAlerts', () => ({
   }),
 }));
 
+jest.mock('@grafana/ui', () => ({
+  ...jest.requireActual('@grafana/ui'),
+  CodeEditor: () => <textarea />,
+}));
```
ckbedwell, line 29: Similar to the reasoning above, I'd do this globally similar to how we mock `@grafana/runtime`.

VikaCep:
@ckbedwell regarding the `invalid` prop on `Field` throwing a [warning](https://github.com/grafana/synthetic-monitoring-app/pull/791#discussion_r1629708272) when running tests, sadly it didn't go away. Even though the `Field` props defines it as `invalid?: boolean;`, for some reason the tests complain when we pass a boolean value to it.

ckbedwell:
> @ckbedwell regarding the `invalid` prop on `Field` throwing a [warning](https://github.com/grafana/synthetic-monitoring-app/pull/791#discussion_r1629708272) when running tests, sadly it didn't go away. Even though the `Field` props defines it as `invalid?: boolean;`, for some reason the tests complain when we pass a boolean value to it.

Yeah, don't necessarily trust the types 😅  I looked into `grafana/grafana` and [this is the culprit of the issue](https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Forms/Field.tsx#L107-L115) from what I can tell. I started a PR to fix this at its root but before I wrote a solution, when I wrote a test it didn't seem to be a problem in `grafana/grafana` itself. So it has me questioning why it gets flagged in our tests but not theirs... 🤔

VikaCep:
> Yeah, don't necessarily trust the types 😅 I looked into grafana/grafana and [this is the culprit of the issue](https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Forms/Field.tsx#L107-L115) from what I can tell. I started a PR to fix this at its root but before I wrote a solution, when I wrote a test it didn't seem to be a problem in grafana/grafana itself. So it has me questioning why it gets flagged in our tests but not theirs... 🤔

It's pretty weird as it only happens in tests that execute the `LabelField` component (which in turn includes `Field` and sets its `invalid` prop correctly as a `boolean`).
There are other places that are tested which directly call the `Field` component, but the warning is not shown in those cases (such as `CheckTarget.test.tsx`) So, it's super odd that it only happens when invoking the `LabelField`. I've managed to [remove the silence](https://github.com/grafana/synthetic-monitoring-app/pull/815/commits/bc354138ec792c77e12d3f2ce80cae23227efda9) by slightly changing the way it sets the `invalid` prop (sending `undefined` if there's no `labelError`), but it's still a mistery why that is needed.

__mocks__/@grafana/ui.tsx
```diff
@@ -1,6 +1,8 @@
 // @ts-ignore
 window.__react_router_build__ = undefined;
 
+import { TextEncoder } from 'util';
+global.TextEncoder = TextEncoder;
```
VikaCep, line 5: I don't like to add this here, but without it all tests will fail even though it's also present in `jest.setup.tsx`
image

Should I remove it from `jest.setup.tsx` and just leave it here? It doesn't seem like the best place though. The odd thing is this was not needed when including the workaround in `setupFilesAfterEnv` in a separate file (https://github.com/grafana/synthetic-monitoring-app/pull/815/commits/5a6e7b2174fd5905d480151bf4fcfafc6968f861)

src/test/silenceErrors.ts
```diff
@@ -4,6 +4,7 @@ const originalConsoleError = console.error;
 const ignoreErrorsList = [
   `Could not parse CSS stylesheet`,
   `Using kebab-case for css properties in objects is not supported.`,
+  `Warning: An update to %s inside a test was not wrapped in act(...).`,
```
VikaCep, line 7: This started to happen a lot after upgrading deps. It seems to happen in all tests that deal with forms (calling https://github.com/grafana/synthetic-monitoring-app/blob/upgrade-deps/src/components/CheckEditor/testHelpers.ts#L31) but it doesn't seem trivial to fix.

jest.config.js
VikaCep: I moved it there, but if I do that it will fail if it's not present in the `grafana/ui` mock so I had to add it there as well: https://github.com/grafana/synthetic-monitoring-app/pull/815#discussion_r1635279546

__mocks__/@grafana/ui.tsx
VikaCep: In the end I removed it from here as importing the `encoder-workaround` file directly from `jest-setup.tsx` worked just fine.

ckbedwell (approved):
LGTM! 🔥🔥🔥

I was super intrigued by the `act` error regression so did a bit of digging and found a solution [so put a separate PR up](https://github.com/grafana/synthetic-monitoring-app/pull/820). It was the right call not to fix it in this PR 😄

ckbedwell (approved):

Issue #795, opened 2024-05-24
Consider using a `password`-typed input for HTTP check authentication password


roobre:
**What would you like to be added**:

Currently the password form field for authentication for HTTP checks shows its value in clear text:

![image](https://github.com/grafana/synthetic-monitoring-app/assets/969721/e0c6a41c-dfec-45ed-9c84-9c3a596cbdce)

This is good for ensuring the correct value has been entered, but also shows the value in clear text when adding/editing the check.

**Why is this needed**:

Having the field obscured by default (potentially with a toggle to reveal it) can be useful to prevent secret exfiltration when working in public environments or screen sharing.

---

Originally reported in [Slack](https://raintank-corp.slack.com/archives/C0162C1K9E1/p1716500522607259).

PR #819, opened 2024-06-11
fix: change password type to input on http auth config

VikaCep:

The Authentication step for HTTP checks allows to configure a username and password. This password field was displaying the value entered by the user in clear text, posing a security concern.

This PR adds a `PasswordField` component that allows to mask/show the value in the input and uses it in Auth/bearer and password fields.

**Before**
image

**After**
image

image

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/795


ckbedwell (commented):
I think this could be a good opportunity to create a reveal password / secret re-usable component. What do you think?

I have no idea why[ this component](https://github.com/grafana/grafana/tree/main/public/app/core/components/PasswordField) in `grafana/grafana` isn't in the UI folder for us to use...

ckbedwell (approved):
LGTM 🔥🔥🔥

PR #820, opened 2024-06-12
chore: update test dependencies to remove act error

ckbedwell:
# Problem

The [upgrade to the latest grafana packages](https://github.com/grafana/synthetic-monitoring-app/pull/815) introduced a regression with the tests warning incorrectly about `act` errors which we manually silenced.

# Solution

Upgrade the `@testing-library` dependencies to the latest.

Interestingly, with the update from `@testing-library/jest-dom` from version 5 to 6 it meant we have to re-import `@testing-library/jest-dom` in the `.tsx` jest setup file so we correctly get the types even though it is included in the `.js` jest-setup file provided by the create plugin scaffolded folder.


VikaCep (approved):
I'm so glad you figured this out! 🎉🎉🎉

PR #828, opened 2024-06-18
fix: displaying scenes when applying custom label filters

VikaCep:
In the Synthetic Monitoring Home dashboard, adding filters results in some of the panels (`errorPctTimeseries`, `latencyTimeseries` and `errorRateMap`) not returning data. This happens because the query of each panel is constructed incorrectly by adding the custom label filter in the wrong place in the expression as described here
https://github.com/grafana/support-escalations/issues/10975#issuecomment-2168583743. 

If the datasource is set to "Mixed" mode, the problem doesn't happen. This is the reason the table does return results when filtering by labels. _This solution seems like a workaround of the real problem though._

EDIT: After some more research I figured out the real problem. For custom labels we're using scenes [Adhoc Filters](https://grafana.com/developers/scenes/advanced-adhoc-filters). These filters, are automatically applied to all queries in a datasource by default. In our case, we **only** want to apply it to the `sm_check_info` query. For this reason this kind of filters must be set to [manual mode](https://grafana.com/developers/scenes/advanced-adhoc-filters#manual-mode). By setting it to manual mode then we can use them easily in specific queries as a normal variable.

**Before:**

**After:** 

Fixes https://github.com/grafana/support-escalations/issues/10975

### DIFF
--- a/src/scenes/Summary/errorPctTimeseries.ts
+++ b/src/scenes/Summary/errorPctTimeseries.ts
@@ -38,13 +38,9 @@ function getErrorPercentageQuery() {
 
 function getErrorPercentageQueryRunner(metrics: DataSourceRef) {
   const queryRunner = new SceneQueryRunner({
-    datasource: {
-      uid: '-- Mixed --',
-      type: 'datasource',
-    },
+    datasource: metrics,
     queries: [
       {
-        datasource: metrics,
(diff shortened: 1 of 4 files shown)


peterschretlen (approved):
Nice find, thanks @VikaCep!

PR #840, opened 2024-06-27
fix: change query for uptime stat panel

VikaCep:

This PR fixes the query that is used to calculate the uptime stat panel for each check. As described in this [escalation](https://github.com/grafana/support-escalations/issues/11197), this panel was displaying a value of `100%` even when the probes had 100% errors reaching the target.

Part of https://github.com/grafana/support-escalations/issues/11197. This solution addresses the first problem described here https://github.com/grafana/support-escalations/issues/11197#issuecomment-2195657543

### DIFF
--- a/src/scenes/Common/uptimeStat.ts
+++ b/src/scenes/Common/uptimeStat.ts
@@ -13,40 +13,45 @@ function getMinStep(minStep: string) {
   }
 }
 
-function getQueryRunner(metrics: DataSourceRef, minStep: string) {
+function getQueryRunner(metrics: DataSourceRef, minStep: string, newUptimeQuery: boolean) {
   // The min step for most queries is a minimum of 1 minute. For uptime, however, we want to make sure we have steps of at least 5 minutes in order for the math to work out.
   const uptimeMinStep = getMinStep(minStep);
+
+  const uptimeCalculationQueryV1 = `# the inner query is going to produce a non-zero value if there was at least one successful check during the 5 minute window
+    # so make it a 1 if there was at least one success and a 0 otherwise
+    ceil(
+      # the number of successes across all probes
+      sum by (instance, job) (increase(probe_all_success_sum{instance="$instance", job="$job"}[$__rate_interval]))
+      /
+      # the total number of times we checked across all probes
+      (sum by (instance, job) (increase(probe_all_success_count{instance="$instance", job="$job"}[$__rate_interval])) + 1) # + 1 because we want to make sure it goes to 1, not 2
+    )`;
+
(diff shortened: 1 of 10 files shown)


VikaCep:
@mem I changed the query according to your comment's solution.

src/scenes/Common/uptimeStat.ts
```diff
@@ -22,15 +22,31 @@ function getQueryRunner(metrics: DataSourceRef, minStep: string) {
...
-          sum by (instance, job) (increase(probe_all_success_sum{instance="$instance", job="$job"}[$__rate_interval]))
-          /
-          # the total number of times we checked across all probes
-          (sum by (instance, job) (increase(probe_all_success_count{instance="$instance", job="$job"}[$__rate_interval])) + 1) # + 1 because we want to make sure it goes to 1, not 2
-        )`,
+        expr: `
+        sum_over_time(
+ # the inner query is going to produce a non-zero value if there was at least one successful check during the 5 minute window
+ # so make it a 1 if there was at least one success and a 0 otherwise
+ ceil(
```
mem, line 29: I think this side of the query is missing a `sum_over_time`.

The idea is to compute the number of times at least one probe reported success and add that over the entire range (sum_over_range), and that number should be divided by the number of observations made (count_over_time).

mem:
I'm sorry, I just realized this is the query we had to fix for a different reason.

I think the query itself is correct. We had to add a transformation that collects the data for the entire range because Mimir has a limit on the amount of data that it's willing to retrieve.

I was confused because the "explore" link in the panel drops that transformation.

Let me take another looks.

ckbedwell:
https://github.com/grafana/support-escalations/issues/11197#issuecomment-2218047905 -- update as of 2024/07/09

VikaCep:
I added a new uptime query version from @mem's [proposal](https://docs.google.com/document/d/1ezU4CSUmQUjhSSQ6glarSHMVMKZ8F3oz0TjdKkCYGaM/edit#heading=h.vff2v3vanyg7
):

```
floor(
      # Report a 1 if there's a location where most observations were successful and 0 if most observations failed for all probes.
      max by (instance, job) (
        round(
          # the number of successes for each probe
          (increase(probe_all_success_sum{instance="$instance", job="$job"}[$__rate_interval]))
          /
          # the total number of times we checked for each probe
          ((increase(probe_all_success_count{instance="$instance", job="$job"}[$__rate_interval])))
        )
      )
    )
```

It is under a feature flag named `uptime-query-v2`. By default, we're using the original query.

peterschretlen (approved):
LGTM 🎉 , thanks @VikaCep!

PR #842, opened 2024-07-01
fix: cursor type on cards without links

ckbedwell:
# Problem

I noticed in the [non-initialized pages improvements PR](https://github.com/grafana/synthetic-monitoring-app/pull/839) that cards without links are displaying the hover cursor incorrectly.

# Solution

Remove the `cursor: pointer` from the unneeded places and add it to the `:after` pseudo element on the title link.


VikaCep (approved):
Nice!!

PR #843, opened 2024-07-01
chore: enforce named exports

VikaCep:
- Added a `eslint` rule to enforce named exports
- Replaced existing usages of default exports to named exports.

This was discussed as an enhancement here https://github.com/grafana/synthetic-monitoring-app/pull/839#discussion_r1661022157


ckbedwell (approved):
👏 🔥🔥🔥

PR #847, opened 2024-07-03
Check form redesign

ckbedwell:
**Very very aware of what a bad dev citizen I have been with this branch. I am just putting my hands up and will keep apologising over and over for the amount of files touched and the undisciplined commit history. I don't expect it to be reviewed in depth so that's why I wanted to make sure it had comprehensive test coverage.**

## Problem

We have an [internal document](https://docs.google.com/document/d/1wnAPCC5l17ZL1--oWfXq121oYvwKOMIYvC8XHDWXQEo/edit#heading=h.am5j7ij4fl53) outlining the main problems with the current check creation forms with some discussion from the SM team. The main problems with the current flow are:
- Differing amount of steps according to check type which in most cases come across as very daunting (e.g. 8 steps for Http)
- Disjointed steps where depending on what type of check you are the same settings are located in different places
- Related options are fragmented, in particular the options where users 'define uptime' are across multiple steps
- Difficult DX architecture to work with and scale for the future

## Solution

This PR includes a comprehensive solution to address all of the points above:
- It reduces cognitive overhead when choosing a check type at the start of the process
- It prioritises what is important when creating checks (making requests) and laying the foundations for a tighter feedback loop for testing requests.
- Creates a consistent and predictable experience regardless of check type (5 steps which are always the same)
- Creates separation of concerns for the larger contributing components (e.g. `<CheckForm />`, `<FormLayout />`, etc.) 
- Creates reusable components which follow a familiar design pattern (e.g. `<Request />`)
- Adds a comprehensive testing solution for check creation pages
- Introduced `<CheckFormContextProvider />` -- in particular this is used for sharing `isFormDisabled` but it will also be the mechanism for sharing state when testing individual requests

### Screenshots

**New check creation page**
Screenshot of the new check creation screen with the three over-arching groups: Api Endpoint, Multi step and Scripted

**New api endpoint check creation form**
Screenshot of the new api endpoint check creation form.

**New multistep check creation form**
Screenshot of the new multistep check creation form.

**New multistep check creation form on the define uptime step**
Screenshot of the new multistep check creation form at the define uptime step.

**New scripted check creation form**
Screenshot of the new scripted check creation form.

### Test coverage
I've created and added a meta test framework for the check creation pages. The aim is that we now have individual and specific tests for every individual aspect of the check creation forms. They are full integration tests so we have high confidence that any regressions for any part of the form will be targeted specifically and throw an error and be caught.

Screenshot of VSCode showing an open window and a terminal. The open window shows the diff for DNSCheckRecordPort where the input name has been modified incorrectly. The terminal shows the test suite for /ApiEndPointChecks/DNSCheck/1-request.payload.test.tsx has specifically caught the problem and thrown an error.

Jest test runners don't work well and often timeout when a single test file has many individual tests, so the goal of this test framework is to subdivide the tests into sensible and predictable files where it is easy to pinpoint and locate the tests you are looking for.

All of the tests are located in two `__tests__` folders under `page/NewCheck` and `page/EditCheck`. Each of these folders has subfolders corresponding to the top-level CheckTypeGroups which then contain folders for the CheckType. The file naming convention for each tests follows the following pattern:

```
{SectionNumber-{SectionName}_{?subSection}.{payload|ui}.test.tsx
```

The subsections are optional and necessary to reduce the amount of tests per file. (Http checks have a lot of options in the first section so were proving to be a little flaky and timing out in the CI pipelines). 

A screenshot showing the folder/file structure for /NewCheck/__tests__/ApiEndPointChecks/HTTPCheck.

**Before**
```
Test Suites: 40 passed, 40 total
Tests:       249 passed, 249 total
Snapshots:   0 total
Time:        40.533 s
Ran all test suites.
✨  Done in 41.27s.
```

**After**
```
Test Suites: 73 passed, 73 total
Tests:       3 skipped, 368 passed, 371 total
Snapshots:   0 total
Time:        77.738 s, estimated 100 s
Ran all test suites.
✨  Done in 78.28s.
```

### Whilst I was there
I was debating having a little more discipline and leaving this cleanup and tech debt work to a future PR, but as I was working with all these files and having a bad DX experience with them it was difficult not to just correct them whilst I was there rather than checking out and making separate branches. Many of the changed files in this diff are:

- Changing `styles.stack` and `styles.stackCol` to `<Stack />` and `<Stack direction="column" />`
- Alphabetising props so it was easier to find what I was looking for
- Adding aria-labels to inputs which were missing them
- Removing unnecessary props to components that didn't use/need them (e.g. `disabled` on the `<Field />` components)

### Further work

1. The request components should be fully decoupled from `react-hook-form`. Currently they are difficult to re-use but with a little work their props could be altered to make them 'dumber' and more reusable for use across the application (e.g. like in helping create requests for scripted checks).

2. More test coverage for existing checks. Currently the tests focus on being able to submit new values to the api in the correct shape, but we don't have any tests ensuring that existing data is getting transformed and displayed in the UI correctly. This work shouldn't be addressed until [this ticket creating check object factory](https://github.com/grafana/synthetic-monitoring-app/issues/850) has been completed.

3. More test coverage for read-only checks. As each input has to be individually disabled, it could be easy in future work to miss adding a disabled state when `isFormDisabled` is true. I have added a single test for this ensuring the submit button does not work as that is the most element to disable.

4. Add redirects from the old URLs to the new ones where appropriate. This will be in a PR put up shortly.

5. Add nested error discovery when submitting the form. At the moment nested errors are difficult to locate when submitting the form. Link to PR: https://github.com/grafana/synthetic-monitoring-app/pull/868

6. I have left in several unused components (e.g. `useTestRequests`, `<FormSupportingContent />`) and commented out lines of code which will be reintroduced in the near future (scripted check card mentioning support for gRPC and extension support).

### Extra features

Some of these existed already but I've tidied up the UI and experience to ensure it is as smooth as possible.

**Scripted check limit reached**
Screenshot of the scripted check creation page showing an error message that the script limit has been met. The inputs for the form are also disabled.

Screenshot of the choose a check type page showing an error message that the script limit has been met. The scripted CTA button is disabled.

**Loaded checks but couldn't find the one requested (e.g. bad URL)**
Screenshot of the check editing page with a modal over the top of it saying it couldn't find the check with the option to go back to the checks screen.

**Unable to load check(s) (e.g. 500 server error)**
Screenshot of the check editing page with a modal over the top of it saying we were unable to load the check. It has options to go back to the checks list or retry the request.


src/components/CheckEditor/transformations/toPayload.ping.ts
```diff
@@ -18,6 +18,7 @@ const getPingSettings = (settings: Partial<PingSettingsFormValues> | undefined =
 
   return {
     ...fallbackValues,
+    dontFragment: settings.dontFragment ?? fallbackValues.dontFragment,
```
ckbedwell, line 21: This highlights the need of why we need comprehensive testing 😅

src/components/CheckEditor/transformations/toFormValues.grpc.ts
```diff
@@ -14,10 +17,12 @@ export function getGRPCCheckFormValues(check: GRPCCheck): CheckFormValuesGRPC {
...
 
 const getGRPCSettingsFormValues = (settings: GRPCCheck['settings']): GRPCSettingsFormValues => {
+  const transformedTlsConfig = getTlsConfigFormValues(settings.grpc?.tlsConfig);
+
   return {
     ipVersion: settings.grpc?.ipVersion,
     service: settings.grpc?.service,
     tls: settings.grpc?.tls,
-    tlsConfig: settings.grpc?.tlsConfig,
+    ...transformedTlsConfig,
```
ckbedwell, line 26: And here 😅 (this also needs some tests to catch it as mentioned in the future work part of the PR).

VikaCep:
Seems like the button to go to the next step in the form is showing the current step name instead of the next one. For example, in the screenshot we see `3. Define uptime` but the step 3 is called `Labels`

image

VikaCep:
Not to do here, but just making a note that we'll have to go through the docs and update the images that are outdated after these changes, like:
imageimage

src/components/CheckEditor/FormComponents/RequestBodyContentType.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { CheckFormValues } from 'types';
+
+interface RequestBodyContentTypeProps {
+  disabled?: boolean;
+  name: FieldPath<CheckFormValues>;
+}
+
+export const RequestBodyContentType = ({ disabled, name }: RequestBodyContentTypeProps) => {
+  const { register } = useFormContext<CheckFormValues>();
+  const id = useId().replace(/:/g, '_');
```
VikaCep, line 14: This is done in several places, should we move it to a common function/hook?

src/components/CheckEditor/FormComponents/Timeout.tsx
```diff
@@ -0,0 +1,71 @@
...
+          step={1}
+          validate={(value) => validateTimeout(value, maxTimeout, minTimeout)}
+        />
+      )}
+    </Field>
+  );
+};
+
+function getTimeoutBounds(checkType: CheckType) {
+  if (checkType === CheckType.Traceroute) {
```
VikaCep, line 55: Just thinking that the `Timeout` component could receive its `min` and `max` values as props instead of relying on the `checkType`.  Not a big deal though.

src/components/CheckEditor/FormComponents/VariablesFields.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { Controller, FieldPath, useFieldArray, useFormContext } from 'react-hook-form';
+import { GrafanaTheme2 } from '@grafana/data';
+import { Button, Field, IconButton, Input, Select, Stack, useStyles2 } from '@grafana/ui';
+import { css } from '@emotion/css';
+
+import { CheckFormValuesMultiHttp, MultiHttpVariableType } from 'types';
+import { useCheckFormContext } from 'components/CheckForm/CheckFormContext/CheckFormContext';
+import { MULTI_HTTP_VARIABLE_TYPE_OPTIONS } from 'components/constants';
+
+export const VariablesFields = ({ index }: { index: number }) => {
```
VikaCep, line 11: Should we name this component something more specific to MultiHTTP? Sounds too generic otherwise.

VikaCep (commented):
Amazing work! 🔥🔥🔥 
Left a few comments, I still want to take another look but I really love how this is looking, components are much better organized and I really like the consistency between checks. Also, superb work on improving and adding tests!

ckbedwell:
> Not to do here, but just making a note that we'll have to go through the docs and update the images that are outdated after these changes, like: imageimage

@heitortsergent Tagging you here but will drop a message on Slack, too. We can work out what will need changing in the documentation. Thanks for pointing this out, @VikaCep!

ckbedwell:
> Seems like the button to go to the next step in the form is showing the current step name instead of the next one. For example, in the screenshot we see `3. Define uptime` but the step 3 is called `Labels`
> 
> image

Good catch! This is because one of the last things I added was the alerts section above the form and it was being counted when working out what the next section was so the count was off by 1. I've fixed this [added some tests](https://github.com/grafana/synthetic-monitoring-app/pull/847/commits/646982fb6dbc27dd531c3b291f9722c8891c2cff#diff-bd8841b096319e22157a48928f7bd0eaa8240941aaef35e4df592a9ce5599b6a) to make sure it doesn't regress 😄

src/components/CheckEditor/FormComponents/ChooseCheckType.tsx
```diff
@@ -0,0 +1,114 @@
...
+};
+
+const CheckBadge = ({ status }: { status: CheckStatus }) => {
+  const { text, color } = colorMap[status];
+
+  return <Badge text={text} color={color} />;
+};
+
+// so the text doesn't bounce up and down when there area a mix of badges / no-badges
+const BadgePlaceholder = () => <div style={{ height: `22px` }} />;
```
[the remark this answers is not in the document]
ckbedwell: Yeah, [it is in the absence of a badge](https://github.com/grafana/synthetic-monitoring-app/pull/847/files#diff-aff30bba4ba5499809b5c11ee53eb5994a496c1107c92b7f4ab6b3c199d985ddR75). If it is removed when you click between gRPC and another check type the text 'bounces' up and down.

src/components/CheckEditor/FormComponents/HttpCheckValidStatusCodes.tsx
```diff
@@ -1,24 +1,23 @@
...
+  const { isFormDisabled } = useCheckFormContext();
   const id = 'validStatusCodes';
 
   return (
     <Field
       htmlFor={id}
       label="Valid status codes"
       description="Accepted status codes for this probe. Defaults to 2xx."
-      disabled={!isEditor}
+      disabled={isFormDisabled}
```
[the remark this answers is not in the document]
ckbedwell: Good catch. I've gone through and removed all of the disabled props in the `<Field />` component and added them to the input components directly. I've also started to alphabetise more of the props so it is easier to scan through them (eslint rule might be incoming...)

src/components/Request/RequestTest.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface RequestTestProps {
+  disabled?: boolean;
+  label?: string;
+  onClick?: () => void;
+}
+
+export const RequestTest = ({ disabled, label = `Test`, onClick }: RequestTestProps) => {
+  // todo: temporarily disabling the test button for all requests
+  if (!onClick || true) {
```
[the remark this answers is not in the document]
ckbedwell: When we start work on the supporting content for the check creation forms this will be added back in 😄 I've updated the comment to reflect that.

src/components/CheckEditor/FormComponents/RequestBodyContentType.tsx
ckbedwell: Done, added `useDOMId` with a comment explaining its purpose.

src/components/CheckEditor/CheckEditor.types.ts
```diff
@@ -0,0 +1,74 @@
+import { FormEvent } from 'react';
```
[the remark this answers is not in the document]
ckbedwell: Yeah, good point! I was thinking as Thomas is out, we should maybe bump up this ticket and tackle it ASAP to cause the least disruption: https://github.com/grafana/synthetic-monitoring-app/issues/848. That way we can address this problem at the same time.

I'll put it in our 1:1 notes and we can discuss it on Wednesday 😃

src/components/CheckEditor/CheckEditor.types.ts
```diff
@@ -0,0 +1,74 @@
...
+  queryParams?: FieldProps;
+};
+
+export type DNSRequestFields = {
+  target: FieldProps;
+  ipVersion: FieldProps;
+  recordType: FieldProps;
+  server: FieldProps;
+  protocol: FieldProps;
+  port: FieldProps;
```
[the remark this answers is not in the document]
ckbedwell: Good point, fixed this! 🔧

src/components/CheckEditor/FormComponents/VariablesFields.tsx
ckbedwell: Done, good suggestion!

src/components/CheckEditor/FormComponents/HttpRequest.tsx
```diff
@@ -0,0 +1,246 @@
...
+                    tabSelectsValue={false}
+                  />
+                </div>
+              );
+            }}
+            name={fields.method.name}
+          />
+          <Request.Input
+            aria-label={fields.target['aria-label'] || `Request target *`}
+            data-fs-element="Target input"
```
VikaCep, line 77: These two props seem to already be set in `RequestInput` by default, so it looks like they could be removed from here.

VikaCep (approved):
👏 👏 👏 👏 👏

PR #852, opened 2024-07-03
fix: regression for longitude validation fix

VikaCep:
Adds a regression test for the fix on longitude validation on creating new probes in https://github.com/grafana/synthetic-monitoring-app/pull/836

### DIFF
--- a/src/page/NewProbe/NewProbe.test.tsx
+++ b/src/page/NewProbe/NewProbe.test.tsx
@@ -35,8 +35,13 @@ test(`Doesn't show a validation error for valid longitude values`, async () => {
   const { user } = renderNewProbe();
   const saveButton = await screen.findByText('Add new probe');
   const longitudeInput = await screen.findByLabelText('Longitude');
-  await user.type(longitudeInput, '116.3971');
+  await user.type(longitudeInput, '180.01');
   await user.click(saveButton);
+
   const errorMsg = await screen.queryByText('Longitude must be less than 180');
+  expect(errorMsg).toBeInTheDocument();
+
+  await user.clear(longitudeInput);
+  await user.type(longitudeInput, '116.3971');
   expect(expect(errorMsg).not.toBeInTheDocument());
 });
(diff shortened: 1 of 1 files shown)


mem (approved):

src/page/NewProbe/NewProbe.test.tsx
```diff
@@ -29,3 +29,14 @@ it(`creates a new probe, displays the modal and redirects on close`, async () =>
...
 });
+
+//regression for https://github.com/grafana/support-escalations/issues/11171
+test(`Doesn't show a validation error for valid longitud values`, async () => {
+  const { user } = renderNewProbe();
+  const saveButton = await screen.findByText('Add new probe');
+  const longitudeInput = await screen.findByLabelText('Longitude');
+  await user.type(longitudeInput, '116.3971');
+  await user.click(saveButton);
+  const errorMsg = await screen.queryByText('Longitude must be less than 180');
```
ckbedwell, line 40: Tests for asserting when something _isn't present_ are really easily to accidentally end up with false positives (like when you accidentally completely break your validation so error messages never appear when they are supposed to...), so I like to either include in the test an assertion that the error validation appears correctly and then after fixing it the error leaves the DOM, or have an accompanying test which asserts for its presence:

```suggestion
  await user.type(longitudeInput, '180.01');
  await user.click(saveButton);

  const errorMsg = await screen.queryByText('Longitude must be less than 180');
  expect(errorMsg).toBeInTheDocument();

  await user.clear(longitudeInput);
  await user.type(longitudeInput, '116.3971');
```

src/page/NewProbe/NewProbe.test.tsx
```diff
@@ -29,3 +29,14 @@ it(`creates a new probe, displays the modal and redirects on close`, async () =>
...
+  expect(expect(errorMsg).not.toBeInTheDocument());
```
ckbedwell, line 41:
```suggestion
  expect(errorMsg).not.toBeInTheDocument();
```

ckbedwell (changes requested):
Just a few suggestions but love that you've added this! 💪

ckbedwell (approved):
🔥 🔥🔥

Issue #832, opened 2024-06-19
Test button "times out" on a long running scripted check, even though logs are sent


peterschretlen:
**What happened**:

When using the test button on a new or existing scripted check, the UI sometimes says there is a timeout even though the test succeeded.  This seems to happen most often with long running checks. 

The test button works by:
1. Sending an "adhoc" check to the probe with a specific ID like d48cf85d-35a5-44b1-a8f0-ef800409fe84
2. The probe then sends a log message with the results, that includes the ID.
3. The UI then makes an instant query to loki for the log message.  The query typically looks like this
 ` {type="adhoc"} |= "d48cf85d-35a5-44b1-a8f0-ef800409fe84" `

In some cases the logs are being sent.  But the logs query in the UI never returns any results, and eventually the UI gives up with a "timed out waiting for a response" error

**What you expected to happen**:

If the logs are sent, we should show the results and not a timeout

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a new scripted check
2. Set the timeout to 1 minute
3. Tweak the default script to the following, adding a sleep of 40 seconds:
```
import { check, sleep } from 'k6'
import http from 'k6/http'

export default function main() {
  const res = http.get('http://test.k6.io/');
  sleep(40);
  // console.log will be represented as logs in Loki
  console.log('got a response')
  check(res, {
    'is status 200': (r) => r.status === 200,
  });
}
```

Result: 

If I query in explore for `{type="adhoc"} |= "cf9db6ee-97ed-4391-b9df-a008dbe476db"` (replacing the ID with the one generated for the adhoc query), I will get a result in about 45 seconds:
<img width="1334" alt="image" src="https://github.com/grafana/synthetic-monitoring-app/assets/7903188/26e4e192-5da5-4dd2-bee6-a8215274f15b">

However the UI never gets a result, and eventually times out
<img width="888" alt="image" src="https://github.com/grafana/synthetic-monitoring-app/assets/7903188/fc5103bb-d06b-48a0-9a01-e7f7ea4f5d4b">

This happens even though they are performing the same query.  The notable difference is explore does a range query by default, whereas the UI does an instant query.

PR #855, opened 2024-07-05
fix: make loki query fetch range instead of instant logs

VikaCep:
Instant queries sometimes don't return log data when executing scripts that take a long time. In these cases it's better to use range queries. 
This PR changes the query used to fetch logs when testing ad-hoc checks to be of type `range`. Also, we were sending a fixed `end` value that didn't update to the current time on consecutive requests, staying always in the past. By removing the parameter in the request, it will default to the current `$now` value (https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-logs-within-a-range-of-time).

https://github.com/grafana/synthetic-monitoring-app/assets/6271380/7f067ccb-6989-4519-8b51-212c4f11fca5

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/832

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -213,22 +213,20 @@ export const queryLogsLegacy = async (
   url: string,
   query: string,
   start: number,
-  end: number
+  end?: number // defaults to now
 ): Promise<LogQueryResponse> => {
   const backendSrv = getBackendSrv();
   const params = {
-    direction: 'BACKWARD',
     limit: 1000,
     query,
-    start: start,
-    end: end,
-    // step: ,
+    start,
+    end,
   };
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -213,22 +213,20 @@ export const queryLogsLegacy = async (
   url: string,
   query: string,
   start: number,
-  end: number
+  end?: number // defaults to now
 ): Promise<LogQueryResponse> => {
   const backendSrv = getBackendSrv();
   const params = {
-    direction: 'BACKWARD',
```
VikaCep, line 220: `backward` is the default value

PR #856, opened 2024-07-09
chore: Add datasource to includes

wbrowne:
In order to support eventually loading the synthentic monitoring appn assets from a CDN, the plugin.json needs to list all nested plugins in the includes.


ckbedwell (approved):
LGTM 🌟

Issue #848, opened 2024-07-03
Tech debt: research, document and enact a consistent file/folder naming convention


ckbedwell:
_This is a broad ticket and may be best approached by dividing it up into sub-tasks. There may also be an opportune time to coordinate and do the bulk of this reorganisation depending on what other work is going on simultaneously due to how disruptive it will be to existing and on-going PRs._

E.g. this work wouldn't be appropriate to enact until the [check creation redesign PR is merged](https://github.com/grafana/synthetic-monitoring-app/pull/847). 

# Problem

As the SM plugin has grown and multiple developers have worked on it across the years, the file and folder structure and organisation is very inconsistent and difficult to reason with and find what you are looking for. It adds to cognitive overload and slows down developer velocity.

A non-exhaustive list of the current issues / questions we face in organisation:
- In some places we utilise folders, in some places we have a flat structure
- Some features have separate utility / style files, whilst others contain them all in a single file
- Some imports use absolute paths, some use relative paths
- Some files use a prefix naming solution. Some use a suffixed naming solution. Some are implicit under what folder they exist.
- Components folder has 'universal' components and feature/page-specific components within it
- We have a hooks folder and a data folder. The data folder has many hooks in it which are data-fetching specific. Is it clear what each folder is for and where to find the appropriate hook?
- We have multiple `__mocks__` folders which are buried within child folders. Should we continue to use these or remove them completely and add these mocks elsewhere?

# Completion criteria

- [ ] Research and propose a consistent file/folder organisation structure
- [ ] Research and propose a consistent naming convention
- [ ] Document the agreed upon proposals and add to the developer documentation for reference within the plugin
- [ ] Either create a plan to carry out the work piecemeal or coordinate to do the work in one huge organisation PR

## Additional context

- Is referencing and utilising Bulletproof React the right approach?
- Are there other industry standard practices we should follow?
- Should we base it as closely as possible to what the [k6-app is doing for consistency](https://github.com/grafana/grafana-k6-app)?

PR #868, opened 2024-07-16
feat: nested error discovery in check form

ckbedwell:
## Problem

As noted in the check form redesign work, this is required to help with error discovery: https://github.com/grafana/synthetic-monitoring-app/pull/847

With the wizard form pattern and now with the request components, it means that users may have difficulty finding errors to correct when submitting check creation / updates.

## Solution

Add methods and event listeners which respond to an invalid check form submission. Because of the nature of the individual components which need to respond to the invalid submission it doesn't make sense to rearchitect them to respond via prop drilling or some overarching context. I have chosen to do this a little bit more vanilla with dispatching a custom event, setting up event listeners for the custom event and utilising [`useImperativeHandle`](https://react.dev/reference/react/useImperativeHandle) where appropriate.

I have added test coverage for each of the check types which is doing a single spot check that the field with an error is being navigated to. I did have checks ensuring the input field was receiving focus but I have had to remove them as they were proving a little flaky in the CI/CD. I have left two specific focus checks for the script and probe fields as they are special and need to handled differently to all other form inputs.

**Http check error discovery**

https://github.com/user-attachments/assets/00738b1e-c810-4a9c-8b67-326908ff3b20

**Multihttp check error discovery**

https://github.com/user-attachments/assets/54caab0e-b4bc-47d3-bc15-413bf636a7e2

**Script check error discovery, redirecting back to script tab from examples**

https://github.com/user-attachments/assets/ce9cd96a-6371-4704-b700-ca098b6758bc

**Focussing probes select correctly when it has an error and is the first field**

https://github.com/user-attachments/assets/1d571af8-3c90-478d-87f9-f8713b36b991


src/components/CheckEditor/FormComponents/MultiHttpCheckRequests.tsx
```diff
@@ -70,28 +73,26 @@ export const MultiHttpCheckRequests = () => {
...
-          open: true,
-          index: res.index,
-          tab: res.tab,
+          type: 'openRequestPanels',
+          indexes: res,
         });
 
-        if (panelRefs.current[res.index]) {
-          focusField(panelRefs.current[res.index], res.id);
+        if (source !== `collapsible`) {
```
VikaCep, line 86: Can you expand a bit on this scenario? When does this execute when the `source` equals `collapsible`?

src/components/CheckForm/checkForm.hooks.ts
```diff
@@ -95,3 +102,70 @@ export function useCheckForm({ check, checkType, onTestSuccess }: UseCheckFormPr
     handleInvalid,
   };
 }
+
+function findFieldToFocus(errs: FieldErrors<CheckFormValues>) {
```
VikaCep, line 106: This file is called `checkForm.hooks.ts` but these new functions are not hooks, maybe something to consider when we work on [this](https://github.com/grafana/synthetic-monitoring-app/issues/848) ticket

VikaCep (commented):
👏 👏 👏 👏

src/components/CheckForm/checkForm.hooks.ts
ckbedwell: Good point! I've separated them out to `checkForm.utils.ts`.

src/components/CheckEditor/FormComponents/MultiHttpCheckRequests.tsx
ckbedwell: I've added a comment explaining why it is needed to check the broadcast source (it's to avoid an infinite loop -- if you have two event listeners which rebroadcast the same event they continually trigger each other).

VikaCep (approved):

Issue #864, opened 2024-07-12
Validation of browser and scripted checks


peterschretlen:
With the introduction of browser & scripted checks, we want to enforce that the code matches the type of check that is selected

What happens if you add this to a scripted check?
```
import { browser } from 'k6/browser' // or 'k6/experimental/browser';
```
What happens if you fail to include it in a browser check?

There may be both FE and BE validation needed here, but it would be preferable to validate in the editor rather than when a script is tested or saved.

VikaCep:
The frontend validation has been implemented here https://github.com/grafana/synthetic-monitoring-app/pull/869. Not sure if we also want to add backend validation or we can close the issue.

peterschretlen:
This is coupled to supporting archives

PR #869, opened 2024-07-17
Feat: add browser check validations (frontend)

VikaCep:
Using the `acorn` library, the script from both the browser and scripted checks is parsed to make sure that it contains the `browser` import (from `k6/browser`) for browser checks, and validate that it is not present in scripted checks.
It also validates the script exports the `options` configuration in browser checks. 

Frontend part of https://github.com/grafana/synthetic-monitoring-app/issues/864

### Browser:

**Valid scripts**
image
image

**Not importing browser from `k6/browser`**
image
image
image

**Not exporting options**
image

### Scripted check:

**Valid script:**
image

**Importing browser:** 
image


VikaCep:
@ckbedwell regarding tests, I'd rather add them after porting your [form refactor changes](https://github.com/grafana/synthetic-monitoring-app/pull/847) into this branch as you implemented a test framework that I think will facilitate writing them here. Let me know if you disagree though.

package.json
```diff
@@ -109,6 +109,8 @@
     "@popperjs/core": "^2.9.2",
     "@tanstack/react-query": "5.8.4",
     "@tanstack/react-query-devtools": "5.8.4",
+    "acorn": "^8.12.1",
+    "acorn-walk": "^8.3.3",
```
ckbedwell, line 113: Can we remove the ^ prefix? I don't trust developers with semvar enough 😄

ckbedwell (commented):
Super minor comment but otherwise this is great work! I've tried to break it and haven't managed yet. 👏👏👏

Not for this PR but I did noticed the validation gets triggered like four or five times on a single character press but it doesn't make any discernible UX impact.

Also not for this PR but I know acorn is relatively big so at some point in the future we can look at lazy loading it as it is only needed for a very specific part of the application.

ckbedwell (approved):
🔥🔥🔥

PR #871, opened 2024-07-18
feat: add hover/focus-visible interaction to form sidebar buttons

ckbedwell:
## Problem

Whilst I was writing the updated documentation for the check form redesign, I remembered we had previously had feedback about the form sidebar buttons not being very obvious they are clickable.

### Solution

I've added an underline on hover/focus-visible psuedo states to make it clearer these are buttons which will navigate to each section.

**Hover state**
Screenshot of Grafana Cloud Synthetic Monitoring on the check creation page. The form sidebar second step is being hovered and an underline is visible.

**Focus-visible state**
Screenshot of Grafana Cloud Synthetic Monitoring on the check creation page. The form sidebar second step is being focussed and an underline is visible.


VikaCep (approved):
💯 💯 💯

PR #874, opened 2024-07-19
Restrict access to certain pages/actions when user is Viewer

VikaCep:
Users with the role of `viewer` should be more restricted in what they can do/click in the app. 

For `viewer` users, this PR:
- Disables "Edit check" button in dashboards page
image
- Redirects the edit page route to home
- Redirects the config page to home
- Disables the "Populate default alerts", "Reset default alert rules" and "Save alerts" buttons in Alerting

Fixes https://github.com/grafana/support-escalations/issues/11601

### DIFF
--- a/src/page/CheckRouter.test.tsx
+++ b/src/page/CheckRouter.test.tsx
@@ -9,16 +9,9 @@ import { CheckType, CheckTypeGroup, ROUTES } from 'types';
 import { PLUGIN_URL_PATH } from 'components/Routing.consts';
 
 import { CheckRouter } from './CheckRouter';
+import { runTestAsViewer } from 'test/utils';
 
 describe(`<CheckRouter />`, () => {
-  beforeAll(() => {
-    jest.replaceProperty(config, 'bootData', {
-      // @ts-expect-error
-      user: {
-        orgRole: OrgRole.Editor,
-      },
-    });
-  });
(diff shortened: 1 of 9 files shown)


VikaCep:
I still intend to add few more tests to improve coverage.

src/page/CheckRouter.test.tsx
```diff
@@ -9,6 +11,14 @@ import { PLUGIN_URL_PATH } from 'components/Routing.consts';
...
 
 describe(`<CheckRouter />`, () => {
+  beforeAll(() => {
+    jest.replaceProperty(config, 'bootData', {
+      // @ts-expect-error
+      user: {
+        orgRole: OrgRole.Editor,
+      },
+    });
+  });
```
ckbedwell, line 21: We have a global afterEach set up in `test/jest-setup.tsx` that clears all mocks so you can remove these, too.

This is my bad, I'm going to make a ticket to document how our testing framework works and what assumptions and things it does for you out of the box to make tests clearer and more predictable to write 😅

ckbedwell (changes requested):
This is great, well done. I'm actually working on adding to this right now with RBAC but it will take longer to implement so this is a great solution 💪

ckbedwell (approved):
LGTM 🌟

ckbedwell (approved):
⭐

Issue #877, opened 2024-07-22
Add datasource RBAC in the plugin


ckbedwell:
## Problem

Despite the Synthetic Monitoring api doesn't support RBAC yet, the plugin does receive enough information from Grafana regarding the three datasources it relies on to know if a user has the right permissions to do certain actions. We should implement RBAC in the plugin ASAP to reduce escalations and so that we will be ready to add additional RBAC features when the api supports them.

## Completion criteria

Reference draft PR to help inform these steps: https://github.com/grafana/synthetic-monitoring-app/pull/876

The plugin can be in several different states dependent on what datasource permissions a user has. At the moment it is difficult to reason with what state the plugin can be in as a lot of important information is all added in the `InstanceContext`. The first steps to considering this ticket complete are to split the instance context up into three separate contexts:

```[tasklist]
### Ground work
- [x] `MetaContext` and to create a meta data 'gate' to verifies that the meta data we expect is present instead of having a lot of type casting / optional chaining in consuming components: https://github.com/grafana/synthetic-monitoring-app/pull/878
- [x] `SMDatasourceContext` - if this isn't available or set-up access to the plugin is impossible: https://github.com/grafana/synthetic-monitoring-app/pull/879
- [x] Logs and metrics hooks - there should be a single source of truth for these across the application. https://github.com/grafana/synthetic-monitoring-app/pull/880
```

Related tickets:
https://github.com/grafana/synthetic-monitoring-app/issues/845
https://github.com/grafana/synthetic-monitoring-app/issues/846

```[tasklist]
### Meat and bones
- [ ] Unify where the datasources are read from and ensure they provide what RBAC permissions a user has
- [ ] Create a permissions feature with relevant context / components which support RBAC features
- [ ] Ensure the config pages are useful / correct for each area they can be displayed in dependent on the users' permissions
```

All of the above are addressed in: https://github.com/grafana/synthetic-monitoring-app/pull/883

## Additional context

Please read the RBAC docs in `grafana/plugin-tools`: https://github.com/grafana/plugin-tools/pull/1019

PR #878, opened 2024-07-22
Create meta context

ckbedwell:
## Problem

This is to address the first checklist item of the plugin RBAC epic: https://github.com/grafana/synthetic-monitoring-app/issues/877

## Solution

Creating a meta context allows us to create a better gate so we can write components with the correct assumptions of what data is available.

### DIFF
--- a/src/components/LinkedDatasourceView.tsx
+++ b/src/components/LinkedDatasourceView.tsx
@@ -4,6 +4,7 @@ import { Spinner } from '@grafana/ui';
 import { ROUTES } from 'types';
 import { findLinkedDatasource } from 'utils';
 import { InstanceContext } from 'contexts/InstanceContext';
+import { useMeta } from 'hooks/useMeta';
 import { useNavigation } from 'hooks/useNavigation';
 
 interface Props {
@@ -12,13 +13,14 @@ interface Props {
 
 export const LinkedDatasourceView = ({ type }: Props) => {
   const navigate = useNavigation();
-  const { instance, meta } = useContext(InstanceContext);
+  const { instance } = useContext(InstanceContext);
+  const { jsonData } = useMeta();
   if (!instance.metrics || !instance.logs) {
     return <Spinner />;
   }
 
   const info = type === 'prometheus' ? instance.metrics : instance.logs;
-  const hostedId = type === 'prometheus' ? meta?.jsonData?.metrics.hostedId : meta?.jsonData?.logs.hostedId;
+  const hostedId = type === 'prometheus' ? jsonData?.metrics.hostedId : jsonData?.logs.hostedId;
   const datasource = findLinkedDatasource({
     grafanaName: info.name,
     hostedId: hostedId ?? 0,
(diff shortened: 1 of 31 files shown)


src/contexts/MetaContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+function verifyMeta(meta: AppPluginMeta<GlobalSettings>): VerifiedMeta {
+  // if (!meta.jsonData) {
+  //   console.log(meta);
+  //   throw new Error('verifyMeta: meta.jsonData is required');
+  // }
+
+  // if (!meta.jsonData.stackId) {
+  //   console.log(meta);
+  //   throw new Error('verifyMeta: meta.jsonData.stackId is required');
+  // }
```
VikaCep, line 45: Should this be uncommented?

ckbedwell: I've removed the commented out code and left a todo comment. I want to circle back to this problem in a future PR because there's a bit of a problem with how we share components from the Plugin to the Datasource and what assumptions they expect of data to be available.

The Datasource provides us the meta data for the Datasource plugin and the App plugin provides us the meta data for App Plugin and they aren't 1:1 so we need to have some sort of normalisation for components they want to share and not just assume they are the same. This is especially true when the state of provisioned / initialised gets involved.

To go someway to addressing this I've removed the `<LinkedDatasource />` components we use in the datasource's `<ConfigEditor />` as they expect plugin meta data. I'll add something back at a later date but I want to address the problem of unifying where we read the logs / metrics datasources from because at the moment it is difficult to reason with.

src/components/LinkedDatasourceView.tsx
```diff
@@ -12,13 +13,14 @@ interface Props {
...
-  const { instance, meta } = useContext(InstanceContext);
+  const { instance } = useContext(InstanceContext);
+  const { jsonData } = useMeta();
   if (!instance.metrics || !instance.logs) {
     return <Spinner />;
   }
 
   const info = type === 'prometheus' ? instance.metrics : instance.logs;
-  const hostedId = type === 'prometheus' ? meta?.jsonData?.metrics.hostedId : meta?.jsonData?.logs.hostedId;
+  const hostedId = type === 'prometheus' ? jsonData?.metrics.hostedId : jsonData?.logs.hostedId;
```
VikaCep, line 23: `jsonData` is a bit ambiguous and it's hard to understand what data it refers too. This refers to global settings, right? If so, what do you think of renaming it to something like that?

VikaCep (approved):
Nice work!!! 🏅

src/components/LinkedDatasourceView.tsx
ckbedwell: The thing is that is what it is called in the provisioning file for the plugin so I have some reluctance to rename it. I like that if we leave it as `jsonData` we know that this is what is being provided in `provisioning/plugins/default.yaml` file.

I am actually tempted to go the other route and rename `GlobalSettings` to `PluginJsonData` 😅

PR #880, opened 2024-07-22
feat: create logs/metrics DS hooks

ckbedwell:
## Problem

Currently it is difficult to reason with where the Logs (Loki) and Metrics (Prometheus) datasources attached to Synthetic Monitoring are being read from across the code base.

## Solution

Create `useLogsDS` and `useMetricsDS` hooks and unify across the codebase these are the source of truth when wanting to access each datasource.

This is the third point of the RBAC task list: https://github.com/grafana/synthetic-monitoring-app/issues/877


VikaCep (approved):
Nice! I love how much simpler it is to access ds data now.

PR #885, opened 2024-07-23
Fix tests browser checks

VikaCep:
After the `user.type` method was [changed](https://github.com/grafana/synthetic-monitoring-app/pull/882) to replace content instead of typing it for performance reasons, the recently merged tests in browser checks needed to be adjusted to prevent errors with certain characters in the script text.


ckbedwell (approved):
I love that it means we don't have to use the escape characters any more 🥳 

LGTM

Issue #866, opened 2024-07-12
Browser check dashboard


peterschretlen:
What should the dashboard for a browser check include? Beyond web vitals, is there anything additional it should include beyond what we display for scripted checks?  This will take some iteration.

PR #890, opened 2024-07-26
feat: Browser checks dashboard

VikaCep:

This PR adds a dedicated dashboard page for Browser checks. It displays the same panels as in the Scripted Checks dashboard, in addition to new sections displaying web vitals information. 

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/866

This section copies the same format as the web vitals components used in the Frontend Observability page for consistency between products.

image
image
image


src/scenes/BROWSER/browserScene.ts
```diff
@@ -0,0 +1,121 @@
...
+            direction: 'row',
+            children: [getAssertionTable(logs, checkType, minStep)],
+          }),
+          new SceneFlexLayout({
+            direction: 'row',
+            height: 200,
+            children: [distinctTargets, probeDuration],
+          }),
+          getDataTransferred(metrics),
+          /* @todo: implement a table displaying web vitals metrics per url
```
ckbedwell, line 107: I was just thinking this would be a great addition.

src/scenes/SCRIPTED/AssertionsTable/AssertionsTable.tsx
ckbedwell, line 1: As this is shared can we create a new folder to make it clear. Maybe something like `scenes/k6common/AssertionsTable`?

Naming things is hard 😭

ckbedwell (commented):
This is great! I certainly have opinions on how we can improve this but it's a great start to get out and start working with and I am eager to get this out and progressively enhance it as we go! 👏

src/scenes/SCRIPTED/AssertionsTable/AssertionsTable.tsx
VikaCep: There was already a `Common` folder for shared scenes, so I've moved it there 
image

src/scenes/BROWSER/browserScene.ts
VikaCep: I'm working on this table, but I'm thinking to add it in a following PR as not to delay this first version ⭐️

src/scenes/BROWSER/distinctTargets.ts
```diff
@@ -0,0 +1,34 @@
...
+import { ExplorablePanel } from 'scenes/ExplorablePanel';
+
+function getQueryRunner(metrics: DataSourceRef) {
+  return new SceneQueryRunner({
+    datasource: metrics,
+    queries: [
+      {
+        expr: `count by (job, target) (
+          count by (url, method) (
+            probe_browser_web_vital_fcp{probe=~"\${probe}", job="\${job}", instance="\${instance}"}
```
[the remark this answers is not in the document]
VikaCep: So taking into account the available metrics we have, grouping by the url on a web vitals metric seemed to me like the best option to distinguish targets.

With this implementation, we'll show 2 targets when a browser check visits two distinct urls:
image
image

If the browser check only goes to one, we'll just show one target. But I understand this leaves out information about the different interactions a user might have within a page. 

If we think this is not useful information or might be misleading, there's also the option of removing this panel. What do you think? We could also explore the logs to see if there's any other relevant information that we can obtain regarding targets, instead of just relying on the metrics data for this panel.

Regarding the `label` in the query, I removed it as you correctly pointed out it's not a present in the metric's response.

ckbedwell (approved):
LGTM 🌟

PR #895, opened 2024-08-01
Hackathon: i18n all the things

ckbedwell:
Will fill this out accordingly.


.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   extends: ['./.config/.eslintrc', 'plugin:@tanstack/eslint-plugin-query/recommended'],
-  plugins: ['simple-import-sort'],
```
joey-grafana, line 3: This is the manual piece we need to automate!

.eslintrc.js
```diff
@@ -1,6 +1,6 @@
...
+  plugins: ['simple-import-sort', '@grafana'],
```
joey-grafana, line 3: This is the manual piece we need to automate!

ckbedwell:
Going to close this to reduce open PR noise. We have it as reference if we need it :)

Issue #900, opened 2024-08-14
Add validation to Checks' job name to prevent single quotes and commas


VikaCep:
When checks contain single quotes or commas in their job names, it causes the queries executed by the panels to return no data. We should add a frontend validation in the job name field in order to prevent users to input these characters.

Note:
Regarding commas, there is a problem with how scenes/grafana is parsing values to put in the URL.
Single quotes seems to be a problem with how Mimir handles quote escaping with executing a query.

Ref https://github.com/grafana/synthetic-monitoring-app/issues/610

Issue #676, opened 2024-02-07
Usage of single quote in a check job or instance causes dashboard to show no data


ka3de:
**What happened**:
An check created with a `job` name or `instance` value that contains a single quote `'` makes the dashboard panels to show `No Data` even if the check is actually running and the data is available. Notice that exploring the panel queries seems to work OK, and the queries themselves are shown correctly.

**What you expected to happen**:
The dashboard panels show the check data.

**How to reproduce it (as minimally and precisely as possible)**:
Create an HTTP check with values:
- job: `Test single quotes`
- instance: `https://grafana.com/'test'`
Or:
- job: `Test single 'quotes'`
- instance: `https://grafana.com`

![Screenshot from 2024-02-07 16-25-52](https://github.com/grafana/synthetic-monitoring-app/assets/6529572/d96af0d5-691b-4132-a9fa-4839c39c2ef1)

**Anything else we need to know?**:

**Environment**:
- SM plugin version: 1.12.5
- OS Grafana is installed on:
- User OS & Browser: Linux Ubuntu, Firefox 122.0
- Others:

rdubrock:
See also https://github.com/grafana/synthetic-monitoring-app/issues/610

This is, I think, because of how Grafana itself escapes dashboard variable values, not necessarily connected to synthetics itself

ka3de:
> See also https://github.com/grafana/synthetic-monitoring-app/issues/610
> This is, I think, because of how Grafana itself escapes dashboard variable values, not necessarily connected to synthetics itself

Ah, I did search for issues with `quote` or `'` in the issue name but found nothing, missed `apostrophes`. I guess we can close this one as duplicate an prioritize the other one if there's something we can do.

PR #902, opened 2024-08-14
Prevent commas and quotes in job names

VikaCep:
When checks contain single quotes or commas in their job names, it causes the queries executed by the panels to return no data. We should add a frontend validation in the job name field in order to prevent users to input these characters.

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/900 

image

There have been several escalations related to this problem:
- https://github.com/grafana/support-escalations/issues/11944
- https://github.com/grafana/support-escalations/issues/6235
- https://github.com/grafana/synthetic-monitoring-app/issues/676

### DIFF
--- a/src/schemas/general/Job.ts
+++ b/src/schemas/general/Job.ts
@@ -7,9 +7,9 @@ export const JobSchema = z
   .min(1, { message: 'Job name is required' })
   .max(128, { message: 'Job name must be 128 characters or less' })
   .superRefine((value, ctx) => {
-    if (value.includes("'") || value.includes(',')) {
+    if (value.includes("'") || value.includes(',') || value.includes('"')) {
       ctx.addIssue({
         code: z.ZodIssueCode.custom,
-        message: `Job names can't contain commas or single quotes`,
+        message: `Job names can't contain commas or quotes`,
       });
     }})
--- a/src/schemas/general/Job.ts
+++ b/src/schemas/general/Job.ts
@@ -5,4 +5,11 @@ export const JobSchema = z
     required_error: 'Job name is required',
   })
   .min(1, { message: 'Job name is required' })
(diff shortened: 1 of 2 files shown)


src/schemas/general/Job.ts
```diff
@@ -5,4 +5,11 @@ export const JobSchema = z
     required_error: 'Job name is required',
   })
   .min(1, { message: 'Job name is required' })
-  .max(128, { message: 'Job name must be 128 characters or less' });
+  .max(128, { message: 'Job name must be 128 characters or less' })
+  .superRefine((value, ctx) => {
+    if (value.includes("'") || value.includes(',')) {
```
peterschretlen, line 10: This would not be easy to figure out, but could there be other characters that cause problems besides these two?

VikaCep: I noticed it also fails with double quotes, so added extra validation for it.
image
image

Tested it with other special characters and it seems to work fine, even backticks.
image
image

peterschretlen: We just found another case that breaks: when there is leading whitespace.  A job name of "   ABC" will get converted to a query for "ABC" at somepoint (somewhere along the way, part of the system is stripping the leading whitespace), and return no results. 

So we might want to also disallow leading or trailing whitespace (or strip it out automatically before we create the check).

ckbedwell (changes requested):
Can you just remove the licence stuff from this PR otherwise it is good to go! I looked into it and we can add the license to version control. I split out the license PR to here so we have a nicer history to record it:

https://github.com/grafana/synthetic-monitoring-app/pull/904

peterschretlen (approved):

ckbedwell (approved):
LGTM ⭐

Issue #786, opened 2024-05-21
Multihttp & Scripted check assertions - success rate by probe graph doesn't show failures properly


peterschretlen:
**What happened**:
Assertions in a multihttp and/or scripted check are failing, but the chart does not show when the failures occur. 

Here is an example of an assertion that failed 58 times over the course of a week.  The success chart shows some gaps and some dots, but it never goes below 100%.  This makes it very hard to narrow down when/where the failures occurred.
<img width="1564" alt="image" src="https://github.com/grafana/synthetic-monitoring-app/assets/7903188/e189fa46-207b-441e-a071-25ff0bef8201">

**What you expected to happen**:

If I open explore from the dashboard panel I can see that the interval is every 2m,  for a check that runs on a 2m interval. 
When I change the interval to 5m, I can now see where the failures occured

<img width="1590" alt="image" src="https://github.com/grafana/synthetic-monitoring-app/assets/7903188/e54e30aa-9b10-4aea-9f09-81963e9d1733">

**How to reproduce it (as minimally and precisely as possible)**:

This script will randomly fail about 10% of the time.  Run this as a scripted check, and look at the assertion graph for '2. is result < 10'.  The plot lines will never go below 100% even though it is failing. 
```
import { check, fail } from 'k6'
import http from 'k6/http'

export default function main() {
  const res = http.get('https://www.randomnumberapi.com/api/v1.0/random?min=1&max=10');
  // console.log will be represented as logs in Loki
  console.log(`response ${res.json('[0]')}, ${res.status}`);
  check(res, {
    '1. is status 200': (r) => r.status === 200,
    '2. is result < 9': (r) => parseInt(res.json('[0]')) <  9,
  }) || fail(`check failed`);;
}
```

peterschretlen:
I think an attempt was made to fix it here (the panel used to have a fixed interval of 5m)  https://github.com/grafana/synthetic-monitoring-app/pull/772/files, but the interval still seems to low?

peterschretlen:
This might be partly related to https://github.com/grafana/synthetic-monitoring/issues/55.  Since we have gaps in the logs, we also show gaps in the charts.  Even if the logs are fixed though, I think showing the failures might still be an issue

peterschretlen:
https://github.com/grafana/synthetic-monitoring/issues/55 should be fixed after v0.53 of k6 is released.  That will help this problem somewhat.

VikaCep:
[...]

PR #903, opened 2024-08-15
Fix success rate graph when showing failures

VikaCep:
## Description

The current Success Rate graph on the dashboards for MultiHTTP, Scripted, and Browser checks does not accurately display failure events. The issue arises because the query used for this visualization uses an interval that matches the frequency of the checks. This setup can leads to an incomplete visualization, as failures are nont captured effectively within the same interval.

**Problem**
The interval used in the query is equal to the check frequency, which can cause the graph to miss or inadequately represent failure events.
This results in a visualization that may not accurately reflect the true success rate, particularly in the presence of failures.

**Solution**
To improve the accuracy of the failure representation in the Success Rate graph, this PR modifies the interval to be a multiple of the check frequency. Specifically, the interval is set to twice the check frequency.

See the following visualizations for a check that runs every 2m:

- With an interval of `2m`
image

- With an interval of `4m`
image

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/786

### DIFF
--- a/src/scenes/BROWSER/browserScene.ts
+++ b/src/scenes/BROWSER/browserScene.ts
@@ -96,7 +96,7 @@ export function getBrowserScene(
           }),
           new SceneFlexLayout({
             direction: 'row',
-            children: [getAssertionTable(logs, checkType, minStep)],
+            children: [getAssertionTable(logs, checkType, checks?.[0]?.frequency)],
           }),
           new SceneFlexLayout({
             direction: 'row',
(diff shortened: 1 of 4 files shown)


ckbedwell (changes requested):
This is really an approval but a reminder to get rid of the license code 😂 

Great job! 💪

peterschretlen (approved):
Nice 🎉 , LGTM!

ckbedwell (approved):
LGTM 🎉

PR #918, opened 2024-08-20
Browser Checks Dashboard - Iteration 2

VikaCep:
_Draft PR - WIP_

Part of https://github.com/grafana/synthetic-monitoring-app/issues/866


src/scenes/BROWSER/WebVitals/pageLoad.ts
```diff
@@ -1,24 +1,26 @@
...
-        legendFormat: 'FCP {{url}}',
+        expr: `avg by (instance, job) (quantile_over_time(0.75, probe_browser_web_vital_lcp{instance="$instance", job="$job"}[$__range]))`,
+        legendFormat: 'LCP',
       },
       {
         refId: 'C',
-        expr: `sum by (url) (quantile_over_time(0.75, probe_browser_web_vital_lcp{instance="$instance", job="$job"}[$__rate_interval]))`,
-        legendFormat: 'LCP {{url}}',
+        expr: `avg by (instance, job) (quantile_over_time(0.75, probe_browser_web_vital_ttfb{instance="$instance", job="$job"}[$__range]))`,
+        legendFormat: 'TTFB',
```
ckbedwell, line 23: I have quite literally spent days seeing if there is an alternative way to write this because it could be misleading to what I would expect to see as a user and this query is the same formula for all of the other PromQL queries in this PR. It also might be totally fine and I have been overthinking this so I am going to write out my thinking below as I think it is still a useful exercise to run through.

Let's take some examples and decide what we think would be most useful to show:

## The check
A check utilising **three** probes that runs **every minute** and hits **two** URLs in its duration. Let's say these TTFB results are over the course of an hour:

First 50 checks: 

| | Probe 1    | Probe 2 | Probe 3
| ------- | -------- | ------- |  ------- |
| URL A  |    200ms  | 200ms | 1800ms |
| URL B | 150ms     | 150ms | 250ms |

10 other checks:
| | Probe 1    | Probe 2 | Probe 3
| ------- | -------- | ------- |  ------- |
| URL A  |    1700ms  | 1700ms | 5000ms |
| URL B | 150ms     | 150ms | 250ms |

So, the text summary here is:

URL 1 is generally fast except when serving to probe 3. Probe 3 always seems to be slow so should be investigated. However for 10 minutes of the hour URL A had serious degradation across all probes but URL B was unaffected.

Now, frankly that is a **very** difficult scenario to accurately capture in a single stat panel but I am interested in if there is a way that this behaviour gets masked with the query and asking several questions.

So let's break down the query: `probe_browser_web_vital_ttfb{instance="$instance", job="$job"}[$__range])` this is going to return something like:

```
{url="URL A" probe="Probe 1" rating="good"} -- 200
{url="URL A" probe="Probe 2" rating="good"} -- 200
{url="URL A" probe="Probe 3"  rating="poor"} -- 1800

{url="URL A" probe="Probe 1" rating="needs-improvement"} -- 1700
{url="URL A" probe="Probe 2" rating="needs-improvement"} -- 1700

{url="URL B" probe="Probe 1" rating="good"} -- 150
{url="URL B" probe="Probe 2" rating="good"} -- 150
{url="URL B" probe="Probe 3" rating="good"} -- 250
```

Notice that the results get split because of the `rating` label (there's a further complication that they also get split if the `config_version` differs over the course of the time range which we will ignore for the purpose of this example). What is worth highlighting here is that the 10-minute 5000ms degradation we see for URL A in Probe 3 is **not** reported in these results. It doesn't appear because it shares the same label and because it lies outside of the 75th percentile we don't report it.

Now with each label set we have returned we average them. They all share the same job and instance and every other label is ignored. So we end up with a result of: 769ms

((200 + 200 + 1800 + 1700 + 1700 + 150 + 150 + 250) / 8)

Depending on what we expect to see here as a user, this might not seem such a bad figure to report when we compare it to the text summary above.

Armed with this knowledge we can see a few ways that we could adjust the scenarios and the results might differ _massively_ and not represent accurately what is going on at all. Say we had a check with 22 probes and they all worked fantastically all the time over the course of 10 hours (200ms TTFB) except 1 probe for 1 check blipped and had a 30 second TTFB, I don't think the PromQL query makes a distinction across time in this case so it is essentially going to be reporting a 1554ms TTFB for the whole 10-hour query: ((21 x 200) + 30000 / 22))

So we get reported an almost 800% degradation on our dashboard (200ms -> 1500ms) because 1 check out of 1320 blipped.

---------------------

So in short, I think the issue comes from mixing the collection of p75 metrics and also averaging. Before we  change this query I think fundamentally we need to make some decisions: 

1. Do we commit to trying to report the p75 for everything? If we do, we should agree what that looks like for panels which are trying to represent a complicated scenario
2. Do we mix and match -- so some are p75, some are averaged metrics?

I think in either case we need to be more explicit how we inform the user what the panel is showing. I actually forgot that web vitals use p75 so unanimously as their basis for reporting so that might go some way to help users understand what their dashboards are showing them.

VikaCep: This is a really good analysis! Thanks for laying it out so clearly 👏 

I completely agree with your concern that we might be masking important information in certain scenarios, and it’s indeed challenging to capture these situations in a stat panel or timeseries.

Regarding reporting `p75`, I think it's effective for understanding the usual performance and providing a benchmark for most users. I checked and this is what they're using as well in [FE O11y](https://github.com/grafana/app-o11y-kwl/blob/main/plugin/src/scenes/components/apps/overview/queries.ts#L38-L62) for their web vitals panels and similarly in [k6](https://github.com/grafana/grafana-k6-app/blob/942def741bc022f17e37cd2a8190692829461d3c/src/constants/metrics.ts#L694), but it is true that there are special situations that might make the numbers represent something in a wrong way.

Averaging (generally by instance and job) is the way I found to solve the problem of how to visualize this information in a concise way. Without averaging, we'd end up with a new series for each label difference, which might clutter the visualization and make it less useful.

Maybe we could add additional panels for reporting significant deviations or degradation events in order to provide a more complete view of performance? I imagine something like showing the percentage of total URL responses that take long/longer to respond. 

In any case, i think you're very right in that we should be more explicit in the way we inform what is it we're showing to users.

ckbedwell (approved):
LGTM 🎉🎉🎉 

@VikaCep and I have discussed this PR in detail and have agreed the best course of action is to merge it as and iterate on it in the future to make small and incremental improvements.

The main concern right now is that it is not clear how the summed up web vital panels are calculated so Vika will make a separate PR to add explore links so users can see how the number is calculated.

Issue #920, opened 2024-08-21
Changing the name of the SM DS breaks the plugin


VikaCep:
The frontend is currently assuming the Synthetic Monitoring datasource is named `Synthetic Monitoring` , as it's the default name for it. (https://github.com/grafana/synthetic-monitoring-app/blob/main/src/data/useSMSetup.ts#L18)

However, this can be changed from the datasource configuration which will make the app not find it and always show the "Get Started" screen. 

![image](https://github.com/user-attachments/assets/7a11cf25-9f39-4ba5-b877-eb88a573623a)
![image](https://github.com/user-attachments/assets/4dae4948-3bd2-432b-acc8-e509b03987d8)

To prevent this, the frontend shouldn't make assumptions that the datasource is called `Synthetic Monitoring`. Possibly, it should be obtained by searching by its type (`synthetic-monitoring-datasource`) in the datasources list.

PR #921, opened 2024-08-21
fix: obtain sm datasource by type and not name

VikaCep:
Given that the name of the SM datasource (by default `Synthetic Monitoring`) can be changed in the datasource configuration page and the type is a fixed string (`synthetic-monitoring-datasource`), we should obtain this DS it by that value instead of the name.
 
Otherwise, if the app fails to find it by the default name, it won't be able to obtain it and will stay in an uninitialized status.

Closes https://github.com/grafana/synthetic-monitoring-app/issues/920

**Before:**
image
image

**After:**
image
image

### DIFF
--- a/src/data/useSMSetup.ts
+++ b/src/data/useSMSetup.ts
@@ -12,10 +12,24 @@ export const queryKeys: Record<'list', QueryKey> = {
   list: ['get-sm-datasource'],
 };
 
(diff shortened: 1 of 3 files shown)


ckbedwell (approved):
LGTM! 🚀

PR #923, opened 2024-08-23
fix: revert smwrite permissions to orgroles

ckbedwell:
## Problem

In release 1.14.13 we added RBAC support across the SM plugin. This was intended as a halfway house before we added more pointed and fully-featured plugin RBAC support. Unfortunately we made a bad assumption about the default permissions each role is given.

When creating a Datasource an editor gets `datasource:read` by default and not `datasource:write`. The intent and explanations are documented in the [data source management](https://grafana.com/docs/grafana/latest/administration/data-source-management/) section of the Grafana docs.

## Solution

For the time-being we are going to revert SM permissions to respect organisation roles rather than the access policies they have. This is how it worked prior to `1.14.13`.

### DIFF
--- a/src/hooks/useDSPermission.ts
+++ b/src/hooks/useDSPermission.ts
@@ -1,3 +1,6 @@
+import { OrgRole } from '@grafana/data';
+import { config } from '@grafana/runtime';
+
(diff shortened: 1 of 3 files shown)


VikaCep (approved):
💚 💚 💚

PR #930, opened 2024-09-06
feat: add validation to make sure browser type is chromium

VikaCep:
After a review session with @roobre, she gave me a heads up that the browser script options must be configured to use `chromium`, otherwise it will fail.

In this PR I improve the validation to make sure that it is correctly set.

Part of https://github.com/grafana/synthetic-monitoring-app/issues/864

Examples:

- `type` property missing
image

- `type` property set to firefox
image

- `type` property correctly set
image


roobre:
🎉

PR #931, opened 2024-09-11
Browser checks dashboard - Iteration 3

VikaCep:
Improvements to the browser checks dashboard:

- Adds explore links in web vitals panels so users can see how the number is calculated by getting access to the query in Explore.
image

- Removes the unit in the CLS column of the web vitals by url table
image

Part of https://github.com/grafana/synthetic-monitoring-app/issues/866
Addresses https://github.com/grafana/synthetic-monitoring-app/pull/918#pullrequestreview-2297156876

### DIFF
--- a/src/scenes/BROWSER/WebVitals/WebVitalGauge.tsx
+++ b/src/scenes/BROWSER/WebVitals/WebVitalGauge.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { GrafanaTheme2 } from '@grafana/data';
-import { Icon, LinkButton, Tooltip, useStyles2 } from '@grafana/ui';
+import { useStyles2 } from '@grafana/ui';
 import { css } from '@emotion/css';
 
 import { WebVitalName } from './types';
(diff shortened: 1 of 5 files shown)


ckbedwell (commented):
This makes a big difference so thank you for turning it round so fast ⚡ 

In terms of the UI, my preference would be adapting the web vital gauge so it is more in line with the other panels; so the explore link is in the three-dot hover menu on the panel. I found it a little jarring that the pattern is different to all the other panels on the page.

It shouldn't be too difficult to bring in `<PanelChrome />` from `@grafana/ui` so most of the behaviour is baked in by default in that component but I was looking at `src/scenes/ExplorablePanel` and wondering if this was usable with a custom component rather than having to provide a pluginId?

Lastly, the Explore link should stay in the same tab and not open a new one with `target=_blank`.

VikaCep:
Thanks for your feedback @ckbedwell! I went ahead and used the `PanelChrome` component as you suggested. I took a look at `ExplorablePanel` but it doesn't seem that straightforward to adapt it to support custom scene objects, and `PanelChrome` has all the behavior we need so I opted for that one, but let me know what you think 🙏

ckbedwell (approved):
LGTM 💪

PR #933, opened 2024-09-13
chore: Setup release-please

ka3de:
Sets up release-please GH action.

Updates grafana/synthetic-monitoring/issues/50


.github/release-please/release-please-config.json
```diff
@@ -6,18 +6,8 @@
...
       "changelog-sections": [
         { "type": "feat", "section": "Features" },
-        { "type": "fix", "section": "Bug Fixes" },
-        { "type": "perf", "section": "Performance Improvements" },
-        { "type": "revert", "section": "Reverts" },
-        { "type": "internal", "section": "Internal Changes" },
-        { "type": "deps", "section": "Dependencies" },
-        { "type": "docs", "section": "Documentation" },
-        { "type": "chore", "section": "Miscellaneous Chores" },
-        { "type": "refactor", "section": "Code Refactoring" },
```
VikaCep: We also use `chore` and `refactor`, so we might wanna add them as well.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,49 @@
+name: Release Please
+
+on:
+  push:
+    branches:
+      - main
```
VikaCep, line 6: So, merging to main will always trigger the release, is that right?

VikaCep (commented):
This is amazing! It will simplify a lot of manual tasks, so thanks a lot for setting it up!  🚀🚀🚀
 The only thing I think is missing is some docs describing how the process would work (in particular, for the frontend app) from now on after this PR (and https://github.com/grafana/synthetic-monitoring-app/pull/947) are merged. Apologies if it's already created and I didn't see it.

.github/workflows/release-please.yml
ka3de: Ah apologies, I should have linked here the actual process definition, you can see it in https://github.com/grafana/deployment_tools/pull/177384, this is the part specifically for the app [here](https://github.com/grafana/deployment_tools/blob/sm-release-process-docs/docs/synthetic-monitoring/release-process.md#synthetic-monitoring-app).

.github/release-please/release-please-config.json
ka3de: Ah I missed this, I thought these were not being included in the changelog/release notes. Since you are already using these in your workflow and are included in the release I think it makes sense to add them in this case. See [d2c69a1](https://github.com/grafana/synthetic-monitoring-app/pull/933/commits/d2c69a16b37fc95eb69571e881e087744f54a2a4).

PR #937, opened 2024-09-18
chore: upgrade k6 types version to 0.53.0

VikaCep:
The agent is using `0.53.0` so let's update the k6 types version used in the frontend to match that version.


ckbedwell (approved):
LGTM! (ignore my previous message, I misread the docs and had added the viewport to the wrong object originally).

PR #942, opened 2024-09-25
Feat: extra validations for browser checks

VikaCep:
To prevent execution errors, we're adding some extra validations, in particular, these properties are not allowed in the `options` object:
- `duration`
- `vus` > 1
- `iterations` > 1

image
image
image

### DIFF
--- a/src/schemas/forms/script/parser.ts
+++ b/src/schemas/forms/script/parser.ts
@@ -32,7 +32,7 @@ function getPropertyValueByPath(obj: ObjectExpression, path: string[]): any {
     } else if (property?.value.type === 'Literal') {
       current = property.value.value;
     } else {
-      return undefined;
+      return property?.value;
     }
   }
   return current;
--- a/src/schemas/forms/script/parser.ts
+++ b/src/schemas/forms/script/parser.ts
@@ -61,11 +61,21 @@ const optionExportMatcher: SimpleVisitors<{ options: ObjectExpression | null }> 
 };
 
 export function checkForChromium(objectExpression: ObjectExpression): boolean {
-  // Path to the property we're interested in
-  const path = ['scenarios', 'ui', 'options', 'browser', 'type'];
-  const value = getPropertyValueByPath(objectExpression, path);
+  const scenarios = getPropertyValueByPath(objectExpression, ['scenarios']);
+  if (!scenarios || scenarios.type !== 'ObjectExpression') {
+    return false;
+  }
 
-  return value === 'chromium';
+  for (const scenario of scenarios.properties) {
+    if (scenario.key.type === 'Identifier') {
+      const type = getPropertyValueByPath(scenario.value, ['options', 'browser', 'type']);
+      if (type === 'chromium') {
+        return true;
+      }
+    }
+  }
+
+  return false;
 }
 
 interface ImportBrowserState {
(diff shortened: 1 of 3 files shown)


src/schemas/forms/script/parser.ts
```diff
@@ -61,11 +61,44 @@ const optionExportMatcher: SimpleVisitors<{ options: ObjectExpression | null }>
...
+  return false;
+}
+
+export function hasInvalidProperties(objectExpression: ObjectExpression): boolean {
+  const scenarios = getPropertyValueByPath(objectExpression, ['scenarios']);
+  if (!scenarios || scenarios.type !== 'ObjectExpression') {
+    return false;
+  }
+
+  const invalidProps = ['duration', 'vus', 'iterations'];
```
roobre, line 87: If possible it might be better to be more granular with these. More specifically:

- `duration` must not be set, ever.
- `vus` must be either unset or `<= 1`
- `iterations` must be either unset or `<= 1`

roobre: Specifically, what I'd prefer to avoid is rejecting a script with `{vus: 1, iterations: 1}`, as those are perfectly valid.

VikaCep: Sure, let's tackle the validations in a more specific way.

w1kman (commented):
When entering values as one "should", validation works.
But it's also very easy to bypass the validation.

image
image
(Note: when valid values are used in the same way, the test passes)

IMO, duration, vus and iterations should be stripped before running the script (if they aren't already).
If anything, we should notify the users that the mentioned properties doesn't have any effect 🤷🏻

ckbedwell:
> But it's also very easy to bypass the validation.

@VikaCep this comes back to what we discussing yesterday and could explore further if we could modify the k6 types that we pass to Monaco.

VikaCep:
I enforced the validation to ensure that `iterations` and `vus` can only be either `undefined` or `1`, addressing the potential bypass issue identified by Thomas.

> Specifically, what I'd prefer to avoid is rejecting a script with {vus: 1, iterations: 1}, as those are perfectly valid.

We don't want to reject scripts with configurations like the one above, as Nadia pointed out. That's the reason I don't mark it as invalid when these properties are defined.

> IMO, duration, vus and iterations should be stripped before running the script (if they aren't already).
If anything, we should notify the users that the mentioned properties doesn't have any effect 🤷🏻

I think that’s a valid point, but it might confuse some users who may not understand why those configurations aren’t taking effect, especially when copy/pasting scripts between k6 and SM. However, with clear messaging, we could mitigate that confusion. Personally, I believe marking it as a validation error is a good way to teach users that those values are not valid for a synthetic monitoring check and should not be wrongly set.

> @VikaCep this comes back to what we discussing yesterday and could explore further if we could modify the k6 types that we pass to Monaco.

I agree—this is a great idea since it’s easy to expand and add new rules. We should investigate this further; I’m not sure how we can implement it yet, but it’s definitely worth exploring in a follow-up PR.

peterschretlen:
👍 this seems like a good 80/20 solution for the time being.  Does seem like there's followup work we can do to tighten loopholes on both the frontend and backend.

w1kman (approved):
LGTM! 👍🏻

Issue #914, opened 2024-08-20
Disable check creation for free tier users that have exceeded their limits


ckbedwell:
## Problem

Currently the free tier for Grafana Cloud is 100,000 executions per month. If a user exceeds this limit they aren't informed in the UI until they go through the process of creating / updating a check and receiving a 'Save failed' error message:

![Screenshot of Grafana Cloud Synthetic Monitoring check form on the submission section of check creation. It shows a save failed error with a check executions quota description.](https://github.com/user-attachments/assets/cfe4ae6e-c1a6-4d70-a71d-58d5b74183d5)

## Completion criteria

Do a manual lookup up of how many executions all of a tenant's checks create and if it exceeds 100,000:

- [ ] Disable all check creation avenues across the plugin
- [ ] Display a banner explaining why it is disabled. We show similar messaging / UI for tenants that have exceeded their scripted checks quota so please use this as reference
- [ ] Integration tests for all scenarios presented above

### Considerations

Please ensure your solution will account for when browser checks are GA. It is unconfirmed yet but it is likely that 1 browser check will be the equivalent to 10 test executions.

The manual calculation is somewhat naive and may yield incorrect results. e.g. we count an execution per minute of a check's execution time. If a user's tests are long-running they may have used more of their quota each month then our manual calculation results.

This is a known limitation which is **not** expected to be considered for this ticket to be considered completed. However it may be worth investigating how the `cloud-home-app` queries the Synthetics usage as that would be more accurate than our own calculations:

![Screenshot of Grafana Cloud cloud home app dashboard with usage metrics being displayed.](https://github.com/user-attachments/assets/8abd8b37-4ea3-44a6-a94b-4455a7235b01)

PR #944, opened 2024-09-26
[feat] Disable check creation and add alert for Hosted Grafana Free Tier members who have exceeded their monthly test execution limit

ckbedwell:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/914

## Problem

For Free Tier customers In Hosted Grafana (Grafana Cloud) they have 100,000 free executions per month. We have backend validation to ensure that users can't submit checks that would go past their limit but that is only discoverable after filling in a new check and trying to submit it.

## Solution

~~We use the gcom api (provided by the `hg-cloud-home-admin` app -- more on that below) so we can look up the user's org and then check the org's subscription plan to see if it is free or not.~~

**Update:** As per @VikaCep's suggestion I've decided to use @grafana/runtime's `config.buildInfo.edition` instead of relying on the hg-cloud-home-admin routes. I looked into the way it was populated in core Grafana and it seems reliable. I've also do some testing and it works as I expect.

~~The choose a check type page cards whilst waiting for a response from the gcom api. If the gcom api can't be reached it fails silently and defaults to allowing check creation.~~

**Update:** This logic still exists but while it is fetching all checks to total executions.

The choose a check type page when the user has exceeded their monthly execution limits. It shows an alerting banner outlining the user's options and disables the cards.

image

The create a check page when the user has exceeded their monthly execution limits. It shows the same alerting banner as above and disables the check creation form.

### Total execution calculation

The calculation we use to work out how many executions a check makes over the course of the month differs ever-so-slightly to the way the [SM api does the calculation](https://github.com/grafana/synthetic-monitoring-api/blob/1a6b6e0af252aa838ce02df801f5b36295e1a9c9/internal/dto/limits/calculations.go#L7). There is an internal [Slack thread discussing this](https://raintank-corp.slack.com/archives/C0175SS6SA3/p1714059415879179) which resulted in agreeing the FE over-calculates to ensure users were never caught out. I think this solution is okay for now but I think it is worth exploring in the future how we can ensure BE and FE align perfectly on a user's monthly execution count.

### hg-cloud-home-admin dependency

~~Currently the way we are working out whether a user has exceeded their execution limit is by utilizing the `hg-cloud-home-admin` app's gcom api routes (we are able to do this because their plugin is preloaded and they have already provisioned the necessary information to access the api) which is a brittle dependency and something we would like to remove in the near future. An issue for this has been made: https://github.com/grafana/synthetic-monitoring-app/issues/946~~

~~The other problem this creates is it makes local testing very difficult. Setting up the `hg-cloud-home-admin` app is very difficult as it has a **cascade** (to say the least...) of dependencies outside of itself to get running locally. I have opted to make this as simple as possible and I have added a check in the `useGcom` hook that if it is in a development environment (which should only be true for local builds) it will not look for the hg-cloud-home-admin routes but it will use our proxied response instead.~~

~~The default response is the Pro plan response (`ORG_RESPONSE_PRO`) but to test this feature and enable it in the future the response has to be changed to utilize the `ORG_RESPONSE_FREE`.~~

~~I have ensured this solution is non-blocking for the user if for any reason the gcom api routes result in a failure and have added tests appropriately to reflect this. As mentioned in the linked issue, the `hg-cloud-home-admin` app is free to update their code without worrying about who might be piggy-packing off their routes so the gcom dependency we have may break at any time for us and we want to ensure it has no impact on customers other than a slightly degraded experience.~~

**Update:** This is no longer relevant as this PR is using @grafana/runtime's `config.buildInfo.edition` to work out if the user is free or not. All of this code is saved in the commits in this branch so if we do need to consult the gcom api using hg-cloud-home admin's routes we can easily bring it back in.


__mocks__/@grafana/ui.tsx
```diff
@@ -47,6 +39,5 @@ module.exports = {
   ...ui,
   Icon,
   BigValue,
-  LinkButton,
```
ckbedwell, line 50: Removing this as I wanted to assert on the `aria-disabled` attribute of LinkButton to ensure if the gcom api was unavailable that checks could still be created.

It resulted in a few non-related tests in this PR having to be modified which I think is entirely acceptable.

src/hooks/useAtHgExecutionLimit.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { useCurrentHGSubscription } from './useCurrentHGSubscription';
+import { useMonthlyTotalExecutionCount } from './useMonthlyTotalExecutionCount';
+
+export const FREE_EXECUTION_LIMIT = 100000;
+
+export function useAtHgExecutionLimit() {
+  const { data: totalMonthlyChecks, isLoading: isLoadingMonthlyChecks } = useMonthlyTotalExecutionCount();
+  const { data, isLoading: isLoadingHGSubscription } = useCurrentHGSubscription();
+  const isFree = data === SubscriptionCodeType.FREE;
```
VikaCep, line 11: In the `grafanaBootstrapData` object, we have the `grafanaBootData.settings.buildInfo.edition` which for free users is `Cloud Free`. Have you explored reading this value from there instead calling the gcom api? It might not be enough, but worth to ask. 
image

VikaCep, line 11:
```suggestion
  const isFree = data === SubscriptionCodeType.FREE;
```

Shouldn't we also consider `SubscriptionCodeType.FREE_TRIAL`?

VikaCep: There's also the `licenseInfo` object:
image

majavojnoska:
Hey Vika and Chris. I apologize for responding a bit late. I believe the banner message and actions could be simplified. 

Given that upgrading is the primary goal, I'll consider removing the 'Back to Checks' button as the main call to action. Additionally, deleting checks, reducing probes, and lowering the frequency could be grouped as a more involved option. Finally, I'm reconsidering using an orange warning color instead of an error color, as it seems the user or system may not be technically affected. 

Here is the UI proposal in [Figma](https://www.figma.com/design/WXbmUSpCMlJHG8fpjvmGT1/%F0%9F%8F%97%EF%B8%8F-Synthetics-%5BDEV-READY%5D?node-id=2236-7546&t=8nh57uKjavxaIKWc-4)
Plan limit reached v1
Plan limit reached v2

jmarbach:
Suggested messaging edit in @majavojnoska's rewrite: 

**From**
"Optimize existing checks by deleting unnecessary ones, reducing probes, and lowering the execution frequency."

**To**
"Optimize usage by eliminating unnecessary checks, reducing probes, and decreasing execution frequency."

Feel free to take it or leave it.

src/hooks/useAtHgExecutionLimit.ts
ckbedwell: Good spot! I've gone with this approach in the end. I thought long and hard about it and decided this was the right way to go after researching how the values get populated in core grafana. We still might want access to the gcom api directly at a later date and if that is the case we can bring back in the relevant code. I didn't see the need to leave it and complicate things as our memories get hazy and we forget why it is there 😅

ckbedwell: With the updated approach I checked and Free Trial users have the same edition string as Free Cloud users so this doesn't need to be taken into account now.

ckbedwell:
@VikaCep I've updated the branch by cutting out the `hg-cloud-home` gcom route dependency and I've updated all the limit warning messages as per @majavojnoska's designs. I've ever so slightly changed the copy as I thought it could be misinterpreted from John's new suggestion.

I also noticed we were missing a warning for being at the Browser check limit so added that in, too.

**All the updated limit messages**

Check limit reached.


Scripted check limit reached.


Browser check limit reached.

src/components/OverLimitAlert/AlertContainer.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <Alert severity="warning" title={title} className={styles.container}>
+      {children}
+    </Alert>
+  );
+};
+
+const getStyles = () => ({
+  container: css({
+    maxWidth: `1648px`,
```
VikaCep, line 17: Nitpick, but Is there any reason for this particular value? Shouldn't we use a theme value like `theme.breakpoints.values.xxl`?

VikaCep (approved):
👏 👏 👏 

Nice work! Let's just not forget about this PR in case we need to go back to the previous implementation in the future 🥇

src/components/OverLimitAlert/AlertContainer.tsx
ckbedwell: Yeah, but it's not a very good one 😅 It happens to be the exact width of 4 check group cards and their spacing.

This was a lazy stop gap whilst I was waiting for the other PR you have for redesigning the check group cards.

src/components/CheckForm/CheckForm.tsx
```diff
@@ -81,9 +81,13 @@ export const CheckForm = ({ check, disabled, pageTitle }: CheckFormProps) => {
...
   const status = useCheckTypeOptions().find((option) => option.value === checkType)?.status;
   const isExistingCheck = Boolean(check);
-  const { isLoading, isOverCheckLimit, isOverScriptedLimit, isReady } = useLimits();
+  const { isLoading, isOverBrowserLimit, isOverHgExecutionLimit, isOverCheckLimit, isOverScriptedLimit, isReady } =
+    useLimits();
   const overLimit =
-    isOverCheckLimit || ([CheckType.MULTI_HTTP, CheckType.Scripted].includes(checkType) && isOverScriptedLimit);
+    isOverHgExecutionLimit ||
+    isOverCheckLimit ||
+    (checkType === CheckType.Browser && isOverBrowserLimit) ||
```
w1kman, line 89: Feels like the `useLimit` hook could return the `overlimit` value, which would make the code more readable and for the `overlimit` to have a single source of truth (also DRY).

`const { ..., isOverlimit } = useLimits(checkType)` 🤷🏻

w1kman (changes requested):
Looks good 👌🏻 

Would be nice if the `overlimit` value could be returned directly from `useLimit` (with `checkType` as param).
It's not a show stopper, so feel free to ignore that CR and just re-request a review.

src/components/CheckForm/CheckForm.tsx
ckbedwell: I explored this as I really liked this idea but I realised I struggled with it more than I'd like due to needing to know the reason in many cases why the limit had been reached to show the answer (is it the browser limit, the total check limit, execution count, etc.) so I am going to leave it as it is and we can circle back to improve this when we can.

src/test/handlers/types.ts
```diff
@@ -2,12 +2,14 @@ import { DefaultBodyType, PathParams, RestRequest } from 'msw';
 
 export type RequestRes = RestRequest<never, PathParams<string>> | RestRequest<DefaultBodyType, PathParams<string>>;
 
-type SuccessStatusCodes = 200 | 201 | 202 | 203 | 204 | 205 | 206 | 207 | 208 | 226;
+export type SuccessStatusCodes = 200 | 201 | 202 | 203 | 204 | 205 | 206 | 207 | 208 | 226;
 // prettier-ignore
-type ClientError = 400 | 401 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 421 | 422 | 423 | 424 | 425 | 426 | 428 | 429 | 431 | 451;
-type ServerError = 500 | 501 | 502 | 503 | 504 | 505 | 506 | 507 | 508 | 510 | 511;
+export type ClientError = 400 | 401 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 421 | 422 | 423 | 424 | 425 | 426 | 428 | 429 | 431 | 451;
+export type ServerError = 500 | 501 | 502 | 503 | 504 | 505 | 506 | 507 | 508 | 510 | 511;
```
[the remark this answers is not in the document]
ckbedwell: You are right that the Frontend will never send a status code so it's effectively irrelevant.

These are defined (maybe unnecessarily at the moment) to help when mocking the responses for the tests. Their only purpose is to check that if you've added a 2xx TypeScript can ensure you have remembered to add the correctly typed payload response. I exported them in this PR initially to type the gcom 500 responses for those tests but after I removed them I didn't need to use them in any other files but left this in anyway 😅

PR #949, opened 2024-09-30
chore: update fillform example to use async check

VikaCep:
The browser squad now recommends using the async `check` polyfill in all code examples and has updated the `fillform` example in https://github.com/grafana/xk6-browser/tree/main/examples

In order to be consistent, the `fillform` example for synthetics browser checks has been updated as well.

### DIFF
--- a/src/components/BrowserExamplesMenu/snippets/fillform.js
+++ b/src/components/BrowserExamplesMenu/snippets/fillform.js
@@ -36,8 +36,8 @@ export default async function () {
     await Promise.all([page.waitForNavigation(), page.locator('input[type="submit"]').click()]);
 
     await check(page.locator('h2'), {
-      header: async (lo) => {
-        return (await lo.textContent()) == 'Welcome, admin!';
+      header: async (element) => {
+        return (await element.textContent()) == 'Welcome, admin!';
       },
     });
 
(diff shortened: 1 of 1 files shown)


src/components/BrowserExamplesMenu/snippets/fillform.js
```diff
@@ -35,15 +35,17 @@ export default async function () {
     // to resolve.
     await Promise.all([page.waitForNavigation(), page.locator('input[type="submit"]').click()]);
 
-    const h2 = page.locator('h2');
-    const headerOK = (await h2.textContent()) == 'Welcome, admin!';
-    check(headerOK, { header: headerOK });
+    await check(page.locator('h2'), {
+      header: async (lo) => {
```
w1kman, line 39: I would skip abbreviations in example code (with some exceptions).
I assume `lo` is an abbreviation for "locator" (which resolves into an element-ish).
I suggest that we change `lo` into `element`.

w1kman (changes requested):
I don't think `lo` is a suitable variable name in the context of being an example.

ckbedwell (commented):
Do we know if the polyfill `check` registers for our assertions table in the dashboard correctly? I copied this new example and have had it running for an hour.

The zero state for the assertions panel in a browser dashboard scene.

src/components/BrowserExamplesMenu/snippets/fillform.js
VikaCep: Good catch! changed it 👍

VikaCep:
> Do we know if the polyfill check registers for our assertions table in the dashboard correctly? I copied this new example and have had it running for an hour.

This is odd, I'm seeing the assertions correctly for this script

image

Are they still not showing up?

w1kman (approved):
LGTM 🥇

PR #951, opened 2024-10-02
Feat: improve delete private probe flow

w1kman:
This PR aims to improve the management of probes, including better error handling (specifically when deleting probes). Probe listing now uses more of the components shipped by `@grafana/ui`, which (hopefully) will allow for the view to be more in line with Grafana.

UI has been updated in collaboration with the design dep. It's intended to be good enough, until a potential UI overhaul. 

**Probe Listing**
image

Updates:
- Removed badges
- Added state (online/offline) indicator with tooltip
- Labels now have their actual name listed
- Improves key/value UI for labels
- Added unsupported check types in Probe listing (browser/scripted)
- Possibility to mark, copy & paste values from the probe card
- Moved the onClick (edit) to only be attached to the probe name and the Edit/View button (no longer hidden when not hovering)
- Remove "public/private" from probe listing as probes are group by "public/private" already
- Region is moved into the card title
- Added usage link for when a probe is used in one or several checks (link filters checks by probe)
- Added icons to View/Edit buttons

**Probe delete button**
image
Updates:
- Disabled if a probe is in use
- Tooltip when disabled (with link to checks that are using the probe)

**Confirm modal (async)**
image
image
image

Updates:
- Added new component `ConfirmModal`
- When prop `async` is not present or `false`, the `ConfirmModal` from `@grafana/ui` is returned
- With `async`
  - An enhanced version of `@grafana/ui` is returned instead.
  - `onConfirm` callback is `.catch`ed by the modal
  - If `onError` is passed, the error message can be handled (and passed as prob) from the outside
  - If `onError` is not passed, the component tries to use `.name` and `.message` from the caught error (with generic fallbacks). 

**Additional fixes and refactoring**
- `DataSource.fetchAPI` was added to keep the code bit more DRY
- Removed some redundant variables

Fixes https://github.com/grafana/synthetic-monitoring/issues/97

To trigger error handling for the `ConfirmDialog`, change line 48 in `DeleteProbeButton` from
`  if (!canDelete) {` to `  if (!!canDelete) {` (which will allow you to press the delete button for probes that are in use). You can also use `tweak` if you want to go all-in on testing different types of errors/responses


w1kman (commented):
self-review 👍🏻

VikaCep:
Maybe we could rewrite this error on the frontend to make it more user friendly (capitalize the title and change the id to to the probe name or even add the link to the checks that use it)

image

VikaCep (commented):
Really nice improvement! Love how it looks now 💯 💯 💯 
I left a few comments/questions.

w1kman:
> Maybe we could rewrite this error on the frontend to make it more user friendly (capitalize the title and change the id to to the probe name or even add the link to the checks that use it)
> 
> image

While I agree that the error message could look better, I wouldn't focus on this particular error message, as it results from modifying code. Ideally, this error shouldn't be shown to a user of the app (unless someone starts using a probe while another one tries to delete it).

We could definitely upper case first letter of the `error.name` and we could match and replace numeric portions of the `error.message`. 👍🏻

src/datasource/DataSource.ts
```diff
@@ -185,105 +194,56 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
...
   async addProbe(probe: Probe) {
-    return firstValueFrom(
-      getBackendSrv().fetch<AddProbeResult>({
-        method: 'POST',
-        url: `${this.instanceSettings.url}/sm/probe/add`,
-        data: probe,
-      })
-    ).then((res) => {
-      return res.data;
+    return this.fetchAPI(`${this.instanceSettings.url}/sm/probe/add`, {
```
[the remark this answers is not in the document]
w1kman: If the type is missing, it's because it was also missing previously.
I will take a look at how the usage trickles down, and update with types.

src/datasource/DataSource.ts
```diff
@@ -295,144 +255,76 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
...
   }
 
   async getTenant(): Promise<any> {
-    return getBackendSrv()
-      .fetch({ method: 'GET', url: `${this.instanceSettings.url}/sm/tenant` })
-      .toPromise()
-      .then((res: any) => {
-        return res.data;
-      });
+    return this.fetchAPI<any>(`${this.instanceSettings.url}/sm/tenant`);
```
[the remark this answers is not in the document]
w1kman: I'll try to figure it out 👌🏻

w1kman:
@VikaCep how about this?

- The `ConfirmModal` will uppercase first char
- The error message (the id) is handled by the `DeleteProbeButton` error handler.

image

VikaCep (approved):
👏👏👏

src/components/DeleteProbeButton/DeleteProbeButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+    if (!error) {
+      setError(undefined);
+    }
+
+    setError(getPrettyError(error, probe));
+    // if ('data' in error && 'err' in error.data && 'msg' in error.data) {
+    //   setError({ name: error.data.err, message: error.data.msg });
+    // } else {
+    //   setError({ name: 'Unknown error', message: 'An unknown error occurred' });
+    // }
```
ckbedwell, line 36: ✂️ ?

src/components/DeleteProbeButton/DeleteProbeButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+    ) : (
+      <>
+        You do not have sufficient permissions
+        <br />
+        to delete the probe <span className={styles.probeName}>&apos;{probe.name}&apos;</span>.
+      </>
+    );
+
+    return (
+      <Tooltip content={tooltipContent} interactive={canEdit && !canDelete}>
```
ckbedwell, line 59: You can add the `tooltip` prop directly to button now and it works better for accessibility as it becomes keyboard accessible and reads out the tooltip despite the button being 'disabled'.

With that said, I don't love the link in the tooltip as Grafana doesn't support accessing interactive content via the keyboard in tooltips but as there is an alternative presentation of the link in the sidebar this is okay 😄

src/components/ProbeCard/ProbeCard.test.tsx
```diff
@@ -1,67 +1,140 @@
...
 it('handles probe click', async () => {
-  const probe = PRIVATE_PROBE;
+  const probe = probeToExtendedProbe(PUBLIC_PROBE);
   const { history, user } = render(<ProbeCard probe={probe} />);
   await screen.findByText(probe.name);
   await user.click(screen.getByText(probe.name));
   expect(history.location.pathname).toBe(`${getRoute(ROUTES.EditProbe)}/${probe.id}`);
 });
+
+it.each<[ExtendedProbe, string]>([
```
ckbedwell, line 112: TIL. I always manually looped over when I wanted to do something like this 😅

src/components/ProbeCard/ProbeCard.tsx
```diff
@@ -1,59 +1,92 @@
...
-  const href = `${getRoute(ROUTES.EditProbe)}/${probe.id}`;
+  const probeEditHref = `${getRoute(ROUTES.EditProbe)}/${probe.id}`;
   const labelsString = labelsToString(probe.labels);
+  const styles2 = useStyles2(getStyles2);
 
   return (
-    <Card className={styles.card} href={href}>
-      <div className={styles.cardContent}>
+    <Card>
+      <Card.Heading>
```
ckbedwell, line 24: I am torn on moving back to using @grafana/ui's `Card` component. IMO they have painted themselves into a corner with how prescriptive it is and it is doing way more work than what I would expect a Design System's card to provide.

I think you've done a good job with what they provide here but I do hate that the heading is a h2 regardless as I'd flag this as an accessibility issue on this page when doing an audit. We could probably add props to the `Card.Heading` upstream in grafana/grafana to fix this at a later date so can leave it but just want to flag why I made a local `Card` component in the first place.

src/components/ProbeCard/ProbeCard.tsx
```diff
@@ -1,59 +1,92 @@
...
-  const onlineTxt = probe.online ? 'Online' : 'Offline';
-  const onlineIcon = probe.online ? 'heart' : 'heart-break';
-  const color = probe.online ? 'green' : 'red';
-  const probeTypeText = probe.public ? 'Public' : 'Private';
-  const probeTypeIcon = probe.public ? 'cloud' : 'lock';
-  const styles = useStyles2(getStyles);
-  const href = `${getRoute(ROUTES.EditProbe)}/${probe.id}`;
+  const probeEditHref = `${getRoute(ROUTES.EditProbe)}/${probe.id}`;
   const labelsString = labelsToString(probe.labels);
+  const styles2 = useStyles2(getStyles2);
```
ckbedwell, line 20: Another nitty nit nit: `styles`?

src/components/ProbeCard/ProbeCard.tsx
```diff
@@ -62,46 +95,14 @@ const getStyles = (theme: GrafanaTheme2) => {
...
     }),
-    badges: css({
-      display: `flex`,
-      gap: theme.spacing(0.5),
-    }),
-    meta: css({
-      color: theme.colors.text.secondary,
+    secondaryActionsExtra: css({
+      // justifyContent: 'flex-end',
     }),
```
ckbedwell, line 106: ✂️ ?

src/components/ProbeCard/ProbeCard.tsx
```diff
@@ -121,32 +122,22 @@ const getStyles = (theme: GrafanaTheme2) => {
         marginTop: theme.spacing(1),
       },
     }),
-    link: css({
-      marginBottom: theme.spacing(1),
+    extendedTags: css({
+      gridRowEnd: 'span 2',
     }),
```
ckbedwell, line 127: ✂️ ?

src/components/ProbeCard/ProbeCard.tsx
```diff
@@ -62,46 +95,14 @@ const getStyles = (theme: GrafanaTheme2) => {
...
-    info: css({
-      display: `flex`,
+    badgeContainer: css({
+      display: `inline-flex`,
       gap: theme.spacing(0.5),
-      flexDirection: `column`,
-      gridArea: `info`,
+      flexWrap: `wrap`,
     }),
```
ckbedwell, line 103: ✂️

src/components/ProbeCard/ProbeCard.tsx
```diff
@@ -62,46 +95,14 @@ const getStyles = (theme: GrafanaTheme2) => {
...
-      },
-
-      [mediaQuery]: {
-        gridTemplateAreas: `
-        "info action"
-        "gauge action"
-        `,
-        gridTemplateColumns: `1fr auto`,
-      },
     }),
```
ckbedwell, line 98: ✂️

src/components/ProbeUsageLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const checksCount = hasChecks && probe.checks.length > 0 ? probe.checks.length : 0;
+  const checksHref = `${getRoute(ROUTES.Checks)}?probes=${probe.name}`;
+  const noun = hasChecks && checksCount > 1 ? 'checks' : 'check';
+
+  if (!hasChecks && !showWhenUnused) {
+    return null;
+  }
+
+  return (
+    <TextLink data-testid="probe-usage-link" className={className} variant={variant} color="link" href={checksHref}>
```
ckbedwell, line 27: Nit: There is a file (`src/test/dataTestIds`) to unify the data test ids strings to ensure they stay consistent (and an easy reference for seeing what is available on a page when writing tests) so you could move this testid to there.

src/components/DeleteProbeButton/DeleteProbeButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { useCanEditProbe } from 'hooks/useCanEditProbe';
+import { ConfirmModal } from 'components/ConfirmModal';
+import { ProbeUsageLink } from 'components/ProbeUsageLink';
+
+import { getPrettyError } from './DeleteProbeButton.utils';
+
+export function DeleteProbeButton({ probe }: DeleteProbeButtonProps) {
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
+  const onDeleteSuccess = useCallback(() => {
+    setShowDeleteModal(false);
```
ckbedwell, line 17: Can we handle the redirect back to the Probes list here explicitly?

I don't like it gets handled by the `useEffect` in `EditProbe` as there is a render in-between which threw me off somewhat.

ckbedwell (changes requested):
Mostly nitty nits. A few other comments but on a whole this is ace 💪

ckbedwell:
Last comment but I don't think this is for this PR but future enhancements for the probes UI. 

I was experimenting with what happens when I:
- go to edit a probe that no longer exists: I get redirected -- I think it'd be better to explicitly say the probe doesn't exist as it is less confusing that way.
- and what happens if the probe list request fails --  it doesn't mention it errored but just says there are no private or public probes available.

Some things for us to do in the future 😃

src/components/DeleteProbeButton/DeleteProbeButton.tsx
w1kman: I've added a `onDeleteSuccess` (optional) prop to the `DeleteProbeButton`.
It will now navigate on delete (the `useEffect` has not been removed as still has a purpose).

ckbedwell (approved):
LGTM 🔥🔥🔥

PR #952, opened 2024-10-03
chore: change browser checks to public preview status

VikaCep:
Changes browser checks to public preview status

image
image

We're also removing the public preview badge for scripted checks as they go GA.

image


peterschretlen (commented):
We might as well remove the public preview badge from Scripted checks at the same time?

peterschretlen (approved):

PR #970, opened 2024-10-11
Fix: Update asc / desc reachability sorting

ckbedwell:
## Problem

We had a [bug reported internally in Slack](https://raintank-corp.slack.com/archives/C0162C1K9E1/p1727773255480849) that our reachability Asc / Desc. were reporting the wrong way round.

## Solution

This PR switches the order of Asc / Desc. As I was fixing this I noticed that undefined and other falsey values of reachability were given a value of 101 so put them above 100% (which I think is what led to this bug implementation in the first place).

I noticed as I was fixing it that the code could be much more performant and use less CPU cycles, too 😄 

**Before**
The checks list page with ascending reachability. The checks show 0, N/A, 0, 100, 100 for their reachability values.
The checks list page with descending reachability. The checks show 48.5, 59.8, 98.6, 100, 100 for their reachability values.

**After**

The checks list page with ascending reachability. The checks show N/A, 48.5, 59.6  for their reachability values.
The checks list page with ascending reachability. The checks show 100, 100, 100, 100, 100 for their reachability values.

### DIFF
--- a/src/components/CheckList/CheckList.tsx
+++ b/src/components/CheckList/CheckList.tsx
@@ -253,21 +253,21 @@ function sortChecks(checks: Check[], sortType: CheckSort, reachabilitySuccessRat
 
   if ([CheckSort.ReachabilityAsc, CheckSort.ReachabilityDesc].includes(sortType)) {
     const checkWithMetrics = checks.map((check) => {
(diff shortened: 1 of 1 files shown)


VikaCep (approved):
💯 💯 💯

w1kman (approved):
LGTM 👌🏻

Issue #822, opened 2024-06-12
Re-design probe selection when creating / editing checks


ckbedwell:
**What would you like to be added**:
To reevaluate and update the design for selecting probes when creating and editing checks. Currently the multi-select pattern doesn't help a huge amount for visibility of what is available or help group the probes in any meaningful way.

It would be ideal if we could find a way to display all of the probes simultaneously and also in a way where they are organised by region and/or cloud provider.

![Screenshot of the probes in the multi-select UI pattern.](https://github.com/grafana/synthetic-monitoring-app/assets/6906380/2dc9e3f3-a9f5-4ff5-b11e-07e6754865b9)

**Why is this needed**:
Users may want to pick probes based on their region and/or cloud provider. At the moment they have to go back and forth to different parts of the app and documentation to gather all of that information.

**Additional context**
[Issue for the synthetic-monitoring-api to provide the cloud provider information in the probes list endpoint.](https://github.com/grafana/synthetic-monitoring-api/issues/884)

mem:
I feel I said this before, we can probably take a hint from the metrics explore interface, where the UI presents the existing labels to narrow down the search.

![image](https://github.com/grafana/synthetic-monitoring-app/assets/189701/7d8028a5-8811-48e5-aca1-91625a64fcfc)

Selecting some labels narrows the list of displayed metrics to the ones where those labels exist.

We might to something along those lines for probes: all of them have region, and optionally they also have labels.

mem:
Tangentially related, there's an old idea about "probe groups", where the user is able to select probes based on predefined groups. Maybe an option is to be able to select probes based on labels. The problem is that the users cannot add labels to public probes, that's the reason why the idea of probe groups (which are editable by the user) exists.

https://docs.google.com/document/d/1IUeg5nwEpJU0C8M29ChE4WYU7Z5IVV1G0QgUqcECcDI/edit#heading=h.5sybau7waq2q

majavojnoska:
Hello team, 
I've created two proposals to test and improve the probe selection process. Please review them and share your feedback.

[...]

PR #973, opened 2024-10-25
Re-design probe selection when creating / editing checks

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/822

Implements option 2 from https://github.com/grafana/synthetic-monitoring-app/issues/822#issuecomment-2404777020 where we display the probe locations in checkbox lists separated in columns according to whether they're private or the region they belong to.

image
image


src/components/CheckEditor/CheckProbes/PrivateProbesAlert.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const [display, setDisplay] = useState(true);
+
+  if (!display) {
+    return null;
+  }
+
+  return (
+    <Alert
+      title="You haven't set up any private probes yet."
+      severity="info"
```
VikaCep, line 17: The original design used a different color and icon for this message, but with the intention of reusing and not modifying the elements we get from `grafana/ui` for consistency with other parts of the app, I've set a `info` severity which displays in light blue. Happy to adapt it to the original design if you think that's better though.

image
image

src/components/CheckEditor/CheckProbes/ProbesFilter.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Probe } from 'types';
+
+export const ProbesFilter = ({ probes, onSearch }: { probes: Probe[]; onSearch: (probes: Probe[]) => void }) => {
+  const styles = useStyles2(getStyles);
+
+  const handleSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const searchValue = event.target.value.toLowerCase();
+
+    const filteredProbes = probes.filter(
+      (probe) => probe.region.toLowerCase().includes(searchValue) || probe.name.toLowerCase().includes(searchValue)
```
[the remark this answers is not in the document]
ckbedwell: So, on that note, let's create our own internal mapping of it and make a tech-debt ticket that we would rather the api provides us this info at some point in the future. Our probes are static enough (despite the current migrations...) that I feel comfortable we can hard-code this in our repo as it gives a huge UX boost to the user.

As a user I'd just be confused why the [docs website](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/create-checks/public-probes/) provides this level of info but the app doesn't.

A table of Public Probes in the Americas. It shows the Location, Country and Cloud Provider.

src/components/CheckEditor/CheckProbes/PrivateProbesAlert.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+import { Alert, LinkButton, Stack } from '@grafana/ui';
+
+import { ROUTES } from 'types';
+import { getRoute } from 'components/Routing.utils';
+
+export const PrivateProbesAlert = () => {
+  const [display, setDisplay] = useState(true);
```
ckbedwell, line 8: Let's keep a localStorage memory of their dismissal of this banner. Some people just won't be interested in setting up private probes so I don't see the point in us reminding them for a feature they aren't interested in.

src/components/CheckEditor/CheckProbes/PrivateProbesAlert.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <Alert
+      title="You haven't set up any private probes yet."
+      severity="info"
+      onRemove={() => {
+        setDisplay(false);
+      }}
+    >
+      <Stack gap={1} direction={'column'} alignItems={'flex-start'}>
+        Probes are automated tools that test websites and apps. They act like users by sending requests and checking the
+        responses.
```
ckbedwell, line 24: Let's update the copy here as it isn't too specific to private probes but probes on a whole. @heitortsergent any thoughts?

src/components/CheckEditor/CheckProbes/ProbesList.tsx
```diff
@@ -0,0 +1,121 @@
...
+          {title} ({probes.length})
+        </Label>
+      </div>
+      <div className={styles.probesList}>
+        {probes.map((probe: Probe) => (
+          <div key={probe.id} className={styles.item}>
+            <Checkbox
+              id={`probe-${probe.id}`}
+              onClick={() => handleToggleProbe(probe)}
+              checked={selectedProbes.includes(probe.id as number)}
```
ckbedwell, line 63: We need to fix the typings for when things are in a draft versus published state at some point so we can avoid this kind of type casting 😅

src/components/CheckForm/CheckForm.tsx
```diff
@@ -196,9 +196,9 @@ export const CheckForm = ({ check, disabled, pageTitle }: CheckFormProps) => {
                 <CheckFormAlert />
               </FormLayout.Section>
               <FormLayout.Section label="Execution" fields={[`probes`, `frequency`, ...probesFields]} status={status}>
+                <CheckUsage checkType={checkType} />
                 <CheckProbeOptions checkType={checkType} />
                 {ProbesComponent}
-                <CheckUsage checkType={checkType} />
```
ckbedwell, line 201: We agreed with @majavojnoska we'd keep this at the bottom for now so the workflow made more sense.

Configure probes -> set frequency -> review usage -> submit

If we move it to the top it goes:

Configure probes -> set frequency -> scroll back to the top of the page -> review usage -> scroll to the bottom of the page -> submit

src/components/CheckForm/checkForm.utils.ts
```diff
@@ -57,14 +56,9 @@ function getFirstInput(errs: FieldErrors<CheckFormValues>) {
 }
 
 function searchForSpecialInputs(errKeys: string[] = []) {
-  const probes = errKeys.includes(`probes`) && document.querySelector(`#${PROBES_SELECT_ID} input`);
```
ckbedwell, line 60: I think we could keep this still? Focus on the search input for the probe component?

ckbedwell (changes requested):
🔥 🔥 🔥 Awesome job so far. Few comments I've left inline.

Two slightly more general comments.

There actually is a third (!) state that a checkbox can have and we do utilise at Grafana: [indeterminate](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate) so we could add that in situations where not all the locations are used in a region.

Grafana Dashboard page showing indeterminate state with folder selection versus The Probe region state which shows an unticked state.

We'll have to check this one with @majavojnoska but I wasn't sure if the brackets beside the region header were supposed to be showing how many regions are available or how many you have selected currently?

So in the following example it would be: Private Probes (1) - AMER (1) - EMEA (0)
The Private probe selection component. The region headers have an accompanying number which shows the number of available probes rather than what is selected.

src/components/CheckEditor/CheckProbes/PrivateProbesAlert.tsx
VikaCep: Changed it to the following, based on the text from the docs (https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/set-up/set-up-private-probes/#set-up-private-probes)

image

src/components/CheckEditor/CheckProbes/PrivateProbesAlert.tsx
VikaCep: Good idea 👍

src/components/CheckForm/checkForm.utils.ts
VikaCep: Done 👍

src/components/CheckEditor/CheckProbes/ProbesFilter.tsx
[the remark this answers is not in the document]
VikaCep: Good idea. Addressed in https://github.com/grafana/synthetic-monitoring-app/pull/973/commits/c4a5d4251449e319d50a68c7c2c9a27e1b27cd3e and https://github.com/grafana/synthetic-monitoring-app/pull/973/commits/32e47753c7a53da2f8ca7f721977d6bbd893b146

VikaCep:
I believe I addressed all your comments @ckbedwell 🕺 Let me know otherwise, thanks!

majavojnoska:
> We'll have to check this one with @majavojnoska but I wasn't sure if the brackets beside the region header were supposed to be showing how many regions are available or how many you have selected currently?
> 
> So in the following example it would be: Private Probes (1) - AMER (1) - EMEA (0) The Private probe selection component. The region headers have an accompanying number which shows the number of available probes rather than what is selected.

At first, I thought it would be good to show how many probes are available. But now that you've asked, I think it makes more sense to show how many probes you've actually chosen.

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react';
+
+export const useLocalStorage = <T>(key: string, initialValue: T): [T, (value: T) => void] => {
```
w1kman, line 3: I'm not super keen on this hook.

- This hook will always set a default value. Even if we change the default value at some point, the user will still be stuck on the old default (without making an active choice).

- The hook has no context/listeners to sync value change between components.
Recording 2024-11-05 102209


I'd rather see us using `usehooks-ts`.
Note: even if we'd use `usehooks-ts`, I'd suggest to use the option `initializeWithValue: false`

I could live with this hook existing if:
- keys were namespace (eg `grafana.syntheticMonitoring[<instanceId>].<key>`)
- There were comments (pref JSDoc) clearly stating that state is not synced
- `initialValue` changed name to `defaultValue`, with an option `{setDefaultValue: Boolean = false}` (which would return the default value without writing it to localStorage).
- (feature request 😅) if the hook would would allow the developer to select `local|session` storage with an option/param eg. `useClientStorage<T>(key: string, defaultValue?: T, options?: ClientStorageOptions)`

w1kman (changes requested):
Don't care so much about the comments. Overall I love it! 🤩 

The checkbox and probe identifier are not aligned
image

Have we considered including private probes in their respective region (including the fact that a private prove can have no region) (there is no indication of what region a private probe belongs to)?

The count feels like a bug
image

expectations
image



It's a bit

w1kman:
I wasn't expecting the check to be updated/created when I pressed `Enter` in the filter input
image

src/hooks/useLocalStorage.ts
VikaCep: Added `usehooks-ts` instead, love the set of hooks it provides!

VikaCep:
Thanks for the review @w1kman! Really good suggestions 🎖️ 
I believe I've addressed all of them.

> Have we considered including private probes in their respective region (including the fact that a private prove can have no region) (there is no indication of what region a private probe belongs to)?

Regarding this, we could do that as the API provides us with the region. But I also think it's good to be able to differentiate them from all other public locations in a separate column as a private probe for easy access. We don't have the provider or the country name as shown in the designs (https://www.figma.com/design/WXbmUSpCMlJHG8fpjvmGT1/%F0%9F%8F%97%EF%B8%8F-Synthetics-%5BDEV-READY%5D?node-id=2244-15776&node-type=frame&m=dev) but I wanted to respect the intention of showing them separately from all other probes. What do you think about Thomas idea, @majavojnoska?

ckbedwell:
I need to re-review the code but wanted to leave a quick comment with two things I noticed UX-wise:

Can we add some sort of clear or x button when the search input has a value? This is a screenshot below when I've just put a space, so it's hidden a load of the probes but it isn't clear there are more probes and/or why they are hidden:

Screenshot of the new probe selection UI. The search input and two probe regions with their respective probes are present.

And on a similar note, if you search for something and no probes match, can we have a zero state mentioning no probes match that criteria or some such?

Screenshot of the new probe selection UI. The search input is present with the text "no matches" and nothing else probe-related is rendered beneath it.

src/page/NewCheck/__tests__/NewCheck.test.tsx
```diff
@@ -99,16 +99,6 @@ describe(`<NewCheck />`, () => {
...
 
-  it(`should focus the probes select correctly when appropriate`, async () => {
-    const { user } = await renderNewForm(CheckType.HTTP);
-
-    await fillMandatoryFields({ user, checkType: CheckType.HTTP, fieldsToOmit: ['probes'] });
-    await submitForm(user);
-
-    const probesSelect = await screen.findByLabelText(/Probe locations/);
-    await waitFor(() => expect(probesSelect).toHaveFocus());
-  });
```
ckbedwell, line 110: Can probably put this back with the suggestion above? :P

ckbedwell (changes requested):

VikaCep:
Improved the filter input by adding a Clear button when it has value and a message when there are no results:

https://github.com/user-attachments/assets/2f5fe8b4-20e2-4012-9be6-e527c710fdc9

ckbedwell (approved):
LGTM! 👏 🔥 🌶️

majavojnoska:
> What do you think about Thomas idea, @majavojnoska?

I'm unsure if Thomas's proposal is to categorize public probes by region or to mix public and private probes within the same regional groups. I was hesitant to add regions to private probes because it seemed overly granular and because of: 

**User flexibility:** As Chris explained to me, users can input any coordinates they desire, which may not align with specific regions.

**Usage patterns:** The data shows that users primarily create a small number of private probes (average of 1.66 per user). Assigning regions to these could clutter the UI and potentially hide the private probes.

**Promotional opportunity:** A dedicated section for private probes can serve as a valuable promotional space for this feature.

**Some usage stats for private probes:**
* Total users with private probes: 9.60%
* Number of private probes per user:
    * 1 probe: 7.64% (895 users)
    * 2 probes: 0.89% (105 users)
    * 3-5 probes: 0.58% (63 users)
    * 5-10 probes: 0.34% (40 users)
    * 10-15 probes: 0.06% (8 users)
    * More than 16 probes: 0.06% (8 users)
* Average private probes per user:
    * For users with private probes: 1.66
    * For all users: 0.15

**Usage stats for public probes: 98.32%**
* Total users with public probes: %
* Number of public probes per user:
    * 0 probes: 1.67% ( 196 users)
    * 1 probe: 34.74% ( 4 608 users)
    * 2 probes: 14.54% ( 1 703 users)
    * 3-5 probes: 22.73% ( 2 662 users)
    * 5-10 probes: 8.05% ( 943 users)
    * 10-15 probes: 2.04% ( 240 users)
    * More than 16 probes: 11.58% ( 1 356 users)
* Average public probes per user: 4.85

PR #977, opened 2024-11-01
Autofix: Ignore empty labels when validating check

lattaai1:
I have modified the `getBasePayloadValuesFromForm` function in the `toPayload.utils.ts` file to filter out labels that don't have both a key and a value. This change will ensure that empty label entries are ignored when validating check submission, addressing the issue where users were unable to create checks if they had added a label entry without filling it in. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission


ckbedwell (changes requested):
I don't think this is right. What about updating the schema, too?

Maybe we should remove the whole check form instead, what do you think?

Issue #975, opened 2024-10-29
Remove the deprecated react router used in the app


peterschretlen:
To build a future-proof solution for #892 as well as get rid of code debt tied to `react-router-dom@v5`, we need to upgrade to v6.

Related to https://github.com/grafana/synthetic-monitoring-app/issues/892

Issue #854, opened 2024-07-05
Update routing for check creation


ckbedwell:
_Note this should be addressed after the [check creation redesign](https://github.com/grafana/synthetic-monitoring-app/pull/847) has been merged._

# Problem

Currently our routing for check creation is confusing and unpredictable. The breadcrumbs in-app don't represent the URL routing and it is impossible to go back to the 'Add new check' page.

# Completion criteria

Update the check routing so it is predictable and each step of the process is easy to navigate by both breadcrumbs and modifying the URL directly.

Issue #875, opened 2024-07-22
Add documentation to how our testing framework is setup and how to write tests


ckbedwell:
## Problem

We have quite a comprehensive testing framework that does a lot of the heavy lifting for making writing tests as pain free as possible however none of this is documented.

## Completion criteria

Please create documentation in the repo that outlines:

- [ ] How our testing framework is set up and initialised
- [ ] What assumptions it makes and does for you out of the box (e.g. the custom render method, MSW happy paths)
- [ ] What helper functions are available (e.g. `runTestAsViewer`), options you can provide the custom `render` method (e.g. routing options, custom meta, etc.)
- [ ] How to write tests
- [ ] Outlines how larger features have a meta testing framework with their own helpers to reduce writing boilerplate. See PR: https://github.com/grafana/synthetic-monitoring-app/pull/847

PR #980, opened 2024-11-01
chore: `react-router-dom` migration

w1kman:
Update code base to use `react-router-dom@v6` APIs with `react-router-dom-v5-compat`.
v5 is obsolete and Grafana now bundles with v6 (opt-in, actually `react-router-dom-v5-compat`). 

With these changes, SM removes a lot of code debt tied to `react-router-dom`.

This PR also updates routing with new paths and some and a `NotFound` component.

**Route changes**
From `/checks/edit/<id>/<checkTypeGroup>` to `/checks/<id>/edit`
From `/checks/<id>/dashboard` to `/checks/<id>`
From `/probes/edit/<id>` to `/probes/<id>/edit` (edit) | `/probes/<id>` (view)

**Not found page**

`/checks/2230/not-a-valid-route`
image
Note: An invalid check ID in the route will keep the old (redirect) behavior (for now).

`/checks/new/not-a-valid-check-type-or-group`
image

`/probes/<invalid-id>`
image

*View mode for private probes*
There is now a View mode for the private probe edit page
image

Those who can edit can click the edit button to switch to edit mode. Those without sufficient rights to edit, will not see the button.

**Breadcrumbs**
Attempts have been made to improve breadcrumbs (mainly checks and probes routes).

image

image

image

Resolves: https://github.com/grafana/synthetic-monitoring-app/issues/854
Resolves: https://github.com/grafana/synthetic-monitoring-app/issues/975

Getting tests to work again was a battle. Pay extra attention to strange test setup etc (very likely there are some exploratory code in the code base). 

Try to make the routes break! 🙇🏻


src/components/ProbeEditor/ProbeEditor.tsx
```diff
@@ -31,9 +32,11 @@ export const ProbeEditor = ({
   probe,
   submitText,
   supportingContent,
+  readOnly,
```
ckbedwell, line 35: Can we add a few tests as a safety and implicit documentation for the `readOnly` prop?

My worry is reading this in the future it is confusing  there is a `readOnly` prop and also a `useCanEditProbe` hook so it looks like something is redundant and can be removed.

I think much of the confusion stems from the component names. Should we make a `ViewProbe` component to add to the router? Maybe we can tackle that when we do https://github.com/grafana/synthetic-monitoring-app/issues/848?

src/page/NewProbe/NewProbe.test.tsx
```diff
@@ -25,15 +25,15 @@ it(`creates a new probe, displays the modal and redirects on close`, async () =>
...
+  const { user } = renderNewProbe();
   await fillProbeForm(user);
 
   const saveButton = await screen.findByText('Add new probe');
   await user.click(saveButton);
   await waitFor(() => expect(screen.queryByText(ADD_PROBE_TOKEN_RESPONSE)).toBeInTheDocument());
   const dismiss = screen.getByText('Go back to probes list');
   await user.click(dismiss);
-  await waitFor(() => expect(history.location.pathname).toBe(getRoute(ROUTES.Probes)));
+  // await waitFor(() => expect(history.location.pathname).toBe(getRoute(ROUTES.Probes)));
```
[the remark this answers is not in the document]
ckbedwell: Nit: I see a test AND commented lines 😬

src/test/render.tsx
```diff
@@ -28,16 +31,64 @@ type CreateWrapperProps = {
...
-              </Router>
+              <MemoryRouter initialEntries={history.entries}>
+                <CompatRouter>
+                  <Routes>
+                    <Route path={APP_ROOT}>
+                      <Route
+                        path={route}
+                        element={
+                          <>
+                            <TestRouteInfo />
```
ckbedwell, line 64: Can we document what this is doing, please.  It took me a while to work out its purpose and had to look at some of the tests. Would help me out on getting started on this ticket, too: https://github.com/grafana/synthetic-monitoring-app/issues/875 🙏

ckbedwell (changes requested):
This is so good and is paying off a ton of debt. I didn't even find the 79 files that daunting because of how much good work it was doing 🙏🙏🙏

Main asks are a few of the mundane dev tasks: tests and documentation.

The only thing I didn't get a chance to test before I wrap up for the week (and go on PTO for another week) is what happens with all the existing routes and if the redirects are handled correctly. I saw some legacy redirect logic but didn't get a chance to test it.

src/components/Routing.tsx
```diff
@@ -38,35 +48,67 @@ export const InitialisedRouter = ({ onNavChanged }: Pick<AppRootProps, 'onNavCha
...
+      <Route path={ROUTES.Alerts} element={<AlertingPage />} />
+
+      <Route path={ROUTES.Config} element={<ConfigPage initialized />} />
+
+      <Route path={ROUTES.Redirect} element={<SceneRedirecter />} />
+
+      <Route path={ROUTES.Scene} element={<SceneRedirecter />} />
+
+      {/* TODO: Create 404 instead of navigating to home(?) */}
+      <Route path="*" element={<Navigate to={`../${ROUTES.Home}`} replace />} />
```
VikaCep, line 110: Wouldn't this work as a 404 page?

```suggestion
       <Route
          path="*"
          element={
            <PluginPageNotFound>
              The page you are looking for does not exist. Here is a working link to{' '}
              <TextLink href={getRoute(ROUTES.Checks)}>checks listing</TextLink>.
            </PluginPageNotFound>
          }
        />
```

src/page/ChecksPage.test.tsx
```diff
@@ -1,169 +1,201 @@
...
+                  </Routes>
+                </CompatRouter>
+              </MemoryRouter>
+            </PermissionsContextProvider>
+          </SMDatasourceProvider>
+        </FeatureFlagProvider>
+      </MetaContextProvider>
+    </QueryClientProvider>
+  );
+}
```
VikaCep, line 47: image

😅😅😅

src/test/fixtures/datasources.ts
```diff
@@ -83,6 +83,9 @@ export const METRICS_DATASOURCE: DataSourceInstanceSettings<any> & WithAccessCon
       grafanaDependency: '',
       grafanaVersion: '*',
       plugins: [],
+      extensions: {
+        exposedComponents: [],
+      },
```
VikaCep, line 88: Just wondering, but why was this needed?

VikaCep (commented):
This looks really great 🎉🎉🎉 
Thank you for putting the time into this! I know fixing the tests wasn’t trivial 🙏  
I’ve left a few comments and questions, but overall, everything looks fantastic. I tried to make the routes break, but I wasn't successful 💯

src/test/fixtures/datasources.ts
w1kman: Typescript

src/components/CheckForm/useCheckType.tsx
```diff
@@ -15,7 +15,17 @@ export function useFormCheckType(existingCheck?: Check) {
...
+
+export function useFormCheckTypeGroup(check?: Check) {
+  const { checkTypeGroup } = useParams<CheckFormPageParams>();
+
+  if (check) {
+    const checkType = getCheckType(check.settings);
+    return CHECK_TYPE_OPTIONS.find(({ value }) => value === checkType)?.group;
+  }
+
+  return checkTypeGroup;
```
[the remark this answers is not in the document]
w1kman: @VikaCep are you we want to do this? 
We'd have to add `// eslint-disable-next-line react-hooks/rules-of-hooks` before the conditional hook.

> ESLint: React Hook "useParams" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?(react-hooks/ rules-of-hooks)

src/components/CheckEditor/FormComponents/ChooseCheckType.tsx
```diff
@@ -32,7 +33,7 @@ export const ChooseCheckType = ({ checkType, checkTypeGroup, disabled }: ChooseC
 
     const values = updateCheckTypeValues(ref.current, newCheckType, checkType);
     reset(values);
-    history.replace({ search: `?checkType=${newCheckType}` }); // todo: preserve all query params
+    navigate({ search: `?checkType=${newCheckType}` }, { replace: true }); // todo: preserve all query params
```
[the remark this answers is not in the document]
w1kman: 🤷🏻 I'll remove the comment.

src/page/NewCheck/NewCheck.tsx
```diff
@@ -1,16 +1,30 @@
...
-import { CheckFormPageParams } from 'types';
+import { CheckFormPageParams, ROUTES } from 'types';
 import { CHECK_TYPE_GROUP_OPTIONS } from 'hooks/useCheckTypeGroupOptions';
 import { CheckForm } from 'components/CheckForm/CheckForm';
+import { getRoute } from 'components/Routing.utils';
+
+import { PluginPageNotFound } from '../NotFound/NotFound';
 
 export const NewCheck = () => {
   const { checkTypeGroup } = useParams<CheckFormPageParams>();
```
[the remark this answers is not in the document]
w1kman: IMO, using `useFormCheckTypeGroup` doesn't add anything. The use of `useParams` is a fallback mechanism in `useFormCheckTypeGroup` (for when `check` is `undefined`).

src/test/render.tsx
w1kman: I added some JSDoc to the component.
Depending on your IDE, it should allow for some additional clarification

**Using JetBrains (GoLand)**
image

src/components/ProbeEditor/ProbeEditor.tsx
w1kman: - updated the prop name `forceViewMode`
- added test for the new prop

This is an intermediate solution for the lack of a `ViewProbe` component.

src/test/render.tsx
ckbedwell: A screenshot of the JSDoc info for TestRouteInfo in VSCode.

📜🥳 🎉

package.json
```diff
@@ -80,23 +84,26 @@
...
-    "@grafana/data": "11.0.0",
+    "@emotion/css": "11.10.6",
+    "@grafana/data": "^11.3.0",
     "@grafana/faro-web-sdk": "1.3.6",
-    "@grafana/runtime": "11.0.0",
+    "@grafana/runtime": "^11.0.0",
     "@grafana/scenes": "5.1.0",
     "@grafana/schema": "11.0.0",
-    "@grafana/ui": "11.0.0",
+    "@grafana/ui": "^11.0.0",
```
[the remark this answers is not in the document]
ckbedwell: Sorry, I should have been clearer. I am all for updating the packages -- little and often is great and if we can continually update the Grafana packages especially so they are as close to the runtime as possible that's great.

My concern is in utilising the `^` caret syntax for the dependencies. I'm not a fan of it because `package.json` becomes opaque and you have to delve into `yarn.lock` to know what you are really using. It also assumes that every package you utilise accurately follows semver and I have trust issues after this being a source of bugs in other projects I have worked on previously 😅

src/components/Routing.tsx
ckbedwell: Sure thing 👍

package.json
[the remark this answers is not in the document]
w1kman: I can see why pinning `@grafana/*` versions makes sense (even though it doesn't mean that the code will run with that version in the wild), but pinning "other" packages for no other reason than "what if..." seems like a false sense of security. Many of our dependencies have their own dependencies that will install `minor` and `patch`, meaning that we have no actual control over potential side-effects by installing a minor/patch.
Also, when two (including our package) or more packages rely on the same major version, they may share the same package instead of installing their specific version.

Note: is `bundleGrafanaUI: false` something we want to use? I think it ties into this.

TL;DR: I could live with (but, wouldn't recommend) pinned versions with `~` (so that patches are installed).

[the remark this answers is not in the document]
ckbedwell: Alright, let's agree to pin the runtime `@grafana` packages and update the plugin.json grafanaDependency to match.

We haven't encountered any problems with the caret syntax and you are right, it does help manage the dependency tree better.

> Note: is bundleGrafanaUI: false something we want to use? I think it ties into this.

No, I don't think we do. It just creates a different set of problems and goes against the grain of what everyone else is doing. Because we have SLOs with CVEs now we should naturally get a pretty good rhythm of keeping our dependencies updated.

ckbedwell (approved):
🔥🔥🔥 LGTM 🔥🔥🔥

VikaCep (approved):
👏 👏 👏

PR #982, opened 2024-11-04
Feat: Add RBAC to Synthetic Monitoring

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring/issues/166
Closes https://github.com/grafana/synthetic-monitoring/issues/167

This PR adds the configuration for the new roles associated with the different actions that can be performed within the plugin by following the [RBAC in app plugins](https://grafana.com/developers/plugin-tools/how-to-guides/app-plugins/implement-rbac-in-app-plugins) approach, and enforces this permissions in the frontend.

There's this [document](https://docs.google.com/document/d/1GjtdFV5l00ubtoORnfUJnGmxbQU4u7krcU58uSdaH7k/edit?tab=t.0) where different scenarios are combined to show how the app behaves in those cases.

I've taken the list from [166](https://github.com/grafana/synthetic-monitoring/issues/166) and created the following categories so that users and teams can be assigned one or several of the following:

**Admin:** Provides permissions to manage all aspects of the app, including enabling/disabling the plugin and managing checks, probes, alerts, thresholds, and access tokens.

**Editor:** Similar to Admin access but does not include enabling/disabling the plugin or creating/deleting tokens.

**Reader:** Grants read access to all components (checks, probes, alerts, thresholds, and access tokens) in the app. 

**Checks Reader:** Allows users to read checks within the app. 
**Checks Writer:** Enables users to create, edit, and delete checks. 

**Probes Reader:** Provides read access to probes. 
**Probes Writer:** Allows users to create, edit, and delete probes.

**Alerts Reader:** Permits users to read alerts. 
**Alerts Writer:** Enables the creation, editing, and deletion of alerts. 

**Thresholds Reader:** Grants read access to thresholds.
**Thresholds Writer:** Allows users to read and edit thresholds,.

**Access Tokens Writer:** Permits the creation and deletion of access tokens. 

image
image


src/plugin.json
```diff
@@ -73,6 +77,226 @@
...
+        "name": "Checks reader",
+        "description": "Read checks in the Synthetic Monitoring app",
+        "permissions": [
+          { "action": "plugins.app:access", "scope": "plugins:id:grafana-synthetic-monitoring-app" },
+          { "action": "grafana-synthetic-monitoring-app:read" },
+
+          { "action": "grafana-synthetic-monitoring-app.checks:read" }
+        ]
+      },
+      "grants": ["Admin", "Editor", "Viewer"]
```
d0ugal, line 92: Why do we grant this to Admin and Editor? Since they get these permissions with the Checks writer? We didn't do this in k6 and now I'm wondering if we should.

src/plugin.json
```diff
@@ -73,6 +77,226 @@
...
+          { "action": "grafana-synthetic-monitoring-app:read" },
+
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:read" }
+        ]
+      },
+      "grants": ["Admin", "Editor"]
+    },
+    {
+      "role": {
+        "name": "Full admin access",
```
d0ugal, line 226: I have not seen "Full" in the names before. In k6-cloud we just call this "Admin" and then the description explains it more.

src/plugin.json
```diff
@@ -73,6 +77,226 @@
...
+          { "action": "grafana-synthetic-monitoring-app.alerts:delete" },
+          { "action": "grafana-synthetic-monitoring-app.thresholds:delete" },
+          { "action": "grafana-synthetic-monitoring-app.tokens:delete" }
+        ]
+      },
+      "grants": ["Admin"]
+    },
+    {
+      "role": {
+        "name": "Full write access",
```
d0ugal, line 256: Following other existing roles I think this should be "Synthetic Monitoring Writer". For example there is the "Folders Writer" which gives you write access to any folder.

src/plugin.json
```diff
@@ -73,6 +77,226 @@
...
+        "permissions": [
+          { "action": "plugins.app:access", "scope": "plugins:id:grafana-synthetic-monitoring-app" },
+          { "action": "grafana-synthetic-monitoring-app:read" },
+          { "action": "grafana-synthetic-monitoring-app:write" },
+
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:create" },
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:delete" }
+        ]
+      },
+      "grants": ["Admin", "Editor"]
```
d0ugal, line 209: Do we want Editors to create access tokens? This feels like an admin action to me.

src/plugin.json
d0ugal: Screenshot 2024-11-13 at 13 27 18
 Also SLO

src/plugin.json
VikaCep: Yeah I think it makes more sense to be restricted to admins. Changed it.

src/plugin.json
VikaCep: I saw it somewhere else, but probably it was an old PR and was changed later. Totally in favor of using the more conventional names, so changed them to `Admin`, `Editor` and `Reader`.

src/plugin.json
VikaCep: You're right, I think it's redundant. Removed it.

src/datasource/plugin.json
w1kman, line 1: Perhaps I'm not understanding this correctly, but it seems a little loose to allow someone that can write, to modify the data source. Seems like that should have its own role and only be assigned to admins.

w1kman: I guess this has nothing to do with adding/removing the data source 😅

w1kman (commented):
Is this prep work or is it possible to test the roles?

src/plugin.json
```diff
@@ -73,6 +77,224 @@
...
+    {
+      "role": {
+        "name": "Access tokens writer",
+        "description": "Create and delete access tokens in the Synthetic Monitoring app",
+        "permissions": [
+          { "action": "plugins.app:access", "scope": "plugins:id:grafana-synthetic-monitoring-app" },
+          { "action": "grafana-synthetic-monitoring-app:read" },
+          { "action": "grafana-synthetic-monitoring-app:write" },
+
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:create" },
```
d0ugal, line 205: I see we have `access-tokens` here but just `tokens` (for example in ` { "action": "grafana-synthetic-monitoring-app.tokens:read" },`) in other places. I'm not sure if this is intentional or if they are the same tokens?

VikaCep:
> Is this prep work or is it possible to test the roles?

This is the roles definitions only. In this other [PR](https://github.com/grafana/synthetic-monitoring-app/pull/986) I consume the roles and apply the restrictions in the UI.

src/plugin.json
VikaCep: I changed it to only keep `access-tokens` 👍

src/datasource/plugin.json
VikaCep: In order to write checks or probes, users need to be able to write the SM datasource. For `POST` and `DELETE` operations to `"{{.JsonData.apiHost}}/api/v1/"` I've set a `reqAction` of `grafana-synthetic-monitoring-app:write` which before, it was `Editor`. This custom writer role is included in roles that perform write operations (such as Checks Writer, Probes Writer, etc). 

Now, for the `{{.JsonData.apiHost}}/api/v1/register/save` url, I changed the required permission to be `grafana-synthetic-monitoring-app.plugin:enable` which is only assigned to `Admins`. 

As an example, this is what happens if I try to edit a check and I don't have the permissions defined in the `POST {{.JsonData.apiHost}}/api/v1/"`  section of the `datasources/plugin.json` file:

https://github.com/user-attachments/assets/b428759c-6239-494b-89f8-41aeaa92f994

src/plugin.json
```diff
@@ -73,6 +78,224 @@
...
+
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:write" },
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:delete" }
+        ]
+      },
+      "grants": ["Admin"]
+    },
+    {
+      "role": {
+        "name": "Access tokens reader",
```
d0ugal, line 214: Where can access tokens be read? iirc they are only visible when they are generated. If they can be read as an editor, does that essentially mean the user can escalate themself to admin by using the token with the API and avoid RBAC?

src/plugin.json
```diff
@@ -73,6 +78,224 @@
...
     }
   ],
+  "roles": [
+    {
+      "role": {
+        "name": "Checks reader",
+        "description": "Read checks in the Synthetic Monitoring app",
+        "permissions": [
+          { "action": "plugins.app:access", "scope": "plugins:id:grafana-synthetic-monitoring-app" },
+          { "action": "grafana-synthetic-monitoring-app:read" },
```
d0ugal, line 88: What does this action provide beyond the above plugin access?

src/plugin.json
```diff
@@ -73,6 +78,224 @@
...
+      "grants": ["Viewer"]
+    },
+    {
+      "role": {
+        "name": "Checks writer",
+        "description": "Create, edit and delete checks in the Synthetic Monitoring app",
+        "permissions": [
+          { "action": "plugins.app:access", "scope": "plugins:id:grafana-synthetic-monitoring-app" },
+          { "action": "grafana-synthetic-monitoring-app:read" },
+          { "action": "grafana-synthetic-monitoring-app:write" },
```
d0ugal, line 102: As above I'm not sure what the generic write action gives the user.

d0ugal (dismissed):

src/plugin.json
VikaCep: They are actions I added so that they can be used in `datasource/plugin.json` file https://github.com/grafana/synthetic-monitoring-app/pull/982/files/0c573fe1059ca32027fbc2755ec98e84126d87a1#diff-b692ddd2f38b511f155f18798f8101fa7287f2949c496b715163908296df7effR48

Before, we were using org roles to restrict those operations (actually, just `Editor`), so I had to come up with new custom roles that would be appropriate for those endpoints.

src/plugin.json
VikaCep: Same answer as in https://github.com/grafana/synthetic-monitoring-app/pull/982#discussion_r1853908817

src/plugin.json
ckbedwell: They can't be read yet but they should be able to. I probably have dozens (hundreds...) of tokens in my account which I can't revoke and I would love to see them so I can delete them all.

I do find this a difficult part to get my head round RBAC at times. Does it make sense to have separate roles for reading / writing / deleting access tokens or should this be a simple switch of you either have complete control to access tokens or you don't?

I think I can see a use case for saying "I'll grant you access to read our existing tokens so you can set up our SM instance but I don't want you to create any more." and that follows the conventions for all the other RBAC roles.

src/plugin.json
```diff
@@ -73,6 +78,224 @@
...
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:write" },
+          { "action": "grafana-synthetic-monitoring-app.checks:read" },
+          { "action": "grafana-synthetic-monitoring-app.probes:read" },
+          { "action": "grafana-synthetic-monitoring-app.alerts:read" },
+          { "action": "grafana-synthetic-monitoring-app.thresholds:read" },
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:read" },
+          { "action": "grafana-synthetic-monitoring-app.checks:delete" },
+          { "action": "grafana-synthetic-monitoring-app.probes:delete" },
+          { "action": "grafana-synthetic-monitoring-app.alerts:delete" },
+          { "action": "grafana-synthetic-monitoring-app.thresholds:delete" },
```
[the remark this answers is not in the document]
VikaCep: Not intentional, added it.

ckbedwell (dismissed):
LGTM 🥇

VikaCep:
@ckbedwell I have changed the required permissions that are needed to read the SM home page, which previously were requiring Check Reader access. Please let me know if this addresses your questions.

> Here I have the Synthetic Monitoring: Alerts reader role and if I go back to the homepage I get given an Access denied alert.

After changing the mentioned permission, now in the scenario you described we get this view:

image

BTW, any custom SM role that you add (like Checks reader, Alerts reader, etc) will automatically [include](https://github.com/grafana/synthetic-monitoring-app/pull/982/files#diff-3c1bbe21d50836befcccc2df86e0ceb27afbb4444eb924d015be9298df27604eR145) the `  { "action": "plugins.app:access", "scope": "plugins:id:grafana-synthetic-monitoring-app" },` permission so it's not necessary to add the `Plugins:Application Plugins Access` permission to the user manually.

ckbedwell:
> After changing the mentioned permission, now in the scenario you described we get this view:

You only get this view if you don't have SMDS Query permissions. Once you add them you get a skeleton view of the Scenes homepage with the checks (which I shouldn't have because I don't have 'Checks reader' permissions).

A screenshot of Synthetic Monitoring homepage with the checks table with a single column displaying job names.

I wonder if we should add some sort of testing matrix to try and ensure this remains bug-free for the future or to document what the expected behaviour is?

VikaCep:
> I wonder if we should consider adding a testing matrix to help ensure this remains bug-free going forward, or perhaps document the expected behavior more clearly?

Yes, given the number of testing scenarios involved, I'll work on creating a document that outlines all possible cases and their expected behavior to help streamline the review process.

Additionally, some of the issues raised here should be addressed in [this PR](https://github.com/grafana/synthetic-monitoring-app/pull/986), as my focus here was primarily on the roles in the `plugin.json` file. My original plan was to submit two separate PRs—one for defining the new roles and permissions, and another for implementing the frontend enforcements—to first ensure consensus on the proposed roles. Now that we've mostly agreed on that, I think it makes sense to consolidate both PRs into one for easier testing. I'll proceed with merging [PR #986](https://github.com/grafana/synthetic-monitoring-app/pull/986) into this one.

VikaCep:
> > After changing the mentioned permission, now in the scenario you described we get this view:
> 
> You only get this view if you don't have SMDS Query permissions. Once you add them you get a skeleton view of the Scenes homepage with the checks (which I shouldn't have because I don't have 'Checks reader' permissions).
> 
> A screenshot of Synthetic Monitoring homepage with the checks table with a single column displaying job names.

I added `grafana-synthetic-monitoring-app.checks:read` requirement in order to see the homepage. Also, added a generic page that displays the missing actions:


image
(home page with datasources reader and alerts reader role)

src/components/ConfigActions.tsx
```diff
@@ -2,17 +2,18 @@ import React, { useState } from 'react';
...
 
-import { hasGlobalPermission } from 'utils';
 import { ROUTES } from 'routing/types';
 import { getRoute } from 'routing/utils';
 import { useMeta } from 'hooks/useMeta';
+import { usePluginPermissions } from 'hooks/usePluginPermissions';
 
 import { DisablePluginModal } from './DisablePluginModal';
 
 export const ConfigActions = ({ initialized }: { initialized?: boolean }) => {
```
VikaCep, line 12: I believe we're not using this component, right @w1kman?

VikaCep:
> > I wonder if we should consider adding a testing matrix to help ensure this remains bug-free going forward, or perhaps document the expected behavior more clearly?

> 
> Yes, given the number of testing scenarios involved, I'll work on creating a document that outlines all possible cases and their expected behavior to help streamline the review process.

I've created this [document](https://docs.google.com/document/d/1GjtdFV5l00ubtoORnfUJnGmxbQU4u7krcU58uSdaH7k/edit?tab=t.0) where I outline the different role assignments and the app behavior for each of them, with some screenshots and descriptions of the expected behavior.   Hopefully it will be useful when going through this PR.

src/page/AlertingPage.tsx
```diff
@@ -93,13 +97,23 @@ const AlertingPageContent = () => {
...
+      {hasWriterRole && !canWriteAlerts && (
+        <InsufficientPermissions
+          message={
+            <>
+              You are not able to edit alerts because you are missing <code>alert.instances.external:write</code>{' '}
+              permissions
+            </>
+          }
+        />
+      )}
```
w1kman, line 109: Perhaps this could be consolidated with the alert above?

src/page/AlertingPage.tsx
```diff
@@ -93,13 +97,23 @@ const AlertingPageContent = () => {
           {alertError}
         </Alert>
       )}
+      {hasWriterRole && !canWriteAlerts && (
+        <InsufficientPermissions
+          message={
+            <>
+              You are not able to edit alerts because you are missing <code>alert.instances.external:write</code>{' '}
+              permissions
```
w1kman, line 105:
```suggestion
              permission.
```

src/page/AlertingPage.tsx
```diff
@@ -39,7 +39,11 @@ const Alerting = () => {
           Learn more about alerting for Synthetic Monitoring.
         </a>
       </p>
-      {canRead ? <AlertingPageContent /> : <InsufficientPermissions />}
+      {canReadAlerts ? (
+        <AlertingPageContent />
+      ) : (
+        <InsufficientPermissions message="Contact your administrator to ensure you have Query access to the metrics datasource." />
```
w1kman, line 45: Do we know what roles the user should ask the administrator to add?

src/page/AlertingPage.tsx
```diff
@@ -142,10 +161,10 @@ const AlertingPageContent = () => {
   );
 };
 
-const InsufficientPermissions = () => {
+const InsufficientPermissions = ({ message }: { message: string | JSX.Element }) => {
```
w1kman, line 164: Why not use the `title` as well?
image

Would become
image

```
{
  "accessErrorId": "ACE5917147604",
  "message": "You'll need additional permissions to perform this action. Permissions needed: alert.rules.external:read",
  "title": "Access denied"
}
```

src/page/ConfigPageLayout/tabs/AccessTokensTab.tsx
```diff
@@ -42,8 +42,8 @@ export function AccessTokensTab() {
         documentation to learn more about how to interact with the synthetic monitoring API.
         <Space v={2} />
         <Button
-          tooltip={!canCreateAccessToken ? 'You do not have permission to generate access tokens.' : undefined}
-          disabled={!canCreateAccessToken}
+          tooltip={!canWriteTokens ? 'You do not have permission to generate access tokens.' : undefined}
+          disabled={!canWriteTokens}
```
w1kman, line 46: Nit: It would be more helpful to the user to show an alert with a more detailed message.

"To generate an access token, you need to have \<permission>. To add \<permission>, contact an administrator"

src/page/ContactAdminAlert.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { Alert, Stack } from '@grafana/ui';
+
+export const ContactAdminAlert = ({ permissions }: { permissions?: string[] }) => {
+  return (
+    <Alert title="Contact your administrator to get you started." severity="info">
+      {permissions && (
+        <Stack>
+          <div>You are missing the following permissions:</div>
```
w1kman, line 9:
```suggestion
          <div>You are missing the following permission(s):</div>
```

src/plugin.json
```diff
@@ -73,6 +78,224 @@
...
+
+          { "action": "grafana-synthetic-monitoring-app.checks:write" },
+          { "action": "grafana-synthetic-monitoring-app.probes:write" },
+          { "action": "grafana-synthetic-monitoring-app.alerts:write" },
+          { "action": "grafana-synthetic-monitoring-app.thresholds:write" },
+          { "action": "grafana-synthetic-monitoring-app.checks:read" },
+          { "action": "grafana-synthetic-monitoring-app.probes:read" },
+          { "action": "grafana-synthetic-monitoring-app.alerts:read" },
+          { "action": "grafana-synthetic-monitoring-app.thresholds:read" },
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:read" },
```
w1kman, line 272: What does this do?

src/plugin.json
```diff
@@ -73,6 +78,224 @@
...
+      "role": {
+        "name": "Access tokens writer",
+        "description": "Create and delete access tokens in the Synthetic Monitoring app",
+        "permissions": [
+          { "action": "plugins.app:access", "scope": "plugins:id:grafana-synthetic-monitoring-app" },
+          { "action": "grafana-synthetic-monitoring-app:read" },
+          { "action": "grafana-synthetic-monitoring-app:write" },
+
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:write" },
+          { "action": "grafana-synthetic-monitoring-app.access-tokens:delete" }
```
w1kman, line 208: Is this even a thing?
Generating a new token doesn't delete the old one

src/plugin.json
```diff
@@ -73,6 +78,224 @@
...
+        ]
+      },
+      "grants": ["Admin"]
+    },
+    {
+      "role": {
+        "name": "Access tokens reader",
```
w1kman, line 215: This is the first role that pops up when trying to add SM roles to a users.
It's confusing as it doesn't do anything but at the same time tells the user that it allows users to read access tokens (which is usually a no bueno).

IMO this role should be added when it's needed.

src/test/fixtures/rbacPermissions.ts
```diff
@@ -0,0 +1,49 @@
...
+  'grafana-synthetic-monitoring-app.checks:write': true,
+  'grafana-synthetic-monitoring-app.probes:write': true,
+  'grafana-synthetic-monitoring-app.alerts:write': true,
+  'grafana-synthetic-monitoring-app.thresholds:write': true,
+  'grafana-synthetic-monitoring-app.access-tokens:write': true,
+  'grafana-synthetic-monitoring-app.checks:read': true,
+  'grafana-synthetic-monitoring-app.probes:read': true,
+  'grafana-synthetic-monitoring-app.alerts:read': true,
+  'grafana-synthetic-monitoring-app.thresholds:read': true,
+  'grafana-synthetic-monitoring-app.access-tokens:read': true,
```
w1kman, line 14: My nemesis.

w1kman (changes requested):
Great work! 💪🏻 
I mostly have nit comments, that you may or may not want to ignore.

**This is a show stopper**
I'm logged in with base role Admin, and I'm asked to contact an admin to be able to "initialize" the plugin.
image

The uninitalized UI is not super mint. What is the deal with the hover effect?

src/page/AlertingPage.tsx
```diff
@@ -39,7 +39,11 @@ const Alerting = () => {
           Learn more about alerting for Synthetic Monitoring.
```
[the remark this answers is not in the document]
VikaCep: The page doesn't display if the user lacks the `grafana-synthetic-monitoring-app.alerts:read` permission:
image


However, if the user is missing the `alert.rules.external:read` action, I think it would be better to render the Alerting page and show a message indicating this. Without this message, the user might be unaware of the missing permission and could be confused about why they can't access the page.

src/page/AlertingPage.tsx
VikaCep: The check above needs to check the metrics DS can be read in order to display the `AlertingPageContent` component, otherwise it will try to make the request to fetch alerts and fail. So we need to conditionally render according to this condition.

This message, on the other hand, should be rendered within `AlertPageContent` because we want to display the read-only list of alerts. This only indicates the inability to write them.

But in order to make it more consistent, I've consolidated the component used to display the message so that the text remains the same, only changing the list of missing permissions.

src/page/AlertingPage.tsx
VikaCep: We're currently only returning the message associated with the failed request. I could refactor it to also return the title, but since this message is displayed for any request error (not just 403), I'm unsure if the API provides a user friendly title for all error cases.

https://github.com/grafana/synthetic-monitoring-app/blob/main/src/hooks/useAlerts.ts#L72

src/components/ConfigActions.tsx
VikaCep: I have removed it.

src/page/ConfigPageLayout/tabs/AccessTokensTab.tsx
VikaCep: Good idea. Added the alert:

- without permission
image

- with permission
image

src/plugin.json
VikaCep: I have removed the Access Token Reader role as there is no use for it right now and it might confuse users to see it in the list of available roles to assign. We can add it back once we implement the view to actually list and delete them.

src/test/fixtures/rbacPermissions.ts
VikaCep: It's gone 👋

src/plugin.json
VikaCep: I think you're right. Currently there's no use for it so I removed it. Also removed the related `read` and `delete` actions

src/plugin.json
d0ugal: Yup, makes sense. It’s easy to add roles later

src/data/permissions.ts
```diff
@@ -0,0 +1,56 @@
...
+  canWriteThresholds: isUserActionAllowed('grafana-synthetic-monitoring-app.thresholds:write', OrgRole.Editor),
+
+  canReadTokens: isUserActionAllowed('grafana-synthetic-monitoring-app.access-tokens:read', OrgRole.Admin),
+  canWriteTokens: isUserActionAllowed('grafana-synthetic-monitoring-app.access-tokens:write', OrgRole.Admin),
+  canDeleteTokens: isUserActionAllowed('grafana-synthetic-monitoring-app.access-tokens:delete', OrgRole.Admin),
+
+  canWritePlugin: isUserActionAllowed('grafana-synthetic-monitoring-app.plugin:write', OrgRole.Admin),
+
+  canWriteSM: isUserActionAllowed('grafana-synthetic-monitoring-app:write', OrgRole.Editor),
+});
```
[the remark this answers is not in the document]
VikaCep: If I make it a `const`, the role control tests will break because they modify the `bootData` object to assign the desired roles (as seen [here](https://github.com/grafana/synthetic-monitoring-app/pull/982/files#diff-299028e0e9ca5530a7676fdc0eccf97908cba15694317da923acaa817122d144R165)). Unless you think the change would bring a clear benefit, I’d rather leave it as is.

VikaCep:
> This is a show stopper
>I'm logged in with base role Admin, and I'm asked to contact an admin to be able to "initialize" the plugin.

@w1kman I couldn't reproduce that scenario with the base Admin role:
https://github.com/user-attachments/assets/7ee9cf03-6240-48b5-a21a-8797c5e86d47

I did get to that screen by setting the `SM Admin` role and no `datasources:create`. In this case, I changed the alert message to list the required permissions:

image

Once `datasources:create` is added, the message is gone:

image

VikaCep:
> Without checks:read you are able to view all the checks (not metrics) via config/terraform

I restricted the terraform data requiring `checks reader` and `probes reader` permissions as there is probes information as well:
image

w1kman (approved):
LGTM! Great work! 💪🏻

d0ugal (approved):
Since I have about 10 minutes experience with typescript I didn't attempt to review  that. However, the RBAC json changes look good

ckbedwell (approved):
LGTM 🔥 Great work!

w1kman (approved):

Issue #983, opened 2024-11-05
Test executions per month doesn't change


mem:
**What happened**:

Initially the Test executions per month shows 8928 without any locations configured.

When adding one location the number stays the same.

When adding another location the number stays the same.

**What you expected to happen**:

Initially the number is 0 (no locations).

Adding one locations shows 8640 (60 / 5 * 24 * 30)

Adding another location shows 17280.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a check
2. Get to the execution section
3. Look at "test executions per month"
4. Add one location, look at "test executions per month"
4. Add another location, look at "test executions per month"

**Anything else we need to know?**:

**Environment**:
- Grafana version (if running on prem): cloud
- SM plugin version: 1.16.5
- OS Grafana is installed on: Linux
- User OS & Browser: Chrome
- Others:

PR #984, opened 2024-11-05
Fix check usage calculation for MultiHTTP checks

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/983

We were previously calculating the number of check usages for MultiHTTP checks incorrectly by not considering the number of probe locations. This PR fixes this calculation. Additionally, tests have been added to ensure the calculation works as expected moving forward.

image


ckbedwell (approved):
LGTM 🌟

PR #986, opened 2024-11-08
RBAC: enforce permissions in frontend using user roles

VikaCep:
This PR integrates the newly defined roles from [#982](https://github.com/grafana/synthetic-monitoring-app/pull/982) into the app. The roles are used to restrict access to different parts of the app where applicable, providing more granular control over permissions and user access.

Resolves https://github.com/grafana/synthetic-monitoring/issues/167

**Reader**
image

https://github.com/user-attachments/assets/6e2d3492-0aea-4739-af94-9c87895a0235

**Editor**
image

https://github.com/user-attachments/assets/937549c7-bdfe-4118-8b26-5261695e9b2b

**Admin**
image

https://github.com/user-attachments/assets/e8b5fd1b-2cea-4f81-ac5f-95eba1d422eb

**Mixed: Viewer + can write checks**
image

https://github.com/user-attachments/assets/92ba6ab1-68bc-4f11-a991-7b9e0743c0cd

**Mixed: Viewer + can write probes**
image
https://github.com/user-attachments/assets/65d268b9-62c8-4a7f-adb5-a7f5381f7c27


w1kman:
Note: need to supply Grafana with an enterprise license to review this PR. :)

w1kman (commented):
The config page, probably dont need to be visible for non-writers?

Without a basic role (`orgRole), it seems I cannot do much, even if I have SM checks:reader and probe:reader/writer
The only thing that changes are the nav items.

image

VikaCep:
> Without a basic role (`orgRole), it seems I cannot do much, even if I have SM checks:reader and probe:reader/writer The only thing that changes are the nav items.

I've discussed this issue with @d0ugal, as it seems to be a common scenario. If a user is assigned a custom role but lacks the necessary fixed roles required for an action, the custom role alone will not be sufficient. In these cases, it’s important to highlight the missing roles in the UI so that users understand what’s happening and can request the required roles.
As an example, this [PR](https://github.com/grafana/grafana-k6-app/pull/1277) shows how it's handled in k6, by clarifying what are the missing permissions when an action is not allowed.

In Synthetic Monitoring, if a user is not assigned the global `datasources: read` role then the app won't display any data. This role is a required global permission that must be assigned to users in order for them to use the app (unless the Viewer role is assigned). I’ve updated the message (as shown in the screenshot) to clarify this so that it's not confused with the "app not initialized" state.

Note no `datasource:read` permission assigned:
image

Note the new error message indicating missing permissions:
image 


When the user is assigned the `datasources:read` permission, then the app can be used, even if no org role is assigned along the custom ones:

Note `datasources:read` is assigned:
image

Note the app loads correctly:
image

src/hooks/useUserPermissions.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    canReadThresholds: isUserActionAllowed('grafana-synthetic-monitoring-app.thresholds:read', OrgRole.Viewer),
+    canWriteThresholds: isUserActionAllowed('grafana-synthetic-monitoring-app.thresholds:write', OrgRole.Editor),
+
+    canReadTokens: isUserActionAllowed('grafana-synthetic-monitoring-app.access-tokens:read', OrgRole.Viewer),
+    canWriteTokens: isUserActionAllowed('grafana-synthetic-monitoring-app.access-tokens:write', OrgRole.Editor),
+    canDeleteTokens: isUserActionAllowed('grafana-synthetic-monitoring-app.tokens:delete', OrgRole.Editor),
+
+    canEnablePlugin: isUserActionAllowed('grafana-synthetic-monitoring-app.plugin:enable', OrgRole.Admin),
+    canDisablePlugin: isUserActionAllowed('grafana-synthetic-monitoring-app.plugin:disable', OrgRole.Admin),
```
w1kman, line 55: Nit: I guess this could also be a const 🤷🏻 
Assuming that `bootData` doesn't change without a reload.

src/hooks/useUserPermissions.ts
ckbedwell, line 1: This doesn't need to be a hook unless we are future-proofing for something in the future? I noticed most components seem to import `getUserPermissions` but we do have some that use the hook exported from it.

Should we commit to just using `getUserPermissions` and find a utility file for it to live in?

Alternatively, should we go the other way and consolidate `useDSPermission` into the `useUserPermissions` hook? Without knowing the underlying apis it could be confusing dx working out when you need to use `useUserPermissions` versus `useDSPermission`?

src/components/ProbeCard/ProbeCard.tsx
```diff
@@ -14,7 +14,7 @@ import { ProbeLabels } from './ProbeLabels';
 import { ProbeStatus } from './ProbeStatus';
 
 export const ProbeCard = ({ probe }: { probe: ExtendedProbe }) => {
-  const canEdit = useCanEditProbe(probe);
+  const { canWriteProbes } = useCanEditProbe(probe);
```
ckbedwell, line 17: Do we need to bring in `canReadProbes` for the alternative action? 🤔

src/test/utils.ts
```diff
@@ -149,6 +150,60 @@ export function runTestWithoutSMAccess() {
   });
 }
 
+export function runTestAsRbacReader() {
```
ckbedwell, line 153: Nit: can we uppercase all the RBAC in the function name (`runTestAsRBACReader`)? (and for the functions below).

src/types.ts
```diff
@@ -831,3 +831,24 @@ export interface CheckFormInvalidSubmissionEvent {
...
+  | 'grafana-synthetic-monitoring-app.alerts:write'
+  | 'grafana-synthetic-monitoring-app.alerts:delete'
+  | 'grafana-synthetic-monitoring-app.thresholds:read'
+  | 'grafana-synthetic-monitoring-app.thresholds:write'
+  | 'grafana-synthetic-monitoring-app.access-tokens:read'
+  | 'grafana-synthetic-monitoring-app.access-tokens:write'
+  | 'grafana-synthetic-monitoring-app.access-tokens:delete'
+  | 'grafana-synthetic-monitoring-app.plugin:enable'
+  | 'grafana-synthetic-monitoring-app.plugin:disable'
+  | 'grafana-synthetic-monitoring-app.tokens:delete';
```
w1kman, line 854: I think this could be a great contender for template literal types
Note: I'm not a TS Jedi, so there are probably better ways to do it than my examples.

Two examples, depending on how strict it needs to be
```
type PermissionBase = 'grafana-synthetic-monitoring-app';
type PermissionAction = 'read' | 'write' | 'delete' | 'enable' | 'disable';
type PermissionSubset = 'checks' | 'probes' | 'alerts' | 'thresholds' | 'access-tokens' | 'tokens' | 'plugin';
export type PluginPermissions =
  | `${PermissionBase}:${PermissionAction}`
  | `${PermissionBase}:${PermissionSubset}:${PermissionAction}`;
```

A bit stricter type
```
type PermissionBase = 'grafana-synthetic-monitoring-app';
type PluginPermissions =
  | `${PermissionBase}:${'read' | 'write'}`
  | `${PermissionBase}.checks:${'read' | 'write' | 'delete'}`
  | `${PermissionBase}.probes:${'read' | 'write' | 'delete'}`
  | `${PermissionBase}.alerts:${'read' | 'write' | 'delete'}`
  | `${PermissionBase}.thresholds:${'read' | 'write'}`
  | `${PermissionBase}.access-tokens:${'read' | 'write' | 'delete'}`
  | `${PermissionBase}.plugin:${'enable' | 'disable'}`
  | `${PermissionBase}.tokens:${'delete'}`;
```

Also, it feels like `enable|disable` are redundant (wouldn't `write|delete` be used instead).

src/components/ProbeEditor/ProbeEditor.tsx
ckbedwell, line 1: If the user doesn't have `canReadProbes` should they be able to see this page?

w1kman (changes requested):
Love the fix for a missing base role 🥳 
There are some nits that you may or may not wanna change.

src/page/AlertingPage.tsx
```diff
@@ -39,7 +39,7 @@ const Alerting = () => {
           Learn more about alerting for Synthetic Monitoring.
         </a>
       </p>
-      {canRead ? <AlertingPageContent /> : <InsufficientPermissions />}
+      {canReadAlerts ? <AlertingPageContent /> : <InsufficientPermissions />}
```
ckbedwell, line 42: I saw the insufficient permissions when I just had the basic 'Viewer' role (I didn't look into why). Should we provide more detail what permissions are missing as there are two places you would need to update to have the right permissions?

Screenshot of the Synthetic Monitoring Alerts page showing the insufficient permissions banner.

ckbedwell (changes requested):
I've done some testing but need to go through in detail to double-check. Wanted to get the initial review done with some feedback -- looks ace, found a few small things to take a look at.

src/hooks/useUserPermissions.ts
VikaCep: I moved the logic to a function in `data/permissions.ts`. I liked the idea of consolidating with `useDSPermission`, but in the end, I opted for this approach since, as you mentioned, it doesn't need to be a hook right now, and I didn't want to introduce a dependency on `PermissionsContext` for checking user permissions.

src/components/ProbeEditor/ProbeEditor.tsx
VikaCep: If the user doesn't have a `Viewer` role, without the `canReadProbes` action assigned to them, they actually _can't_ see this page: 

What they see:
image

user permissions in the screenshot:
image


This is enforced by setting the required `grafana-synthetic-monitoring-app.probes:read` action for the `/a/grafana-synthetic-monitoring-app/probes` path in the `includes` section of the `plugin.json` file. If a user tries to load any view under that page, they will be redirected to the home page.
https://github.com/grafana/synthetic-monitoring-app/blob/1ec89df9a78784c68f84d99df4b630fe052f65bf/src/plugin.json#L58

On the other hand, if the user does have the `Viewer` role assigned to them, then this `grafana-synthetic-monitoring-app.probes:read` action is automatically granted to them in the `grants` section of the role definition in `plugin.json` https://github.com/grafana/synthetic-monitoring-app/blob/1ec89df9a78784c68f84d99df4b630fe052f65bf/src/plugin.json#L122
This is convenient to prevent to have to perform any kind of migrations. Users with the `Viewer` role won't have to be assigned any other additional role in order to navigate the app.

src/components/ProbeCard/ProbeCard.tsx
VikaCep: Read restrictions are enforced in the `includes` section of the `plugin.json` [file](https://github.com/grafana/synthetic-monitoring-app/blob/1ec89df9a78784c68f84d99df4b630fe052f65bf/src/plugin.json#L38) where we set the required permissions to load the respective urls, so without those actions, the pages can't be accessed. See also https://github.com/grafana/synthetic-monitoring-app/pull/986#discussion_r1859223153

src/page/AlertingPage.tsx
VikaCep: Hmm I'm confused as I'm not sure how you got to that screen. 

Users with only the `Viewer` role assigned, should see this:
image
assigned roles:
image

Users without the `Viewer` role, but with the `Alerts reader` role, see the following:
image
assigned roles:
image


The difference lies in that `Viewer` assigns `fixed:alerting:reader` by default and without it it needs to be manually added.

src/hooks/useAlertPermissions.ts
```diff
@@ -0,0 +1,14 @@
+import { getUserPermissions } from 'data/permissions';
+
+import { useDSPermission } from './useDSPermission';
+
+export function useAlertPermissions() {
+  const { canReadAlerts, canWriteAlerts, canDeleteAlerts } = getUserPermissions();
+  const canEditAlertInDs = useDSPermission(`metrics`, `alert.instances.external:write`);
+
+  return {
+    canReadAlerts,
```
ckbedwell, line 10: I removed Query permissions for `Viewer` in my metrics datasource and I get faced with an infinitely loading spinner. This should be a simple case of checking if there is a value being returned for `useMetricsDS()` and adding it to the return for each of the `canReadAlerts` / `canWriteAlerts` / `canDeleteAlerts`.

The Alerts page. It shows the Alerts intro text follow by a loading spinner.

PR #996, opened 2024-11-22
chore: add basic config for `stylelint`

w1kman:
This PR introduces `stylelint` to help ensure CSS is written according to our standards and without unintentional errors.
We currently have a bunch of bad CSS (in template literals). To not depend on the built-in abilities of some IDEs, `stylelint` ensures that the quality of our styles is on par with the code that we expect to pass `eslint`.

Issues with CSS written as CSS (whether it be inside template literal or in .css files).

**Note: this does not address the weak "type safety" of Object style usage of `emotion/css`
**Note: I expect us to tweak and test `stylelint` before we (potentially) decide to merge it.


ckbedwell:
I think the config looks good for this. Shall we fix all the issues it flags, get it set-up in CI/CD and then merge it in or do you want to do those in separate PRs?

PR #999, opened 2024-11-25
fix: filter out disabled checks for execution count total

ckbedwell:
Addresses the FE portion of: https://github.com/grafana/support-escalations/issues/13498

## Problem

We aren't filtering out disabled checks when calculating a user's execution count.

## Solution

Add a filter to remove disabled checks when checking monthly execution counts.

### DIFF
--- a/src/hooks/useMonthlyTotalExecutionCount.test.ts
+++ b/src/hooks/useMonthlyTotalExecutionCount.test.ts
@@ -1,0 +1,74 @@
+import { renderHook, waitFor } from '@testing-library/react';
+import { getTotalChecksPerMonth } from 'checkUsageCalc';
+import { BASIC_PING_CHECK } from 'test/fixtures/checks';
+import { PRIVATE_PROBE } from 'test/fixtures/probes';
+import { apiRoute } from 'test/handlers';
+import { createWrapper } from 'test/render';
+import { server } from 'test/server';
+
+import { Check } from 'types';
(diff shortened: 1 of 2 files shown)


src/hooks/useMonthlyTotalExecutionCount.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      },
+      {
+        ...BASIC_PING_CHECK,
+        enabled: false,
+        frequency: FREQUENCY,
+        probes: PROBES,
+      },
+    ]);
+
+    expect(result.current.data).toBe(EXECUTIONS);
```
VikaCep, line 72: Are we certain `EXECUTIONS` contains the correct value here? Shouldn't we use a mocked value instead to validate it is returning the results we want in a more explicit way?

ckbedwell: I thought about hard-coding it but the problem is if we ever do update the calculation in `getChecksPerMonth` (which is likely as it doesn't match how backend calculate it) this test breaks for no real reason.

The focus of this test is checking whether the execution count matches what we would expect for the amount of enabled tests rather than checking if the precise execution count is correct. You helpfully added those tests so that's why I'm not concerned with this test doubling up and checking them, too 😃

VikaCep (approved):
💯🚀✅

PR #1002, opened 2024-11-26
Feat: prompt on unsaved unload

### LINKED ISSUE #892: Warn you are leaving check edit page when unsaved data is present
## Problem

When creating or editing a check you can spend a long time filling out all the fields, only to accidentally click a link, back/forward button etc. and immediately lose all of your work.

## Completion criteria

A confirmation that you want to leave the page without saving your work when you have 'dirty' form fields on the check creation page should be presented to the user.

Please ensure the appropriate integration tests have been added when submitting the PR.

w1kman:
To avoid users losing modified and tested changes, this PR adds "confirm on leave" for checks.
It also changes the behavior of the ~Submit~Save-button.

"Submit" is now called "Save" and is only enabled if there are unsaved changes to the check.

**Navigation handled by** `react-router-dom(-v5-compat)`
image

**Native navigation**
image

Resolves #892 

**Important: Please try to break it.** It would be very unsatisfying for users to get blocked from navigating because of a bug. 
`livereload` and `beforeunload` doesn't play nice. Get used to hitting `Enter` followed by `Escape`.


src/components/ConfirmLeavingPage/ConfirmUnsavedModal.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface ConfirmUnsavedModalProps {
+  onLeavePage: () => void;
+  onStayOnPage: () => void;
+}
+
+export function ConfirmUnsavedModal({ onLeavePage, onStayOnPage }: ConfirmUnsavedModalProps) {
+  return (
+    <Modal isOpen title="Unsaved changes" onDismiss={onStayOnPage}>
+      <h5 data-testid={DataTestIds.CONFIRM_UNSAVED_MODAL_HEADING}>
```
ckbedwell, line 14: I have no doubt you copied this from somewhere else so isn't on you but it is weird this is an h5. Can it just be a normal div or p tag and then be made to look the same as whatever the design system recommends?

src/page/NewCheck/__tests__/ApiEndPointChecks/CommonFields.payload.test.tsx
```diff
@@ -77,6 +77,7 @@ describe('Api endpoint checks - common fields payload', () => {
         });
       });
 
+      // TODO: Remove/keep? It's a dubplicate of the one in `4-alerting.payload.test.tsx`
```
ckbedwell, line 80: Keep this one and remove the multi step one.

src/components/ConfirmLeavingPage/ConfirmUnsavedModal.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('should call onStayOnPage when modal is dismissed (button)', async () => {
+    const onStayOnPage = jest.fn();
+    render(<ConfirmUnsavedModal onLeavePage={jest.fn()} onStayOnPage={onStayOnPage} />);
+
+    await screen.findByTestId('confirm-unsaved-modal-heading');
+    const closeButton = screen.getByRole('button', { name: 'Close' });
+
+    closeButton.click();
```
ckbedwell, line 40: We should use the `await user.click(closeButton)` returned from the render here.

src/components/ConfirmLeavingPage/ConfirmLeavingPage.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+const TEST_IDS = {
+  INITIAL_PAGE: 'ConfirmLeavingPage.initial-page',
+  LEAVE_PAGE_LINK: 'ConfirmLeavingPage.leave-page-link',
+  OTHER_PAGE: 'ConfirmLeavingPage.other-route',
+} as const;
+
+function Wrapper({ children }: PropsWithChildren<{}>) {
+  const history = locationService.getHistory();
+  // History will not automatically be reset between tests
+  history.replace('/');
```
ckbedwell, line 21: I'm worried this is easy to miss when maintaining this test in the future and it doesn't follow the convention of being in a `beforeEach` or similar where I'd expect to see this resetting behaviour.

Maybe remove this and add to the start of the `describe` block:

```
  beforeEach(() => {
    // History needs to be reset manually between tests as it uses Grafana's global locationService
    locationService.replace('/');
  });
```

ckbedwell (changes requested):
Minor change requests but this is pretty much good to go. 🙌 

I tried to break it in Chrome, Firefox and Safari but it endured. Solid as a rock 🪨🏅

src/page/NewCheck/__tests__/ApiEndPointChecks/CommonFields.payload.test.tsx
w1kman: I noticed a relevant difference `checkType`.
I just removed the comment.

ckbedwell (approved):
LGTM 🔥

ckbedwell (approved):
LGTM ⭐

VikaCep (approved):
LGTM!

PR #1006, opened 2024-11-26
Update scripted check default example/snippet

peterschretlen:

This PR updates the default scripted check example to add use of the `fail()` function, and explains why you would use it. 
 
Many people assume that if a `check()` does not pass,  the test run will fail and this will impact reachability and uptime.  They are unaware of the `fail()` function and why they should use it. 
 
This makes the default script longer, but it is still short enough that it fits on a typical laptop screen:

image


VikaCep (approved):
Nice!

PR #1011, opened 2024-12-06
Feat: Alerts per check

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring/issues/186

UI implementation for the new alerts per check. 

Some considerations:
- The new alerts are under a feature flag. Set `sm-alerts-per-check` to `true` to see them.
- In this PR https://github.com/grafana/synthetic-monitoring-app/pull/1023 I've introduced a new mocked API response which adds the ability to start the app using `yarn dev:msw`, and allows us to mock the requests both in the app and tests.

Screenshots:
image

- POST request
image

- GET request 
image


ka3de:
Thanks for making time to work on this @VikaCep ! :raised_hands: 
I will not make comments on the code because that's beyond my knowledge, but I have a couple of small comments on the UX, I would not consider these as blockers for this PR, but something that we can consider including in the long run.

The first one is that thresholds show no units. I think it could be helpful to specify the units for each threshold, for example whether the threshold refers to a `%`, or if it is supposed to be defined in `seconds (s)`, `milliseconds (ms)` or `days`, etc.

The second one we have already discussed it briefly, personally I think the alerts that support multiple percentiles look better grouped in a single "panel" as the previous iteration? In that case, though, we would need to figure a way to allow users to set different threshold values for the different percentiles, in case that more than one is set, as the same value would probably not make sense for all of them. If we finally opt for this solution and there is something that we have to change in the way the API behaves, let me know and we can work it out also.

src/data/useCheckAlerts.ts
```diff
@@ -0,0 +1,64 @@
...
+import { useSMDS } from 'hooks/useSMDS';
+export const queryKeys: Record<'listAlertsForCheck', QueryKey> = {
+  listAlertsForCheck: ['alertsForCheck'],
+};
+
+const alertsForCheckQuery = (api: SMDataSource, checkId?: number) => {
+  return {
+    queryKey: [queryKeys.listAlertsForCheck, checkId],
+    queryFn: () => api.listAlertsForCheck(checkId),
+    staleTime: 0,
```
ckbedwell, line 19: Why do we have this staleTime?

src/datasource/DataSource.ts
```diff
@@ -343,6 +344,25 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
...
 
+  //--------------------------------------------------------------------------------
+  // ALERTS PER CHECK
+  //--------------------------------------------------------------------------------
+
+  async listAlertsForCheck(checkId?: number) {
+    if (!checkId) {
+      const emptyResponse: CheckAlertsResponse = { alerts: [] };
+      return emptyResponse;
+    }
```
ckbedwell, line 355: Let's kill this. We can handle this problem upstream so the request never gets fired if there isn't a check id.

src/page/EditCheck/EditCheck.tsx
```diff
@@ -18,9 +19,16 @@ const EditCheckContent = () => {
   const { data: checks, isError, isLoading, error, refetch } = useChecks();
   const check = checks?.find((c) => c.id === Number(id));
 
+  const { data: alertsPerCheck } = useListAlertsForCheck(check?.id);
```
ckbedwell, line 22: We could leave this hook call here to increase perceived performance so we fetch the data before reaching the alerting step (maybe encapsulate it in another function with a clear name or leave a comment explaining why) but we don't have to pass down the response.

Just call `useListAlertsForCheck` from the `<AlertsPerCheck />` component because we are going to want the `isLoading` and `isError` states as well as probably the `refetch` method, etc. and render the correct UI accordingly.

When I tested this the dev environment wasn't deployed so naturally I got a 404 but was clueless as a user I had hit an error. We need to decide what we want to present in the UI if this endpoint isn't available in production.

src/components/CheckForm/AlertCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { control, formState, getValues } = useFormContext<CheckFormValues>();
+  const { isFormDisabled } = useCheckFormContext();
+
+  const thresholdError = formState.errors?.alerts?.[predefinedAlert.type]?.threshold?.message;
+
+  const isSelected: boolean = getValues(`alerts.${predefinedAlert.type}.isSelected`) || false;
+  const threshold: number = getValues(`alerts.${predefinedAlert.type}.threshold`) || 0;
+
+  return (
+    <Card
```
ckbedwell, line 24: I'll do a much more thorough review of the UI/UX separately but just a heads up I don't think we should use cards but probably a table of some sort.

ckbedwell (commented):
I need to do a much more thorough review of the UI/UX but I'd like to explore how we can do that by mocking the data in someway first.

I've focussed mainly on things I can see in the code first that will be agnostic to any of the other changes we'll discuss 😃

majavojnoska:
I agree with @ka3de that setting units will clarify what is being measured.

When it comes to the UI, my initial impression was that the cards functioned as radio buttons, allowing only one selection, due to the icon in the top right corner. However, after reviewing Daniel's Slack video, I understand users can set multiple alerts. 

A more elegant solution would be to replace the icons with a clear on/off indicator on the left side of each card. This would improve clarity. Also, I would suggest a more condensed card layout to reduce the prominence of the description. Let's discuss this further at our next meeting.

src/components/CheckForm/AlertsPerCheck.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { CheckAlertType, CheckFormValues, CheckStatus, CheckType } from 'types';
+import { CheckAlertsResponse } from 'datasource/responses.types';
+import { getAlertCheckFormValues } from 'components/CheckEditor/transformations/toFormValues.alerts';
+
+import { NewStatusBadge } from '../CheckEditor/FormComponents/CheckStatusInfo';
+import { AlertCard } from './AlertCard';
+interface PredefinedAlertInterface {
+  type: CheckAlertType;
+  description: string;
+  supportedCheckTypes?: CheckType[];
```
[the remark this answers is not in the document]
VikaCep: Good idea!

VikaCep:
Setting the PR to draft until we settle on the UI/UX changes and address them.

VikaCep:
After discussing with @majavojnoska and @ckbedwell, we agreed to implement the feature using a structure similar to the Probes selection screen, leveraging a list with checkboxes for selection. After applying the changes, the result now looks as follows:

**HTTP Check:**

https://github.com/user-attachments/assets/56fb373c-81b4-4421-b8dc-fd1eb0e1b9a8



**Ping Check:**
image


**All other checks:**
image

VikaCep:
Thanks for the feedback! I've humanized the labels and set a suffix attribute to the inputs where the unit is set. 

image

> Hey Vika. Two things I wanna try: remember we talked about making those labels easier to read? How about something like 'HTTP Request Duration Too High (P50)'?
> 
> Also, it might be cleaner to put the threshold value right next to the input field, like: '[value] ms/d/%'.

majavojnoska:
> Thanks for the feedback! I've humanized the labels and set a suffix attribute to the inputs where the unit is set.
> 
Great job, Vika!

ckbedwell:
I am going to dissent on the design opinion here a little bit just to make sure we have explored all the options available. I still think this should effectively be a table of some sort to reduce cognitive overhead. In the dev tools I changed a few values to see how it looked and I found this a lot easier to digest, especially if the section headings acted as accordions so I could reduce UI noise by closing the alert sections I wasn't interested in.

Per-check alerts presented as a table in rows and columns.

I understand the draw for consistency and referencing the probe UI but they are fundamentally different options with differing actions and a different mindset in identifying what is needed and how to manage them. With probes you can be faced with 30+ options and you want to quickly select 7 or 8 of them across their categorisations which mean the same thing. They are implicitly identical in what they provide, the UI is just presenting them so it is easier to digest their grouping -- it is usually identify the name, click once and carry on, not much else to think about.

With alerts, the workflow is more considered: it's not just a case of identifying what alert groupings are available and what they mean (because they all change dependant on the check type unlike probe regions) but then thinking carefully about two additional steps: which you want to enable AND what accompanying threshold you want to set. The cognitive model is different to enabling probes and sharing the same design adds to the confusion in my opinion.

I think it is likely we will be adding additional options and/or helper text to explain the alerts or provide examples to this in the future so a secondary concern is the existing design doesn't cater very well for extending the feature, it will increase clutter and cognitive overhead.

w1kman (changes requested):
image

**Red**
There are no sane default values when enabling an Alert. The second I enable an alert, I get an error.

**Yellow**
I would avoid using a magic number for wrapping (250px).
With `white-space: nowrap` on the labels, the wrapping is done in a more "natural" way.

src/data/useCheckAlerts.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { MutationProps } from './types';
+import { CheckAlertDraft } from 'types';
+import { FaroEvent } from 'faro';
+import { SMDataSource } from 'datasource/DataSource';
+import { CheckAlertsResponse } from 'datasource/responses.types';
+import { queryClient } from 'data/queryClient';
+import { useSMDS } from 'hooks/useSMDS';
+export const queryKeys: Record<'listAlertsForCheck', QueryKey> = {
+  listAlertsForCheck: ['alertsForCheck'],
```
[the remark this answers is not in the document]
VikaCep: Good catch 🙏

src/data/useCheckAlerts.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const alertsForCheckQuery = (api: SMDataSource, checkId?: number) => {
+  return {
+    enabled: Boolean(checkId),
+    //queryKey: [queryKeys.listAlertsForCheck, checkId],
+    // eslint-disable-next-line @tanstack/query/exhaustive-deps
```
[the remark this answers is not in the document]
VikaCep: We do, but right now it's simplified for mocks usage. I added a comment indicating to remove it when the API is deployed in dev.

VikaCep:
@ckbedwell I've applied your suggestion on the new layout, I like how it looks! 
image

@w1kman Thanks for the review, I've addressed your comments. Regarding the default values, I've asked @ka3de to provide me with realistic values that we can use as defaults for each alert.

src/components/CheckEditor/transformations/toFormValues.alerts.ts
```diff
@@ -0,0 +1,28 @@
...
+        id: undefined,
+        threshold: 0, // Default threshold
+        isSelected: false,
+      };
+    }
+
+    return acc;
+  }, {} as CheckAlertFormRecord);
+
+  return formAlerts;
```
w1kman, line 27: Redundant const `formAlerts`

```suggestion
return Object.keys(CheckAlertType).reduce((acc, alertTypeKey) => {
    const alertType = CheckAlertType[alertTypeKey as keyof typeof CheckAlertType];

    const existingAlert = data.alerts.find((alert) => alert.name.includes(alertType));

    if (existingAlert) {
      acc[alertType] = {
        id: existingAlert.id,
        threshold: existingAlert.threshold,
        isSelected: true,
      };
    } else {
      acc[alertType] = {
        id: undefined,
        threshold: 0, // Default threshold
        isSelected: false,
      };
    }

    return acc;
  }, {} as CheckAlertFormRecord);
```

src/components/CheckForm/AlertsPerCheck/AlertItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  id={`alert-threshold-${alert.type}`}
+                  value={field.value ? field.value : threshold}
+                  onChange={(e) => {
+                    return field.onChange(Number(e.currentTarget.value));
+                  }}
+                  width={10}
+                  disabled={!selected || isFormDisabled}
+                />
+              );
+            }}
```
w1kman, line 67: Nit: Would look cleaner with implicit return.

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.tsx
```diff
@@ -0,0 +1,138 @@
...
+              </Field>
+            )}
+          />
+        </Field>
+      </div>
+    </>
+  );
+};
+
+const headingDisplay = `h4`;
```
w1kman, line 116: Nit: should/could be scoped into the `getStyles` body.

src/components/CheckFormAlert.tsx
```diff
@@ -65,11 +66,18 @@ export const CheckFormAlert = () => {
   );
 };
 
+const headingDisplay = `h4`;
```
w1kman, line 69: Nit: should/could be scoped into the `getStyles` body

src/data/useCheckAlerts.ts
```diff
@@ -0,0 +1,66 @@
...
+import { queryClient } from 'data/queryClient';
+import { useSMDS } from 'hooks/useSMDS';
+export const queryKeys: Record<'listAlertsForCheck', QueryKey> = {
+  listAlertsForCheck: ['alertsForCheck'],
+};
+
+const alertsForCheckQuery = (api: SMDataSource, checkId?: number) => {
+  return {
+    enabled: Boolean(checkId),
+    // @todo add checkId to queryKey when the API is ready and mocks are removed.
```
w1kman, line 18: 👍🏻

src/faro.ts
```diff
@@ -19,6 +19,7 @@ export enum FaroEvent {
   SHOW_TERRAFORM_CONFIG = 'show_terraform_config',
   REFETCH_TENANT_LIMITS = 'refetch_tenant_limits',
   INITIALIZE_ACCESS_TOKEN = 'initialize_access_token',
+  UPDATE_ALERTS = 'update_alerts',
```
w1kman, line 22: Unless "update_check_alerts" was factually wrong, it seems more descriptive than just "update_alerts" (assuming that we're talking about updating check alerts and not alerts in general).

w1kman (changes requested):
Looks good! Just mostly some nit comments (that you can ignore if you don't agree with them). 

Seems like the change of the overflow behaviour caused the UI to always be "wrapped" and for the labels to become above/under (depending on the state of the input) the input.

image
image
image

ka3de:
> I've asked @ka3de to provide me with realistic values that we can use as defaults for each alert.

Ultimately this will depend on each service system and there might be considerable differences from system to system, but trying to come up with some sane defaults I think these could be valid values/ranges:

---

- ProbeFailedExecutionsTooHigh: **10%**

This equals the current _MedSensitivity_ [alert](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/configure-alerts/synthetic-monitoring-alerting/

---

- HTTPRequestDurationTooHighP50: **300** - 500 ms
- HTTPRequestDurationTooHighP90: **500** - 700 ms
- HTTPRequestDurationTooHighP95: **800** -1000 ms
- HTTPRequestDurationTooHighP99: **1500** - 2000 ms

Favoring lower values as there is a pending period of 5m.

---

- HTTPTargetCertificateCloseToExpiring: 30 - **60**

I'm favoring the "bigger" value here as depending on how the certificates are managed, for some "old-style" orgs this might even involve a decent level of bureaucracy and processes.

---

- PingICMPDurationTooHighP50: **50** - 70 ms
- PingICMPDurationTooHighP90: **100** - 150 ms
- PingICMPDurationTooHighP95: **200** - 250 ms
- PingICMPDurationTooHighP99: **400** - 500 ms

Favoring lower values as there is a pending period of 5m.

---

I would like to hear other opinions on this too.

majavojnoska:
> 
> Per-check alerts presented as a table in rows and columns.
> 
Excellent point, Chris! Thanks for bringing it up.

VikaCep:
Thanks @ka3de! Added the default values:

w1kman (approved):
Works mint now.

Nit: The label thingy is a bit off center though.
image

image

VikaCep:
Update with the header text centered

image

VikaCep:
I've removed the mocked data as now the API is available in dev 🚀

VikaCep:
I have made some updates based on @ckbedwell comment on [Slack](https://raintank-corp.slack.com/archives/C075N182HFW/p1737644041366949):

> **Shared (validate both sides?)**
> 1. I can submit values that should be invalid. I submitted 1000% here for Probe Failed Executions Too High and it got converted to 999.99 which is still too high by about 900 :sweat_smile:

> **UI/UX**
> 1. If I delete everything in the input field it resets to 0 and I can't delete the 0. If I want to type 10 it appears as 010 (if I select all then type I can get rid of the 0)
> 2. I found just visiting the alerts section changes the form state so even if I make no changes to my check I get the Unsaved changes modal
> 3. The double alert notification is confusing. I know they are separate APIs but we shouldn't render the alert update even if successful. Which brings us onto:
> 4. I like that if the alerts api fails we stay on the page but we have to make this experience nicer even if it should be very rare as it is very confusing what just happened. We need to be more explicit that the check did save but its associated alerts did not. Let's handle this problem in a separate PR but we need to address it.

- Added validation on percentage values, not allowing to go higher than 100%:
image

- Fixed the issue that wouldn't allow to delete the 0: 


- Removed the double alert notification, showing only the "check saved/updated" one:
https://github.com/user-attachments/assets/7db8709c-d82a-452d-a2b2-0368725b5a80

- Fixed the issue that showed "Unsaved changes" after the alerts loaded, even if there were no alert changes:
https://github.com/user-attachments/assets/fcbf0dec-271c-4cb2-acc5-b672aa4af043

w1kman (changes requested):
When using scientific notation, it's possible to save the form with a `0` as a value. 

E.g. Entering `1e-100` can be saved, which results in a `0`. If that scenario is triggered, the form step is still validated as an error (if you go to the next step) but without indication on what is the actual error.

image

If there is a step error (follow instructions above to trigger), it's not possible to empty the field. If the field says `0`, the `0` cannot be removed, only overwritten.

When in a broken state, there is no error message on entering an invalid  value. The value may also be "randomly" removed when switching step/section.

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.tsx
```diff
@@ -0,0 +1,142 @@
...
+      ),
+    [checkType]
+  );
+
+  if (isLoading) {
+    return <LoadingPlaceholder text="Loading alerts..." />;
+  }
+
+  if (isError) {
+    return <Alert title="There was an error fetching alerts for this check. Please try again or contact support." />;
```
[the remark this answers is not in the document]
VikaCep: By looking at other alerts within the app I saw we're linking to the contact form, so I added that link here too:
image

VikaCep:
> When using scientific notation, it's possible to save the form with a 0 as a value.

I added a validation to prevent users from entering numbers in scientific notation in the threshold inputs. I also added a min value of `0.01` as after discussing with @ka3de, he told me values with more than two decimals are disregarded in the database (so sending 0.001 would be converted to 0.00)

image
image

ckbedwell (changes requested):
Small one to avoid the unnecessary type assertion. This is pretty much ready to go so will approve after that ⭐

ckbedwell:
policy bot: approve

w1kman (approved):
LGTM 🥳 
Great work! 🥇

Issue #850, opened 2024-07-03
Add an object factory to create fixtures


ckbedwell:
# Problem

At the beginning of this year [I refactored the test data](https://github.com/grafana/synthetic-monitoring-app/pull/686) to be consistent across tests so there is a singular point of reference which all of our tests utilise. This is a great improvement but some tests require specific values to ensure they work correctly and often these values are unclear and opaque in nature.

# Completion criteria

- [ ] Introduce [Fishery](https://github.com/thoughtbot/fishery) as an object factory to generate our test fixtures.
- [ ] Where appropriate change some of our tests to utilise Fishery rather than using the global fixtures (e.g. `BulkEditModal.test.tsx`) -- _note: it isn't a requirement to change all of our tests to utilise Fishery to consider this ticket complete but the goal is to get the ball rolling._
- [ ] Ensure your solution allows us to utilise the object factory OUTSIDE of tests easily. 

On the last point: one of the benefits of utilising Fishery is for us to mock data whilst developing locally. An example where this could be applied -- currently our call to the checks list api is not paginated and returns _all_ checks. Some of our users have 1000s of checks and it would be a very interesting experiment to mock these 1000 of checks in the UI and see what happens and aid in developing solutions.

PR #1019, opened 2024-12-16
Chore: introduce mswjs/data to facilitate creating fixtures

VikaCep:
**WIP**

Ref: [grafana/synthetic-monitoring-app#850](https://github.com/grafana/synthetic-monitoring-app/issues/850)

This PR introduces the use of [mswjs/data](https://github.com/mswjs/data) to simplify mocking data in tests, eliminating the need to hard-code objects. It's paired with [Faker](https://github.com/faker-js/faker) to automatically generate test data for more dynamic and flexible testing scenarios.


src/components/ProbeEditor/ProbeEditor.test.tsx
```diff
@@ -90,11 +90,18 @@ it('saves new probe', async () => {
...
-    labels: [...probe.labels, ...UPDATED_VALUES.labels],
-  });
+
+  const updated = JSON.parse(
+    JSON.stringify({
+      ...probe,
+      ...UPDATED_VALUES,
+      labels: [...probe.labels, ...UPDATED_VALUES.labels],
+    })
+  );
```
ckbedwell, line 100: Can you add some detail to why we have to do this?

src/test/db.ts
```diff
@@ -0,0 +1,303 @@
...
+
+  const cleanedObject = keys.reduce((acc, key) => {
+    if (typeof key !== 'symbol') {
+      acc[key as keyof T] = obj[key as keyof T];
+    }
+    return acc;
+  }, {} as T);
+
+  return cleanedObject;
+}
```
ckbedwell, line 303: And can you explain what this is doing and why it is needed?

VikaCep: I introduced this `clean` function because the objects created by the mswjs factory contain `Symbol` properties (such as `Symbol(primaryKey): "id"`), which cause mismatches when comparing object equality in tests. This results in failed tests due to these hidden Symbol properties not matching between expected and actual objects. The `clean` function removes these Symbol properties to ensure objects can be compared correctly.

Without this function, tests fail with error messages like the one below:

image

Ideally, I'd prefer to extend the `db.create` function so that we don't have to manually call `clean` or rely on workarounds like this. I'm still exploring the best way to address this issue at the source.

VikaCep:
I also tried using Fishery in another [PR](https://github.com/grafana/synthetic-monitoring-app/pull/1023), and it turned out to be much better in terms of typing. Specifically, I no longer had to cast nested objects as `any`, as is the case here. Also, Fishery eliminates the need to clean generated objects of their `Symbol` properties as they're not added.

I don’t see significant advantages in keeping `mswjs/data` over `Fishery` based on this experience. So, I’ll close this PR in favor of the other one.

PR #1021, opened 2024-12-18
refactor: move ChooseCheckGroup to page folder

ckbedwell:
Part 3 of addressing: https://github.com/grafana/synthetic-monitoring-app/issues/848

## Problem

`ChooseCheckGroup` was in the components folder despite it being the primary page for choosing a check group in check creation.

## Solution

Move `ChooseCheckGroup` to the page folder and its associated components. I've moved the component it utilizes appropriately to either the localised page component folder or flatten it to the global components folder. 

I've also separated `NewStatusBadge` to its own file rather than being another export from `CheckStatusInfo`. Our component files should only have one export to one day support hot-reloading properly 🤞


VikaCep (approved):
💯 🚀 👏

PR #1023, opened 2024-12-19
Chore: introduce Fishery as object factory and yarn dev:msw

VikaCep:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/850

This PR introduces the following changes:

**Fishery:** Simplifies mocking data in tests, removing the need for hard-coded objects. It works alongside [Faker](https://github.com/faker-js/faker) to automatically generate dynamic test data for more flexible testing scenarios.

It defines factories for:

- All different check types
- Probes
- Check alerts

These factories are used to create fixtures in `src/test/fixtures/probes.ts`, `src/test/fixtures/checks.ts`, and `src/test/fixtures/checkAlerts.ts`.

**Mocked API responses:** Adds the ability to start the app with mocked API responses. To use this, run `yarn dev:msw`. This runs a service worker that uses the same handlers as in test . 

Example of running the app using MSW mocked responses:

https://github.com/user-attachments/assets/16e0778d-073d-454b-9d94-6cead59f007d


src/test/db.ts
```diff
@@ -0,0 +1,236 @@
...
+  Probe,
+  ScriptedCheck,
+  TCPCheck,
+  TCPQueryResponse,
+  TracerouteCheck,
+} from 'types';
+import { Assertion, MultiHttpVariable } from 'components/MultiHttp/MultiHttpTypes';
+
+const baseCheckModel = () => ({
+  id: faker.number.int({ min: 1, max: 999999 }),
```
ckbedwell, line 32: Despite having the range very large, this is going to a potential source of flakiness in tests when IDs collide -- especially if we start creating larger data sets to test with.

Fishery provides a [sequence](https://github.com/thoughtbot/fishery?tab=readme-ov-file#define-and-use-factories) param that you can use as an ID to combat this problem but you'll have to change the factory to be a universal checks factory with a switch clause to the other factories.

src/test/db.ts
```diff
@@ -0,0 +1,236 @@
...
+    modified: Math.floor(faker.date.recent().getTime() / 1000),
+    created: Math.floor(faker.date.past().getTime() / 1000),
+    capabilities: {
+      disableScriptedChecks: faker.datatype.boolean(),
+      disableBrowserChecks: faker.datatype.boolean(),
+    },
+  })),
+
+  alert: Factory.define<CheckAlertPublished>(() => ({
+    id: faker.number.int({ min: 1, max: 999999 }),
```
ckbedwell, line 230: Same case here (and probes) to use sequence instead.

src/test/fixtures/checkAlerts.ts
```diff
@@ -1,35 +1,25 @@
+import { db } from 'test/db';
+
 import { CheckAlertType } from 'types';
 import { CheckAlertsResponse } from 'datasource/responses.types';
 
 export const BASIC_CHECK_ALERTS: CheckAlertsResponse = {
   alerts: [
-    {
+    db.alert.build({
```
ckbedwell, line 8: If you offload the id to the Factory creating a sequence you could change this to something more succinct:

```
[
  CheckAlertType.HTTPRequestDurationTooHighP90,
  CheckAlertType.HTTPRequestDurationTooHighP95,
  CheckAlertType.HTTPTargetCertificateCloseToExpiring,
  CheckAlertType.ProbeFailedExecutionsTooHigh,
].map(name => db.alert.build({ name }))
```

src/test/fixtures/checks.ts
```diff
@@ -72,118 +70,46 @@ SBefoVnBNp449CSHW+brvPEyKD3D5CVpTIDfu2y8+nHszfBL22wuO4T+oem5h55A
...
 const transformedValidKey = btoa(VALID_KEY);
 
-export const BASIC_DNS_CHECK: DNSCheck = {
+export const BASIC_DNS_CHECK: DNSCheck = db.dnsCheck.build({
   id: 1,
   job: 'Job name for dns',
   target: 'dns.com',
-  enabled: true,
   labels: [{ name: 'dnsLabelName', value: 'dnsLabelValue' }] as Label[],
   probes: [PRIVATE_PROBE.id, PUBLIC_PROBE.id] as number[],
```
ckbedwell, line 78: Do we need to add any of these params if the Factory is making them all? (same with all the fixtures)

src/test/fixtures/probes.ts
```diff
@@ -1,85 +1,59 @@
+import { db } from 'test/db';
+
 import { Label, Probe } from 'types';
 
-export const PRIVATE_PROBE: Probe = {
-  name: 'tacos',
+export const PRIVATE_PROBE: Probe = db.probe.build({
   id: 1,
+  name: 'tacos',
   public: false,
```
ckbedwell, line 8: This seems like the only param we need as the factory is generating the params -- alternatively, or do we make a subclass to the probe factories so we have a PublicProbe factory and a PrivateProbe factory?

src/test/fixtures/probes.ts
```diff
@@ -1,85 +1,59 @@
...
-} as const satisfies Probe;
+});
 
-export const OFFLINE_PROBE: Probe = {
-  ...PRIVATE_PROBE,
+export const OFFLINE_PROBE: Probe = db.probe.build({
+  name: 'quesadilla',
+  id: 4,
+  public: false,
   online: false,
```
ckbedwell, line 39: Same with these other fixtures, the only discerning param this probe needs is to be offline. Do we need the other params?

ckbedwell (changes requested):
This is ace but there are some things we can tighten up! 🔧

src/test/db.ts
```diff
@@ -0,0 +1,236 @@
...
+        tls: faker.helpers.maybe(() => faker.datatype.boolean()),
+        tlsConfig: faker.helpers.maybe(() => tlsConfig()),
+      },
+    },
+  })),
+
+  probe: Factory.define<Probe>(() => ({
+    id: faker.number.int({ min: 1, max: 999999 }),
+    name: faker.string.uuid(),
+    public: faker.datatype.boolean(),
```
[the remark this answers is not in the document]
ckbedwell: As per my comment below on reflecting upon this, I think we should have two factories. One that creates public probes and one that creates private probes.

src/test/fixtures/probes.ts
VikaCep: Removed all unneeded params 👍

src/test/db.ts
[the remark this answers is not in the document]
VikaCep: I tried creating two separate factories—one for PrivateProbes and one for PublicProbes—but since the IDs for public and private probes can’t overlap, we need to manage their creation within a single factory. This requires using a transient parameter, similar to how it’s done with checks. But having two separate factories was complicating things, so I found it more straightforward to just use one factory that creates private probes by default. For fixtures that require a public probe, we can simply set the `public: true` property.

src/module.ts
```diff
@@ -27,6 +29,10 @@ if (window.location.hostname !== 'localhost') {
   });
 }
 
+if (process.env.NODE_ENV === 'development' && process.env.REACT_APP_MSW) {
+  setupWorker(...handlers).start();
+}
```
ckbedwell, line 34: Just needs a small modification so it gets tree-shaken properly and doesn't affect the production bundle size:

1. Move line 33 and associated imports to a file like `startServerWorker.ts`:

```
import { setupWorker } from 'msw';
import { handlers } from 'test/handlers';

setupWorker(...handlers).start();
```

2. change line 33 to: `await import(`./startServiceWorker`);`

Then we don't include the additional 850kb overhead our mocks have 😄 

Comparing the webpack console output before and after making this change. Before 2.87MiB. After 2.02MiB

ckbedwell (changes requested):
One thing to change then let's get this in! 🎉

ckbedwell (dismissed):
LGTM ❤️

Issue #1024, opened 2024-12-19
Unable to Save check when using Test-button (adhoc test)


w1kman:
**What happened**: When a Check is modified followed by clicking on the "Test" button, it's no longer possible to press the "Save" button.

**What you expected to happen**: If there are unsaved changes, the "Save" button should be enabled.

**How to reproduce it (as minimally and precisely as possible)**:
Modify a Check, click the "Test"-button on the "Execution" section, see how the "Save" button gets disabled.

**Anything else we need to know?**:

**Environment**:
- SM plugin version: Plugin version: 1.16.10

w1kman:
This has been resolved since early January.

PR #1025, opened 2024-12-19
fix: disable hasUnsavedChanges prop for now

w1kman:
**What this PR does / why we need it**: When using the "Test" button in the CheckForm, the form is technically submitted but not saved. After testing, the formState is no longer "dirty", which results in the "Save" button being disabled and the "You have unsaved changes" dialog to not appear.

This fix removes the logic that disables the "Save" button when the formState is clean.
This fix DOES NOT resolve the fact that a user can navigate away from an unsaved check without getting prompted. 

Resolves https://github.com/grafana/synthetic-monitoring-app/issues/1024

### DIFF
--- a/src/components/CheckForm/CheckForm.tsx
+++ b/src/components/CheckForm/CheckForm.tsx
@@ -191,7 +191,6 @@ export const CheckForm = ({ check, disabled }: CheckFormProps) => {
               onValid={handleValid}
               onInvalid={handleInvalid}
               schema={schema}
-              hasUnsavedChanges={hasUnsavedChanges}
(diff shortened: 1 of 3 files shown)


VikaCep (approved):
Some tests need to be updated but this solution LGTM until we change the logic of the Test button not acting as a submit button.

w1kman:
Changed to draft as we did a rollback on `main`.

ckbedwell:
Closing this as it is now solved in: https://github.com/grafana/synthetic-monitoring-app/pull/1037

Issue #1029, opened 2024-12-20
Browser Dashboard doesn't apply probe filter to Assertions, Data sent/recieved.


d0ugal:
**What happened**:

Browser dashboard is not applying probe filters to assertions, data sent and data recieved

**What you expected to happen**:

The panels should filter and only show results from the selected probes, but all data is currently shown.

**How to reproduce it (as minimally and precisely as possible)**:

1. View browser check dashboard (with multiple probe locations)
2. Select probe in the top left filter
3. Compare assertions, data sent and data received panels and compare without filtering the probes.

![Image](https://github.com/user-attachments/assets/33abbf1c-4860-4a1c-826d-bdf3e691640b)

PR #1030, opened 2024-12-20
fix: filter by probe on data transferred and assertions panels for browser and scripted checks

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1029

The data transferred and assertions panels in the Browser and Scripted dashboards aren't taking into account filtered probes. This PR fixes that. 

Since in dev we only have one running probe location, I've validated the queries work as expected using Explorer in my prod account:

**Before setting the probe filter (as it currently works in the dashboards):**

Data sent chart:
image

Data received chart:
image

Success rate chart:
image

**After setting the probe filter (after this PR implementation):**

Data sent chart:
image

Data received chart:
image

Success rate chart:
image

### DIFF
--- a/src/scenes/BROWSER/WebVitals/cumulativeLayoutShift.ts
+++ b/src/scenes/BROWSER/WebVitals/cumulativeLayoutShift.ts
@@ -9,7 +9,7 @@ function getQueryRunner(metrics: DataSourceRef) {
     queries: [
       {
         refId: 'A',
-        expr: `avg by (job, instance) (quantile_over_time(0.75, probe_browser_web_vital_cls{instance="$instance", job="$job"}[$__range]))`,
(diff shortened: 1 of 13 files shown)


ckbedwell:
This is great but I've noticed there are some other panels that also aren't taking into account the probe filter so think we should add the probe filter to them, too:

**Scripted**
Uptime
Assertions
Results by URL


_Scripted dashboard with a non-existent probe as the filter. Some panels are still showing data._

**Browser**
Uptime
FCP
LCP
TTFB
CLS
FID
INP
Metrics by URL
Assertions

_Browser dashboard with a non-existent probe as the filter. Some panels are still showing data._

VikaCep:
Great catch, I've updated all panels accordingly to take the probe into account: 

Scripted checks dashboard filtering by one probe location:

image
image
image

Browser checks dashboard filtering by one probe location:
image
image
image

ckbedwell (approved):
LGTM 💪

Issue #1026, opened 2024-12-19
Do not submit Check form when starting adhoc test of check


w1kman:
**What would you like to be added**:
Currently, the "Test"-button ("execution" section) will submit the Check form, rendering the formState to be (technically) saved, when in fact the check has not been saved.
Any logic that revolves around wether or not the check has been saved or not stops working the second the "Test"-button is clicked.

I propose that the ad-hoc check is performed without submitting the form, which would let us leverage the built-in functionality of `react-form-hook`. Submitting the form equals saving the form data (saving the check).

The "Test"-button could/should include all the required logic to POST the form data (without submitting the actual form) to the ad-hoc check as well as render the result modal.

Affected features
- Disable "Save" button when there are no modifications to the Check (or modifications are reset)
- Prompt when leaving unsaved form

**Why is this needed**:

**Additional context**
see https://github.com/grafana/synthetic-monitoring-app/issues/1024

PR #1037, opened 2025-01-06
fix: change dirty strategy in check form

ckbedwell:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/1026#event-15826703952

## Problem

When introducing the notify unsaved changes feature, an accidental bug was introduced where after using the test button the form was considered submitted and could no longer be saved.

## Solution

The solution is to change the strategy in assessing whether the form is 'dirty' or not. Instead of utilizing React hook form's `isDirty` value in provided by the `formState`, we are manually checking if the initial values match the current values. It is not the most optimal solution but it does work adequately. 

The desired solution proposed in the issue of not considering the test button a submit button has its own tradeoffs with the way `react-hook-form` works and the current set-up of the Synthetic Monitoring application. Because the `FormLayout` component doesn't expose any of its inner workings (such as moving between steps and the validation invalid / valid triggers provided by `react-hook-form`'s `handleSubmit`) it would require a larger re-work of these components and have the same UX.

The `test` button is likely to evolve when we tackle phase 2 of the check creation flow so this solution is most appropriate right now.

Last note, it is difficult to write tests for the 'test' button because JSDom doesn't expose the submitter element, which we rely upon to know if it is an ad-hoc test or saving the check.


VikaCep:
There is still a strange behavior going on. When there are changes, the form correctly enables the Save button. But when clicking save, the message indicating there are unsaved changes is prompted asking to stay/leave the page. It still saves but it's confusing as there's no clear indication if the save is successful.

https://github.com/user-attachments/assets/47b26049-a561-49d8-97c0-722800f6b82a

VikaCep (changes requested):
https://github.com/grafana/synthetic-monitoring-app/pull/1037#issuecomment-2573515581

ckbedwell:
> There is still a strange behavior going on. When there are changes, the form correctly enables the Save button. But when clicking save, the message indicating there are unsaved changes is prompted asking to stay/leave the page. It still saves but it's confusing as there's no clear indication if the save is successful.
> 
>  Screen.Recording.2025-01-06.at.13.51.33.mov

I'm glad you had the common sense to check submitting it (because I didn't) 😅

I've made the changes and added a few tests as appropriate. I found a few test files I found redundant (the submit ones for the 2 api types as they test the same thing and aren't specific to those tests) so I've removed them and put what they are testing in a more appropriate place.

VikaCep (approved):
LGTM! 🚀

PR #1039, opened 2025-01-13
fix: keep reference of submitting form internally

ckbedwell:
https://github.com/grafana/synthetic-monitoring-app/pull/1037

Fix the actual original bug that the above PR was meant to solve ☝️ 🤦


w1kman (approved):
Seems to work as expected 👌🏻 
- When using Test button (ad-hoc check)
- On error

VikaCep (approved):
💯 👏 ✅

Issue #1043, opened 2025-01-20
New public probes don't show CSP


The-9880:
Looking at the new public probes, I notice they're missing the CSP annotation (AWS):
![Image](https://github.com/user-attachments/assets/33053259-d269-4587-a3c9-4e1a77742832)

After the last shutdown on March 17th (ref: [What's New](https://grafana.com/docs/grafana-cloud/whats-new/2025-01-14-launch-and-shutdown-dates-for-synthetics-probes-in-february-2025/)), all the public probes will be on AWS. We _might_ be able to omit the CSP annotation altogether after that point, but until then it may be good to include it for the recent additions.

PR #1044, opened 2025-01-20
fix: add new probes metadata

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1043

Adds metadata for newly added probes.


ckbedwell (approved):
LGTM ⭐ 

This is a separate issue but when I stubbed the data in dev to spot check it, I noticed that the UI looks a bit squiffy because of the longer lines with `SanFrancisco, US (Digital Ocean) [icon]` and `Bangalore, IN (Digital Ocean) [icon]`.


The probe region selections aren't all in a row and APAC gets pushed below AMER and EMEA. I reckon we could probably bump the maximum height up too.

ckbedwell (approved):
LGTM 💪

PR #1045, opened 2025-01-20
fix: wrap labels to always display View dashboard button

### LINKED ISSUE #1042: View Dashboard button not visible in checks with large amount of labels
The "View dashboard" button is not visible in the Checks overview screen on smaller resolutions when there is a large number of labels.

Reported by Delivery Hero during their migration from DD.

**What happened**:

Image

View dashboard button not visible

I would expect the View dashboard button to be visible/accessible at all times irregardless on the number of labels or the resolution 

1. Create a Synthetic Monitoring check with a larger number of labels (more than 5)
2. open the Checks overivew page on a smaller browser window

VikaCep:
Closes #1042 

With a large number of labels, the View Dashboard, Edit and Delete buttons get overlapped. This PR fixes it by wrapping them.

https://github.com/user-attachments/assets/ba9afe55-6a18-4209-bf74-a2d4ae102537

### DIFF
--- a/src/page/CheckList/components/CheckListItemCard.tsx
+++ b/src/page/CheckList/components/CheckListItemCard.tsx
@@ -95,7 +95,6 @@ const getStyles = (theme: GrafanaTheme2) => ({
     display: 'flex',
     flexWrap: 'wrap',
     flexGrow: 1,
-    maxWidth: '90%',
     gap: theme.spacing(1),
   }),
   heading: css({
--- a/src/page/CheckList/components/CheckListItemCard.tsx
+++ b/src/page/CheckList/components/CheckListItemCard.tsx
@@ -73,7 +73,7 @@ export const CheckListItemCard = ({
             </div>
           </div>
           <Stack>
-            <Stack grow={1} alignItems="center">
+            <Stack grow={1} alignItems="center" wrap={'wrap'}>
               {check.labels.map((label: Label, index) => (
                 <CheckCardLabel key={index} label={label} onLabelSelect={onLabelSelect} />
               ))}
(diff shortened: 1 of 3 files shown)


w1kman (changes requested):
If the idea is to always show "view", "edit" and "delete", this PR doesn't solve that problem.
At 880px width, the edit button and the delete button are not visible.

image

At 729px, none of the buttons/links are visible
image

VikaCep:
Good catch, I've pushed an update:

https://github.com/user-attachments/assets/1e299a78-91db-49a0-947b-0d18ba55f706

src/page/CheckList/components/CheckListItemCard.tsx
```diff
@@ -91,6 +91,13 @@ const getStyles = (theme: GrafanaTheme2) => ({
     backgroundColor: theme.colors.background.secondary,
     borderRadius: '2px',
   }),
+  labelsContainer: css({
+    display: 'flex',
+    flexWrap: 'wrap',
+    flexGrow: 1,
+    maxWidth: '90%',
```
w1kman, line 98: Why not 100%? Looks way better.
image

vs 90%
image

src/page/CheckList/components/CheckListItemCard.tsx
```diff
@@ -72,12 +72,12 @@ export const CheckListItemCard = ({
               )}
             </div>
           </div>
-          <Stack>
-            <Stack grow={1} alignItems="center">
+          <Stack wrap={'wrap'} justifyContent={'flex-start'}>
```
w1kman, line 75:
```suggestion
          <Stack wrap="wrap" justifyContent="flex-start">
```

w1kman (changes requested):
Looks much better!
I don't understand the `max-width: 90%`, 100% seems to look much better.
I open for the fact that I can be missing something obvious.

w1kman (approved):
LGTM!

PR #1049, opened 2025-01-24
Feat: increase check timeout to scripted, browser and multihttp checks

### LINKED ISSUE #1048: UI support for increased MultiHTTP, scripted and browser timeouts
## Problem

[We are increasing the allowed timeout in scripted and browser checks](https://github.com/grafana/synthetic-monitoring/issues/117) to better support longer running checks. Currently the UI only allows these type of checks to timeout after up to a minute.

## Acceptance criteria

_(~~confirm~~ **confirmed** with @d0ugal)_  -- all the below does also apply to MultiHTTP checks, as well.

1. Increase the allowed timeout for checks to 90 seconds. Please note that this will need to be incrementally bumped in PRs at a later date to match the rollout plan as [discussed in this document](https://docs.google.com/document/d/1jiBjN7L18TNHYVxnZhiiPIE_pV8I6kXmicYq--9HSpc/edit?tab=t.0)
2. Add validation that the frequency can not be less than the allowed timeout. The minimum frequency for a browser / scripted check is 1 minute but as we will be increasing the timeout above this, we need to ensure that checks have the opportunity to cross-over as little as possible. 
   - Display an error message beneath the frequency selector explaining it can not be less than timeout. They must adjust one or the other.
   - Do **not** automatically adjust the allowed frequency based on their timeout selection. We have to be explicit to the user in this scenario it is not allowed. If we adjust it for them it will be a source of confusion for the user that their desired frequency is not being honoured.
   - Ensure that the wizard form moves to step 5, displaying the error when the form is submitted from another step.
   - _(~~confirm~~ **confirmed** with @peterschretlen )_ -  Currently we allow timeout to be greater than frequency in the UI but the api rejects it. Please ensure endpoint checks are included in the validation.
3.  Please ensure all the above has the correct integration tests added.

VikaCep:
Closes #1048 

- Allows a timeout of 90 seconds for `browser`, `scripted` and `multihttp` checks
- Adds validation to prevent frequency lower than the chosen timeout
- Adds tests

https://github.com/user-attachments/assets/bc15e366-6940-406c-b366-343a3bed43dc

### DIFF
--- a/src/components/CheckForm/CheckForm.constants.tsx
+++ b/src/components/CheckForm/CheckForm.constants.tsx
@@ -1,0 +1,15 @@
+import { CheckType } from 'types';
+
+const CheckTimeoutValues = {
+  [CheckType.PING]: { min: 1, max: 60 },
+  [CheckType.TCP]: { min: 1, max: 60 },
+  [CheckType.Traceroute]: { min: 30, max: 30 },
+  [CheckType.DNS]: { min: 1, max: 60 },
+  [CheckType.GRPC]: { min: 1, max: 60 },
+  [CheckType.HTTP]: { min: 1, max: 60 },
+  [CheckType.MULTI_HTTP]: { min: 5, max: 90 },
+  [CheckType.Scripted]: { min: 5, max: 90 },
+  [CheckType.Browser]: { min: 5, max: 90 },
+};
+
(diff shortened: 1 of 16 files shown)


ckbedwell (approved):
Great work, works perfectly 🙏

PR #1054, opened 2025-01-30
Feat/uptime query v3

ckbedwell:
## Problem

Currently our uptime query is not only incredibly opaque but also inaccurate in a variety of situations. We need to document how our uptime query is handled for both internal and external users, ensuring the reasoning, explanations and reliability of the query are as clear as possible.

Previously we implemented a [v2 of the uptime query](https://github.com/grafana/synthetic-monitoring-app/pull/840) behind a feature flag after an escalation but we never enabled it for all stacks. We didn't have a sufficiently confident testing strategy when releasing it to our customers and we were unsure if we would be trading one bug for another set.

## Solution

This PR adds an uptime v3 to **replace** all uptime queries. It is not behind a feature flag and it removes the query and feature flag for v2. The query has been tested extensively in a variety of situations and has come out looking very accurate. **Please thoroughly read the [uptime.md](https://github.com/grafana/synthetic-monitoring-app/blob/feat/uptime-query-v3/docs/uptime/uptime.md) internal documentation that this PR adds.**

_I am in the middle of documenting how to set up and use the Prometheus test data library, which is how I settled on uptime v3 and the assessments of how accurately it performs. I will add this to the documentation once it is ready._ 

This PR also addresses some tangentially related UX inconsistencies when assessing uptime across the application:

1. All areas of the application displaying uptime now use the **same query**.
2. All scenes dashboards default to showing the last 3 hours (the same as the check list page)

Both of these changes combined mean that uptime should remain consistent and not unexpectedly 'jump' around as you are navigating from one route to another.

This PR does **not** intend to address reachability or any other metric panels on our dashboards (that will come later 😄).


mem (approved):

docs/uptime/uptime.md
```diff
@@ -0,0 +1,289 @@
...
+| 1 year     | 15 seconds      | 20               | 41,932,800  |
+
+3. A check might have multiple configurations during the time period we are querying. This could be a change in frequency, the number of probes or the check's definition of uptime itself. This could lead to a change in the number of time points and the number of them which are successful.
+
+## How we calculate uptime with PromQL
+
+Taking into account all of the above, our uptime calculation is the following:
+
+```
+clamp_max(sum(max_over_time(probe_success{job="$job", instance="$instance"}[$frequencyInSeconds])), 1)
```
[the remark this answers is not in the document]
mem: > The first option bumps into Mimir's (Grafana Cloud's) limits for how big a range can be for this kind of computation (90 days, IIRC). The second one will bump against how many samples Mimir is willing to return for a single query (somewhere in the 10000 range, IIRC).

The limit is expressed in terms of hours. It's 768 hours, which corresponds to 32 days. Either there's some padding in there to allow a 30-day query to work, or there's something about 768 that I'm missing.

Playing around with Grafana, I can see that if I try to do something like:

```
max by (instance, job) (probe_success{instance=..., job=...})
```

and specify a range query for 90 days, it will happily return some results, but they are 1 hour apart (1 * 24 * 90 = 2160 samples).

If try with an instant query like:

```
max by (instance, job) (probe_success{instance="...", job="..."})[30d:2m]
```

it will return all the results (30 * 24 * 30 = 21600)

This on the other hand returns an error:

```
max by (instance, job) (probe_success{instance="...", job="..."})[90d:2m]
```

```
execution: the query time range exceeds the limit (query length: 2160h3m45.551s, limit: 768h0m0s) (err-mimir-max-query-length). To adjust the related per-tenant limit, configure -querier.max-partial-query-length, or contact your service administrator.
```

Even something like:

```
max by (instance, job) (probe_success{instance="...", job="..."})[90d:1h]
```

which is equivalent to what Grafana is doing, does not work.

docs/uptime/uptime.md
```diff
@@ -0,0 +1,289 @@
...
+3. **"Is our hosting provider reliable?"** - To assess whether more strategic decisions are warranted and to paint a holistic picture, it is equally valid to query much longer time periods, such as a quarter or even a year.
+4. **"We had an incident. I need to know exactly what happened."** - a query of just a few minutes is very likely in the case of an incident that requires a deep level of analysis when problems began and ended so the root cause can be established.
+
+When we consider all of these use cases, we have to cater for an accurate uptime metric for as little as a minute up to one year. This is a **huge** range of time periods and we need to be able to accurately calculate uptime for all of them.
+
+## How we collect and query data
+
+![](./images/sm-architecture.png)
+
+[A more detailed explanation of the Synthetic Monitoring architecture in our on-call training.](https://docs.google.com/presentation/d/1jWuv2VwVFIsdyXq4JQMlhmM6txHXLyHImYpnl50V5XQ/edit#slide=id.g308ff111cb5_0_35).
```
peterschretlen, line 22: Flagging that this is an internal-only link in a public repo, so the general public may click thinking they can access.  It might be worth adding "(internal-only)" or something to set expectations this won't be externally visible. 
```suggestion
[A more detailed explanation of the Synthetic Monitoring architecture in our on-call training (Grafana Labs internal-only).](https://docs.google.com/presentation/d/1jWuv2VwVFIsdyXq4JQMlhmM6txHXLyHImYpnl50V5XQ/edit#slide=id.g308ff111cb5_0_35).
```

docs/uptime/uptime.md
```diff
@@ -0,0 +1,289 @@
...
+
+### max_over_time({{_metric_}}[$frequencyInSeconds])
+
+`max_over_time` is a [Prometheus aggregation function](https://prometheus.io/docs/prometheus/latest/querying/functions/#aggregation_over_time) that returns the maximum value of a metric over a given time range, in this case a check's frequency in seconds: `$frequencyInSeconds`.
+
+We use (`$frequencyInSeconds`) to determine what the 'look back' time should be (which gives us a 'time point'). For example, if the check is running every 15 seconds, the step would be 15 seconds. If the check is running every minute, the step would be 60 seconds.
```
peterschretlen, line 104: I think Marcelo also covers this in his comments, but by using frequencyInSeconds do we create the possibility of getting no samples in a range?  

I think the assumption of a data point in the range will hold much of the time, but what happens when it does not?  I'm thinking of script-based tests in particular, there is variability because of script run duration, lag, and retries that can happen on certain types of failure.

PR #1055, opened 2025-01-31
fix: add displayName on top of name rather than replace

ckbedwell:
## Problem

Introduced a small regression in https://github.com/grafana/synthetic-monitoring-app/pull/1050. Naively mutated the `name` attribute in the probes rather than add an additional field. This caused the linking to the checks list with a probe filter to break.

## Solution

In the interest of keeping backwards compatibility, I've made the displayName field additive rather than mutate the name field. I've added a test for the probe filter (it was the only filter type missing a test 😅 ) on the Checks List page so it doesn't regress again.

I've also gone through and adapted the types so they are more accurate, adding a `ProbeWithMetadata` type after it gets mapped onto.


VikaCep (approved):
LGTM! 👏

PR #1056, opened 2025-01-31
fix: pull updater image from gar instead of gcr

dblinkhorn:
As part of https://github.com/grafana/deployment_tools/issues/197258, we are replacing `updater` image references using GCR to now use GAR. GCR is deprecated and the `updater` image that lives there is no longer being updated.

See: https://github.com/grafana/deployment_tools/issues/197240


ckbedwell (approved):
LGTM

PR #1057, opened 2025-02-03
chore: fix test noise

ckbedwell:
## Problem

We've let some test noise creep into our tests which over time has built up to become pretty noisy and frustrating to scroll through when seeing why tests might be failing -- especially in CI/CD.

## Solution

I've gone through and fixed the problems at the root level where possible in our codebase, otherwise I've added them to the `silenceErrors.ts` file we have so they are muted. We should periodically review them and see if we can remove them, with the aim of being able to delete this file in the near future.

I've also added a little bit of documentation which we can build upon moving forwards.


w1kman (approved):
Lovely! <3 💪🏻 🥇

PR #1059, opened 2025-02-04
fix: get web vitals value for gauge component

VikaCep:
Fixes the web vitals components as they were displaying 0.0 instead of the reduced data series value. This happened because we were relying on the `refId` of the series in order to obtain its value, but this value has been changed recently and it broke our implementation. Since there’s only one series per query, I simplified the implementation to directly access the first series, removing the need for `refId` filtering.

Closes https://github.com/grafana/support-escalations/issues/14520

Before
image

After
image

Notes:

The `refId` value from the data series changed to return a prefix of the transformation applied to it (`reduce` in our case): 

image

which the current app code wasn't expecting:

image


ckbedwell (approved):
LGTM

PR #1064, opened 2025-02-10
feat: improve new alerts UI

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring/issues/242

Improves UI for new alerts per check:

- Removed all alerts except `ProbeFailedExecutionsTooHigh` and `HTTPTargetCertificateCloseToExpiring` as we will launch only those two at first (also, the backend has [removed](https://github.com/grafana/synthetic-monitoring/issues/214#issuecomment-2703013044) support for it as well)

- Added a new `period` dropdown for `ProbeFailedExecutionsTooHigh` alert based on what's defined [here](https://docs.google.com/document/d/1kqzJn66JYRb_FH-WKDRYbvj3LwMv1HEoT_90TNuqwNo/edit?tab=t.0#heading=h.6robak73c8qt): `1m, 2m, 5m, 10m, 15m, 20m, 30m, 1h`

- Some text updates:

  - Changed the page description copy to "Enable and configure thresholds for common alerting scenarios.  Use Grafana Alerting to [create a custom alert rule](https://ops.grafana-ops.net/alerting/new/alerting?returnTo=%2Falerting%2Flist)"
  - Tweaked the certificate expiry copy to "Alert if the target's certificate expires in less than 60d"
  - Changed probe failed executions alert copy to "Alert if at least x of n tests fail in the last T min"
  - Change System Health heading to TLS Certificate
  - Removed predefined alerts heading
  - Renamed alerts per check to Predefined alerts in tab

https://github.com/user-attachments/assets/99807292-f125-4706-a7d5-a186b02ac8b9
image


src/components/CheckForm/AlertsPerCheck/FailedExecutionsAlert.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <Stack alignItems="center">
+        <Checkbox
+          id={`alert-${alert.type}`}
+          data-testid={`checkbox-alert-${alert.type}`}
+          onClick={() => handleToggleAlert(alert.type)}
+          checked={selected}
+        />
+        <Stack alignItems="center">
+          Alert if <ThresholdSelector alert={alert} selected={selected} />
+          of {testExecutionsPerPeriod} tests fail in the last
```
ka3de, line 74: The PR description mentions that this alert is phrased as:

> Alert if at least x of n tests fail in the last T min

Which matches how the alert/threshold behaves, but this does not seem to comply with that.
Should we adopt that sentence?

src/components/CheckForm/CheckForm.tsx
```diff
@@ -242,9 +252,33 @@ export const CheckForm = ({ check, disabled }: CheckFormProps) => {
               </FormLayout.Section>
               <FormLayout.Section label="Alerting" fields={[`alerts`, `alertSensitivity`]} status={status}>
                 <FeatureFlag name={FeatureName.AlertsPerCheck}>
-                  {({ isEnabled }) => (isEnabled ? <AlertsPerCheck onInitAlerts={handleInitAlerts} /> : null)}
+                  {({ isEnabled }) =>
+                    isEnabled ? (
+                      <>
+                        <TabsBar>
+                          <Tab
+                            label="Alerts-per-check"
```
ka3de, line 260: I know we've had various discussions on the naming and we seemed to have agreed to this name, however when discussing the documentation, the naming suggested has slightly changed to _"Per-check alerts"_. I believe that feels more natural for the docs, and probably for this tab itself. 
Should we adopt that naming? Happy to hear other thoughts on this.

ka3de (commented):
Just adding a couple of small comments on the very complex topic of naming.

ckbedwell (changes requested):
Code is fine, just left a few comments. Might be useful to write some internal documentation for ourselves? So want to focus on the experience:

image

1. The default of 10 doesn't make sense for the Failed check alert.
2. The 0 tests fail also makes no sense to me as a user. We also need to change the wording from`tests` to `executions` so are consistent in how we reference our primitives. We should also create a `pluralize` util function so we don't have the awkward grammar when it is singular.
3. I just turned the alert on and then went to save and it failed to save. It didn't take me back to the alerting step or tell me the problem. The error message from the API gives me some idea but it isn't great: `invalid alert definition for ProbeFailedExecutionsTooHigh: invalid alert period: period not supported for alert`.

☝️ From inspecting the payload it looks like the period wasn't included when submitting to the API?
image

We need to tighten up our UI validation because it is an awkward experience because it isn't very clear to me that I've actually made a check that has started running without alerting at this point (I hate 5 second toasts -- a11y hell).

After I went to fix the alert, I couldn't update it because the `target/job combination already exists.` so we need to fix that, too.

The error message when I delete the first input isn't great and also doesn't render very well with the sentence:

image

I could also add numbers (1.5) which weren't integers, which it seemed fine with:

image

I could also add numbers that seemed totally wrong (100). On that note, I tried adding 101 and it wouldn't accept it. Is the underlying check expecting a percentage or a hard-coded execution count?

If I went to edit a check which had an active alert, and went to the alerting step even if I don't change any values I get the 'Unsaved changes' confirmation when trying to navigate away. I didn't seem to get this behaviour if I didn't visit the alerting step or if I went to a check which didn't have an active alert. 

<hr>

The variable part of the sentence `of x tests` I find so awkward as a user. There is nothing that highlights this number changes dependent on my other settings. I know it has a little hover tooltip trying to explain it but even then it still isn't clear to me. It's the fact I read it left to right but the right-most part of the sentence altars something that comes before it. I think we should change it to:

`In the last ${period} alert if ${executionCount} probe execution(s) fail`.

image

Also, an experience that felt backwards is that my alert period wasn't honoured if I changed my frequency at the next step. I originally set it at 1 minute but because I changed my frequency it was removed. We need to take the user back to the alerting step before submitting to fix this. We should also make it clear that some options aren't available because of the chosen frequency. Add something like 'Minimum period is 5 minutes because of check frequency'.

The more I play with this feature the more I am convinced the ideal experience would be if we swap the Alerting and Execution steps but that might be something to explore for @w1kman's Phase 2 check creation?

Perhaps instead of having examples powered by the user's selection for now we should hard-code an example in the tooltip or link to the docs with some examples or recipes for best practice that users can extrapolate from: 'If you have 1 probe enabled with a 5-minute frequency, you would want a 15-minute evaluation period with a threshold of 3 failing checks'.

src/components/CheckForm/AlertsPerCheck/FailedExecutionsAlert.tsx
VikaCep: Fixed it 👍

src/components/CheckForm/CheckForm.tsx
VikaCep: Changed it to `Per-check alerts` to be consistent with the docs https://github.com/grafana/website/pull/24518, but I'll change it if we reach consensus on something else 👍

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
```diff
@@ -3,103 +3,90 @@ import { CheckAlertCategory, CheckAlertType, CheckType, ThresholdUnit } from 'ty
...
+];
+
 export const GLOBAL_PREDEFINED_ALERTS: PredefinedAlertInterface[] = [
   {
     type: CheckAlertType.ProbeFailedExecutionsTooHigh,
     name: 'Probe Failed Executions Too High',
-    unit: '%',
-    category: CheckAlertCategory.SystemHealth,
+    description: '',
+    hide: false,
```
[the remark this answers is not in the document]
VikaCep: I should've deleted it when I removed the code for the alerts that we're not showing anymore. Removed it 🧹

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
```diff
@@ -3,103 +3,90 @@ import { CheckAlertCategory, CheckAlertType, CheckType, ThresholdUnit } from 'ty
...
     default: 10,
+    query: `
+    (
+      (1 - (
+        sum by (instance, job) (
+          increase(probe_all_success_sum{instance="$instance", job="$job"}[$period])
+        ) 
+        / 
+        sum by (instance, job) (
+          increase(probe_all_success_count{instance="$instance", job="$job"}[$period])
```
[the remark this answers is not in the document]
VikaCep: Fixed it (it was not initializing the period correctly when not explicitly setting it in the dropdown)

src/components/CheckForm/AlertsPerCheck/AlertItem.tsx
```diff
@@ -20,54 +34,58 @@ export const AlertItem = ({
...
+  const threshold = getValues(`alerts.${alert.type}.threshold`);
+  const period = getValues(`alerts.${alert.type}.period`);
+
+  const query = alert.query
+    .replace(/\$instance/g, instance)
+    .replace(/\$job/g, job)
+    .replace(/\$threshold/g, threshold)
+    .replace(/\$period/g, period);
+
+  const exploreLink = ds && getValues('id') && threshold && createExploreLink(ds.name, query);
```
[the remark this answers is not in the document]
VikaCep: I chose not to show it in that case because If the test is not created then the explore query will return no results, as there will be no executions. I can still add it, although data in the query might be incomplete (for example, instance or job could be empty)

image

src/components/CheckForm/AlertsPerCheck/AlertItem.tsx
```diff
@@ -1,13 +1,27 @@
...
+
+function createExploreLink(dataSourceName: string, query: string) {
+  return urlUtil.renderUrl(`/explore`, {
+    left: JSON.stringify([
+      'now-5m',
+      'now',
+      dataSourceName,
+      { datasource: dataSourceName, expr: query },
+      { ui: [true, true, true, 'none'] },
+    ]),
```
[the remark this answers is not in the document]
VikaCep: Good point. I changed to 3 hours to be consistent with the other links we provide to explore. Regarding b), I'm not sure at this point where I got it from 🤦 but I removed it as it doesn't seem necessary.

VikaCep:
@ckbedwell I’ve made several improvements to the validation logic and its display, as well as added more descriptive messaging to clarify where the Y value is coming from, based on our conversations in Slack.


image

I’ve kept both the `x of Y` format and the message below for now, as it seemed like the option that would work best for everyone. However, let me know if you feel it’s too redundant, and I can adjust it accordingly.

Additionally, I updated the tooltip to clarify that the periods in the dropdown are displayed based on the check frequency:

image

Also, now if we change the frequency, these values will be adjusted accordingly. 

https://github.com/user-attachments/assets/ad770b54-21b0-4a37-89b5-9517f827bfb5

I'd appreciate any feedback. Thank you!

VikaCep:
In the last commits I 

- Incorporated suggestions from Chris' commit (https://github.com/grafana/synthetic-monitoring-app/compare/alerts-per-check-2...suggestion/alerts-per-check-2?expand=1)
- Changed the alerting endpoint to use `PUT` method instead of `POST` as per https://github.com/grafana/synthetic-monitoring-api/pull/1286
- Changed the wording on the alerting evaluation message as well as formatting the frequency so it displays in the right unit.

image

ckbedwell (approved):
LGTM 😍

ckbedwell (approved):
LGTM 🌟

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
```diff
@@ -3,103 +3,96 @@ import { CheckAlertCategory, CheckAlertType, CheckType, ThresholdUnit } from 'ty
...
+    supportsPeriod: true,
+    unit: 'no.',
+    category: CheckAlertCategory.FailedChecks,
+    defaultValues: {
+      threshold: 1,
+      period: '5m',
+      isSelected: false,
+    },
+    query: `
+    (
```
ka3de, line 41: Notice that this query has now changed. You can see the current query [here](https://github.com/grafana/synthetic-monitoring-api/blob/914e60a309b4503656e0b9ed0fe6f53017164003/internal/alerts/rules.go#L17).

VikaCep: Thanks for the catch, I've updated it [here](https://github.com/grafana/synthetic-monitoring-app/pull/1064/commits/640446657fad542575e276f0653ec635b7c679c2)

ka3de: Should this: https://github.com/grafana/synthetic-monitoring-app/blob/640446657fad542575e276f0653ec635b7c679c2/src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx#L48-L50
Be `$threshold` as it was [before](https://github.com/grafana/synthetic-monitoring-app/blob/8a2c7f294013999537c9c8277da3b2a3208e2511/src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx#L52)? As I understand that references the value that the user set in the form when editing the check? Instead of referencing the current threshold value that is represented by the metric.

VikaCep: Makes sense. Changed it 👍 Attaching screenshots with an example of how the alert configuration translates to the query.
image
image

ckbedwell (approved):
LGTM! 🥇

PR #1066, opened 2025-02-12
fix: probe filtering behaviour and reset button

ckbedwell:
## Problem

We had a bug where you couldn't filter most probes correctly on the check list page. We were made aware of it in this escalation: https://github.com/grafana/support-escalations/issues/14671

We had unit and integration tests covering this problem but none that were comprehensive enough to catch the bug.

Whilst fixing this bug I discovered the reset button also wasn't working so put in a quick fix.

## Solution

I've added the `displayName` property to our hard-coded probe meta data list. Upon reflection this is the safer way to handle this problem as it no longer transforms private probe names, which would likely want to restrict their display to the user-inputted name.

Regardless, I've updated `camelCaseToSentence` to `pascalCaseToSentence`. It ensures that if the first letter is a capital it doesn't get prepended with a space and for values which already have a space it no longer double pads them.

I've also changed the `reset` button so it directly alters the URL, which the state then passes down to the filters to update themselves accordingly. This isn't an ideal fix but as we have re-doing the check list page on the horizon I don't want to invest an unnecessary amount of time for something that is likely to change in some capacity in the future.

### DIFF
--- a/src/components/CheckEditor/ProbesMetadata.ts
+++ b/src/components/CheckEditor/ProbesMetadata.ts
@@ -23,6 +23,16 @@ const COUNTRY_AE = { code: 'AE', long: 'United Arab Emirates' };
 const COUNTRY_ES = { code: 'ES', long: 'Spain' };
 const COUNTRY_ID = { code: 'ID', long: 'Indonesia' };
 
+export const EMPTY_METADATA: ProbeMetadata = {
+  name: '',
+  displayName: '',
+  region: '',
+  longRegion: '',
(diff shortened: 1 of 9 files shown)


VikaCep (changes requested):
I'm still having issues selecting the probes in the filter:


And the reset button is not working:

ckbedwell:
@VikaCep Forgot to `git push` 🤦

VikaCep (approved):
💯 👏 🥇

PR #1068, opened 2025-02-18
fix: prevent submitting alerts per check when FF is off

VikaCep:

Prevents executing the new alerts `POST` request to update alerts when saving a check if the feature flag is off.

This was a regression introduced after a refactor PR. Before, we were relying on the absence of the `alerts` property from the form in order to save them or not. Since after the refactor the `alerts` prop is always present, I'm now relying on the status of the feature flag directly.

Ref https://raintank-corp.slack.com/archives/C0162C1K9E1/p1739894967044089


ckbedwell (approved):
LGTM. Good catch!

Issue #1071, opened 2025-02-20
Bulk probe editing allows you to add the same probe multiple times


peterschretlen:
**What happened**:

My check shows as running from 5 locations, but I only have it set to run in 3.

Check list view:
<img width="848" alt="Image" src="https://github.com/user-attachments/assets/886d0965-2789-4716-9256-4b911a5fb795" />

Edit view shows a count of 4 in EMEA, but only 2 locations are checked off:
<img width="983" alt="Image" src="https://github.com/user-attachments/assets/203b220b-a6b4-44e2-8955-8971a8d7ac1a" />

I managed to get into this state using the bulk probe edit tool. 

**What you expected to happen**:

The bulk editing should validate and remove duplicate entries AND the API should not accept duplicates (logged as a separate issue in https://github.com/grafana/synthetic-monitoring-api/issues/1175 )

**How to reproduce it (as minimally and precisely as possible)**:

Step 1:  Create 3 checks, `a`, `b`, and `c`.  
   - `a` uses Amsterdam
   - `b` uses Bangalore
   - `c` uses Capetown
 
<img width="1120" alt="Image" src="https://github.com/user-attachments/assets/5fada490-a5fd-45c9-a38c-c503489d3cde" />

Step 2:  Bulk edit and select all 3 probes, adding "Amsterdam" to all 3.  Notice how `a` is now running in 2 locations, despite only using Amsterdam. 

<img width="1088" alt="Image" src="https://github.com/user-attachments/assets/54d68db4-f70d-4d38-8938-7d8ea4ba7724" />

Step 3: Look at the check list data in devtools.  For check `a`, notice how it lists the same probe twice

<img width="426" alt="Image" src="https://github.com/user-attachments/assets/4abfc9c9-fc84-4ce3-b227-de555292c796" />

Step 4:  Bulk edit and select all 3 probes, adding "Bangalore" to all 3.  Look at the request the frontend sends to the `check/update/bulk` API for check `b`.  The probe "2" (Bangalore) is being sent twice:

```
{
  "id": 8432,
  "tenantId": 1199,
  "frequency": 60000,
  "offset": 0,
  "timeout": 3000,
  "enabled": true,
  "labels": [],
  "settings": {
    "http": {
      "ipVersion": "V4",
      "method": "GET",
      "noFollowRedirects": false,
      "failIfSSL": false,
      "failIfNotSSL": false
    }
  },
  "probes": [
    2,
    7,
    2
  ],
[...]

PR #1072, opened 2025-02-20
fix: prevent adding duplicated probes when bulk editing

VikaCep:

Fixes a wrong behavior when bulk-editing checks that ended up adding duplicated probe locations to the check.

https://github.com/user-attachments/assets/edf1deba-94d3-4319-b1a7-ccfa7b4a62d2

Closes https://github.com/grafana/synthetic-monitoring-app/issues/1071


ckbedwell (approved):
LGTM ⭐ 🚤 💨

PR #1074, opened 2025-02-21
fix: handle fractional "check result" logs

nadiamoe:

When https://github.com/grafana/xk6-sm/pull/34 is merged, it will be possible for `check result` logs to have fractional values, if the same check name is used twice. For example, the following snippet:
```js
  check(res, {
      'is 200': () => true,
    }
  );
  check(res, {
      'is 200': () => false,
    }
  );
```

Currently yields:
```
time="2025-01-14T17:32:54+01:00" level=info msg="check result" check="is 200" metric=checks_total scenario=default source=sm value=1
time="2025-01-14T17:32:55+01:00" level=info msg="check result" check="is 200" metric=checks_total scenario=default source=sm value=0
```

After https://github.com/grafana/xk6-sm/pull/34 is merged, it will yield
```
time="2025-01-14T17:32:54+01:00" level=info msg="check result" check="is 200" metric=checks_total scenario=default source=sm value=0.5
```

In terms of reporting, and given this will only happen if two `check`s with the same name are defined, I think it is okay to simplify and assume that if not 100% of the `check`s with the same name are okay, the check has failed. So this PR changes the query to look for `!= 1` instead of `= 0`. Other queries already look for `= 1`, so those shouldn't need changing.

### DIFF
--- a/src/scenes/Common/AssertionsTable/AssertionsTable.tsx
+++ b/src/scenes/Common/AssertionsTable/AssertionsTable.tsx
@@ -36,7 +36,7 @@ function getQueryRunner(logs: DataSourceRef) {
           | keep check
           [$__range]
         )
-        / 
+        /
         count_over_time (
             {job="$job", instance="$instance", probe=~"$probe"}
             | logfmt check, msg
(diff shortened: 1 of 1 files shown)


VikaCep:
In the example you provided, then with these changes both checks would fail, is that right?

nadiamoe:
Yes, I think this would be a good approximation to handle this corner case: If at least one instance of `check` of a given name fails, then the check should be reported as failed.

PR #1081, opened 2025-03-05
feat: checks empty state

w1kman:
This PR polishes the empty state for the home route as well as the check list.
Currently, when there are no checks a bland card is used to inform the user.
Since `grafana/ui` has an `<EmptyState />` component, it felt appropriate to use it.

**Home**
image

**Check list**
image

Bonus: Changed the `<CheckUsage />` label to "Estimated usage for this check" from "Approximate expected usage for this check" as well as added tests for said changes.

image


ckbedwell (approved):
LGTM 🌟  Great work 👏

Issue #1090, opened 2025-03-19
"Edit Check" button disabled when comma in `instance`


w1kman:
![Image](https://github.com/user-attachments/assets/7b258b47-9a4c-4bce-b8c2-53714a0856c2)

**What happened**:
When the `instance´ value includes a comma, the "Edit Check" button on the check dashboard page, gets disabled.

**What you expected to happen**:
I expect the "Edit Check" button to not be disabled

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a check that has a comma in the `instance` (scripted check has the `instance` field exposed, for easy manipulation).
2. Go to the check dashboard
3. See the "Edit check" button

**Anything else we need to know?**:
The cause of the bug is related to https://github.com/grafana/scenes/issues/683
There is no need to use `instance` + `job` to figure out what `id` the check button should use.
The `check.id` can be passed to the EditButton (instead of taking `check.job` + `check.target|instance`, parsing them with `CustomVariable`, just to get the `id`).

PR #1091, opened 2025-03-19
fix: edit check button disabled issue

w1kman:
**Before**
image

**After**
image

When an `instance` value includes a comma, we cannot enable the "Edit Check" button on the dashboard page for an affected check. The reason is that we take the `job` and `target|instance` value from the check object (which includes the `id`), then parse both values with `CustomVariable`, then use the parsed values to find an item in an array, that matches `job` and `target|instace`, to get the `id`

This PR takes the `id` directly and uses the `id` to create the route for the Edit Button

Resolves [1090](https://github.com/grafana/synthetic-monitoring-app/issues/1090)

Fixes #


ckbedwell (approved):
LGTM 🦸

PR #1093, opened 2025-03-20
fix: invert order between Alerting/Execution when AlertsPerCheck is e…

VikaCep:
This PR adjusts the order of the "Alerting" and "Execution" sections in the check form based on the state of the AlertsPerCheck feature flag. When the flag is enabled, the "Execution" section is displayed before the "Alerting" section.

This change aligns with the user flow, as customers must first configure the frequency (in the "Execution" section), which then determines the available alert periods in the "Alerting" section. 

When `AlertsPerCheck` is on:
image

When `AlertsPerCheck` is off:
image

### DIFF
--- a/src/components/CheckForm/AlertsPerCheckSection.tsx
+++ b/src/components/CheckForm/AlertsPerCheckSection.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { Tab, TabContent, TabsBar } from '@grafana/ui';
+import { Stack, Tab, TabContent, TabsBar } from '@grafana/ui';
 
 import { AlertingType, CheckAlertFormValues, CheckAlertType } from 'types';
 import { AlertsPerCheck } from 'components/CheckForm/AlertsPerCheck/AlertsPerCheck';
@@ -8,18 +8,16 @@ import { CheckFormAlert } from 'components/CheckFormAlert';
 interface AlertsPerCheckSectionProps {
   handleInitAlerts: (alerts: Partial<Record<CheckAlertType, CheckAlertFormValues>>) => void;
   alertsInitialized: boolean;
-  styles: { wrapper: string };
 }
 
 export const AlertsPerCheckSection: React.FC<AlertsPerCheckSectionProps> = ({
   handleInitAlerts,
   alertsInitialized,
-  styles,
 }) => {
   const [selectedAlertingTab, setSelectedAlertingTab] = useState<AlertingType>('alerting');
 
   return (
-    <>
+    <Stack direction="column" gap={2}>
       <TabsBar>
         <Tab
           label="Per-check alerts"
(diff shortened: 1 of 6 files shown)


src/components/CheckForm/CheckForm.tsx
```diff
@@ -216,6 +216,33 @@ export const CheckForm = ({ check, disabled }: CheckFormProps) => {
...
+      <TabContent>
+        <div className={styles.wrapper}>
+          {selectedAlertingTab === 'alerting' && (
+            <AlertsPerCheck onInitAlerts={handleInitAlerts} isInitialized={alertsInitialized} />
+          )}
+          {selectedAlertingTab === 'sensitivity' && <CheckFormAlert />}
+        </div>
+      </TabContent>
+    </>
+  );
```
ckbedwell, line 242: Let's extract this so it is in the global scope (could even be its own component). I triggered co-pilot as a reviewer because I was interested to see if it picks up on the fact this gets redefined on every render or not (it didn't).

ckbedwell (changes requested):
Just one small change otherwise looks good to me!

src/components/CheckForm/CheckForm.tsx
VikaCep: Good call, done 👍

src/components/CheckForm/AlertsPerCheckSection.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Tab, TabContent, TabsBar } from '@grafana/ui';
+
+import { AlertingType, CheckAlertFormValues, CheckAlertType } from 'types';
+import { AlertsPerCheck } from 'components/CheckForm/AlertsPerCheck/AlertsPerCheck';
+import { CheckFormAlert } from 'components/CheckFormAlert';
+
+interface AlertsPerCheckSectionProps {
+  handleInitAlerts: (alerts: Partial<Record<CheckAlertType, CheckAlertFormValues>>) => void;
+  alertsInitialized: boolean;
+  styles: { wrapper: string };
```
ckbedwell, line 11:
```suggestion
```

src/components/CheckForm/AlertsPerCheckSection.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export const AlertsPerCheckSection: React.FC<AlertsPerCheckSectionProps> = ({
+  handleInitAlerts,
+  alertsInitialized,
+  styles,
```
ckbedwell, line 17:
```suggestion
```

src/components/CheckForm/AlertsPerCheckSection.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useState } from 'react';
+import { Tab, TabContent, TabsBar } from '@grafana/ui';
```
ckbedwell, line 2:
```suggestion
import { Stack, Tab, TabContent, TabsBar } from '@grafana/ui';
```

ckbedwell (changes requested):
🙈  the passing of `styles` as a prop immediately set off my spidey senses. Had a quick test and using `Stack` gets you the same result 😃

PR #1096, opened 2025-03-21
chore: bump k6 typings to v0.57.1

w1kman:
Bumps k6 types. To have the correct types for the current k6 version


VikaCep (changes requested):
Do we need to update `k6.types.ts`? It's complaining now:

image

ckbedwell (approved):
LGTM (🕷️👨👉   👈👨🕷️)

w1kman:
Good catch @VikaCep, thanks for the update @ckbedwell <3

PR #1099, opened 2025-03-25
fix: update drone signature to prevent having to manually approve builds

VikaCep:
Changes the drone signature so that builds don't need to be manually approved.

It's been changed by running `./scripts/drone-sign.sh`


VikaCep:
> LGTM - I assume you didn't have to approve this build?

Right! This build ran automatically :)

PR #1101, opened 2025-03-27
feat: add pipeline dependencies and fix canary config

VikaCep:
# Add Pipeline Dependencies and Fix Configuration

## Changes
1. Added pipeline dependencies to ensure proper release sequence:
   ```yaml
   # staging pipeline
   depends_on:
     - publish-dev

   # canary pipeline
   depends_on:
     - publish-staging

   # prod pipeline
   depends_on:
     - publish-canary
   ```

2. Fixed configuration overlap by removing `prod-canary.libsonnet` update from production pipeline, ensuring:
   - Canary pipeline exclusively manages `prod-canary.libsonnet`
   - Production pipeline only manages `prod.libsonnet`

3. Added Thomas (w1kman) as a reviewer for wave updates.

## Why
- Enforces correct release progression: dev → staging → canary → prod
- Prevents out-of-order deployments
- Maintains clear separation of responsibilities between pipelines
- Fixes the new failing check requiring single wave updates (see deployment_tools [PR](https://github.com/grafana/deployment_tools/pull/241350#issuecomment-2752405212))

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -65,7 +65,7 @@ steps:
           "pull_request_enabled": true,
           "pull_request_existing_strategy": "ignore",
           "pull_request_message": "Triggered by ${DRONE_COMMIT_LINK}. NOTE: dev does not refer directly to an environment it refers to stacks associated with the dev 'wave'. See [here](https://github.com/grafana/deployment_tools/blob/master/ksonnet/environments/hosted-grafana/waves/provisioned-plugins/README.md#waves) for more info.",
-          "pull_request_reviewers": ["VikaCep", "ckbedwell"],
+          "pull_request_reviewers": ["VikaCep", "ckbedwell", "w1kman"],
           "update_jsonnet_attribute_configs": [
             {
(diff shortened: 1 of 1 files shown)


w1kman (commented):
Looks good and most likely is super good.
I just don't know how to verify that it works as intended.

VikaCep:
While testing the `staging` promotion, I discovered that the `depends_on` clause wasn't working as expected. After some investigation, I found that it only works when pipelines are triggered by the same event and target. 

We could add some custom validation instead, but in the end I decided to remove this `depends_on` clause since:
1. It's not critical at this moment
2. We're planning to migrate to GitHub Actions soon
3. The main goal of this PR is to remove the unnecessary canary step from the deployment process

The most important aspect is ensuring that the `production` pipeline doesn't update the `prod-canary.libsonnet` file, which would cause the Single wave check to fail.

We will verify that the release process works as expected in the next release but we can be confident that no canary-related processes are being executed after these changes, as it's all been removed from the drone configuration.

ckbedwell (approved):
LGTM!

...(_Please work_ 😅🤞)

PR #1107, opened 2025-04-04
fix: typing errors after upgrading dependencies

VikaCep:
After recreating the `yarn.lock` file (https://github.com/grafana/synthetic-monitoring-app/pull/1105) we got some typing errors mainly related to React Hook Form upgrading from 7.51.5 to 7.55.0. This PR fixes them.

image

### DIFF
--- a/src/components/CheckForm/FormLayout/FormLayout.tsx
+++ b/src/components/CheckForm/FormLayout/FormLayout.tsx
@@ -8,7 +8,7 @@ import { DataTestIds } from 'test/dataTestIds';
 
 import { flattenKeys } from '../checkForm.utils';
 import { normalizeFlattenedErrors, useFormLayout } from './formlayout.utils';
-import { FormSection, FormSectionInternal } from './FormSection';
+import { FormSection, FormSectionInternal, FormSectionProps } from './FormSection';
 import { FormSidebar } from './FormSidebar';
 
 type ActionNode = {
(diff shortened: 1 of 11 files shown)


src/components/Request/RequestOptions.tsx
```diff
@@ -40,8 +40,8 @@ export const RequestOptions = forwardRef<HandleErrorRef, RequestOptionsProps>(({
 
                 return (
                   <Tab
-                    key={section.props.label}
-                    label={section.props.label}
+                    key={(section as ReactElement<TabProps>).props.label}
+                    label={(section as ReactElement<TabProps>).props.label}
```
ckbedwell, line 44: If we change the interface for `RequestOptions` to:

```
type RequestOptionsChild = ReactElement<TabProps> | undefined | null;

interface RequestOptionsProps {
  children: RequestOptionsChild[] | RequestOptionsChild;
  open?: boolean;
}
```

we can remove all the type casting here (and actually makes it more correct) :)

src/components/CheckForm/FormLayout/FormLayout.tsx
```diff
@@ -59,7 +59,9 @@ export const FormLayout = <T extends FieldValues>({
         if (child.type === FormSection) {
           index++;
 
-          return <FormSectionInternal {...child.props} index={index} activeSection={activeSection} />;
+          const sectionProps = child.props as Omit<FormSectionProps, 'index' | 'activeSection'>;
```
ckbedwell, line 62: I spent a while trying to find a way round this type casting but concluded anything I came up with was essentially a type cast with extra steps so am happy to concede this one is needed 😅

ckbedwell (changes requested):

ckbedwell (approved):
LGTM! 🌟

PR #1109, opened 2025-04-07
Feat/frequency selector v2

ckbedwell:
## Problem
_Repeating content as outlined in: [Project document: SM Frequency V2](https://docs.google.com/document/d/1Jkxqhj-ZPPjjCiuySzh5azu-Gudt5KPCFZzbiUlz9rc/edit?tab=t.2lecoimj8n8e#heading=h.5rh4gn72b49t)_

The current frequency selection experience is difficult and unintuitive to use when utilising the slider. As we are releasing the checks-per-alert feature soon we want to ensure users are guided to using 'sensible' defaults.

## Solution

This PR changes the Frequency input component to the designs provided by @vesnadean, [as found in Figma here](https://www.figma.com/proto/WXbmUSpCMlJHG8fpjvmGT1/%F0%9F%8F%97%EF%B8%8F-Synthetics-%5BDEV-READY%5D?node-id=2457-8737&p=f&t=EABX8zn8HMF4TBbL-0&scaling=min-zoom&content-scaling=fixed&page-id=2457%3A8438). 

image
image

### Why 71 files changed for this PR 🤯 ?

Firstly, if you want me to divide these up into separate PRs I happily will do because this is borderline bad dev citizen behaviour but frustratingly the changes were intrinsically linked so wanted to be clear why they have been changed all at once 😅 :

1. I unified how we refer to duration throughout the application so it is always **milliseconds**. This is a large part of the changes in the PR. I found the inconsistent duration references difficult to reason with so wanted some consistency.
2. I found our validation for timeout and frequency wasn't present in our schemas. I've made it so that the schema values and what appears in the UI are read from the same source of truth.
  - this had a knock-on effect that I discovered that our check type schemas that are derived the `baseCheckSchema` weren't being updated correctly so had to do some updates
3. A fair amount of our tests use the frequency so had to update them


src/schemas/general/Frequency.ts
```diff
@@ -1,8 +1,16 @@
...
-export const frequencySchema = z
-  .number()
-  .max(MAX_BASE_FREQUENCY, { message: `Frequency cannot be greater than ${MAX_BASE_FREQUENCY} seconds` });
+export const MIN_BASE_FREQUENCY = ONE_SECOND_IN_MS * 10;
+export const MAX_BASE_FREQUENCY = ONE_HOUR_IN_MS;
+
+export const MIN_FREQUENCY_ERROR_MESSAGE_START = `Frequency must be at least `;
+export const MAX_FREQUENCY_ERROR_MESSAGE_START = `Frequency cannot be greater than `;
+
+export const frequencySchema = (minFrequency = MIN_BASE_FREQUENCY) =>
```
w1kman, line 12: This is not a schema. It is a schema creator.
IMO the name should be something like `createFrequencySchema`

src/schemas/general/Timeout.ts
```diff
@@ -0,0 +1,14 @@
+import { z } from 'zod';
+
+import { formatDuration } from 'utils';
+import { ONE_MINUTE_IN_MS, ONE_SECOND_IN_MS } from 'utils.constants';
+
+export const MIN_BASE_TIMEOUT = ONE_SECOND_IN_MS;
+export const MAX_BASE_TIMEOUT = ONE_MINUTE_IN_MS;
+
+export const timeoutSchema = (minTimeout = MIN_BASE_TIMEOUT, maxTimeout = MAX_BASE_TIMEOUT) => {
```
w1kman, line 9: This is not a schema. It is a schema creator.
IMO the name should be something like `createTimeoutSchema`

src/utils.test.ts
```diff
@@ -35,3 +35,93 @@ describe(`pascalCaseToSentence`, () => {
...
+      expect(formatDuration(3661000)).toBe('1 hour 1 minute 1 second');
+    });
+
+    it(`formats duration for multiple minutes and seconds`, () => {
+      expect(formatDuration(121000)).toBe('2 minutes 1 second');
+    });
+
+    it(`formats duration for multiple hours and minutes`, () => {
+      expect(formatDuration(7260000)).toBe('2 hours 1 minute');
+    });
```
w1kman, line 121: Nit: Could add tests for all singular/plural combinations.

Currently we wouldn't know (by testing) whether or not it says "1 hour 5 minute" or "1 hour 5 minutes" etc.

w1kman (changes requested):
Mint! Works outstanding, couldn't find any issues (even when I threw stones at the component) 🏅 

There are schema creator functions that are disguised as schemas.

Some general thoughts after looking at all of the code and UI (not expected to be a part of this PR):
- Timeout could/should have the same UX as Frequency.
- We display the Frequency differently between views
  - Checks page (Ns)
  - Dashboard (when applicable): (N s, N minutes)
  - When defining (check form): (Ns, NmNs, etc)

w1kman (approved):
Super!

PR #1111, opened 2025-04-10
chore: rename useQuery hook to useURLSearchParams

ckbedwell:
Since we introduced `react-query` this annoyed me. Today was the day it needed renaming.


VikaCep (approved):
🏅

PR #1114, opened 2025-04-14
feat: add event tracking for check creation and some form navigation

ckbedwell:
## Problem

We have little instrumentation of our application to offer us insights into how users behave and what actions they take whilst using the plugin. With the GA launch of Browser Events upcoming it is essential we start tracking events in the application so we can better inform ourselves what users are doing so we can continuously improve our product.

## Solution

This PR implements an analytics tracking framework which is pretty much directly taken from [this hackathon 12 project](https://github.com/grafana/grafana/pull/101697) in `grafana/grafana` . As that project is only relevant to `grafana/grafana` for the near future it made sense for us to replicate their implementation as we can:

a) offer practical and real dogfooding feedback in a more sandboxed environment (SM is considerably smaller with less contributors than `grafana/grafana`.
b) if and when the project makes it as exposed methods in `grafana/grafana` we will be directly compatible with little pain to switch over.

This PR is only focussed on adding the `createEventFactory` and relevant methods / events. A [companion PR adds the automatically generated analytics documentation](https://github.com/grafana/synthetic-monitoring-app/pull/1119).

https://github.com/user-attachments/assets/3ca9ef97-1764-43bc-a9c1-0a374e613ce9


w1kman (approved):
Sweet! LGTM!

PR #1119, opened 2025-04-15
Feat/event tracking doc gen

ckbedwell:
## Problem

When we add analytics events to the codebase they are difficult to discover for non-developers and others who might want a quick reference of what is available and what properties they provide when utilising Rudderstack and the databases which the data gets piped to.

Despite having a convention  for how we want to handle analytics events (such as putting all event generation in `features/tracking/{FEATURE_NAME}` directories), this still requires a certain level of know-how and a dependency on developers following the convention.

Manual documentation faces the problem where it can easily become out of date as we add and edit existing events.

## Solution

_This solution is based on the [Hackathon 12 project](https://github.com/grafana/grafana/pull/101697) in `grafana/grafana`._

This PR adds several pieces to automate a solution to the problems above which enforces developers to document their analytics events inline, reducing developer friction whilst keeping high quality and accurate analytics documentation.

1. Introduces a custom eslint plugin with its own custom rule which has been defined. This rule does several things:
  - Checks that `createSMEventFactory` has been imported into any files, and if it has:
  - It checks any declared interfaces extend `TrackingEventProps`
  - Those interfaces have JSDoc comments for their properties
  - It also ensures that `createSMEventFactory` is called with only literals and doesn't contain any variables which might create high cardinality events.
2. Adds an eslint rule which forbids the use of `reportInteraction` being used directly.
3. Adds a node script (which can be run using `yarn analytics-events:generate` which creates/updates our `analytics-events.md` file. By using the conventions enforced by the custom eslint rule it ensures the documentation is an accurate representation of what analytics events are in the codebase.

## Open questions

Currently, we have to run the script ourselves to keep our analytics events up to date. Has anyone got suggestions at what point during the CI/CD process we should run it automatically so we don't have to remember to run it?


docs/analytics/analytics-events.md
```diff
@@ -0,0 +1,85 @@
...
+#### synthetic-monitoring_check_creation_add_check_type_button_clicked
+
+Tracks when the 'protocol' buttons on the check type card are clicked.
+
+##### Properties
+
+| name           | type                                                       | description                        |
+| -------------- | ---------------------------------------------------------- | ---------------------------------- |
+| checkTypeGroup | `"api-endpoint" \| "multistep" \| "scripted" \| "browser"` | The check group type of the check. |
+| protocol       | `string`                                                   | The protocol of the check.         |
```
VikaCep, line 40: Shouldn't the protocol here be typed as `CheckType` rather than just `string`?

VikaCep (changes requested):
Works mint! 💪 💪 

Made a small comment about tightening up the `protocol` type in `AddCheckTypeButtonClicked` interface. Right now it's just a string, but we could make it more specific. 

For the analytics docs automation, I was thinking we could:
- Add a pre-commit hook that checks if there are changes and re-generates the file if so (maybe optional)
- Run a task that, before building, compares the generated docs with the committed docs and fails if there are differences.

How does that sound?

docs/analytics/analytics-events.md
ckbedwell: I thought the same and is how I originally had it but realised that didn't make sense for anything that isn't an api-endpoint check. The api-endpoint ones happen to match up with most of our check-types but for the others not so much, especially as we hopefully will add to the browser and scripted ones, too.

VikaCep (approved):
👏 👏 👏 

> I have a bit of reluctance adding a pre-commit hook because they inevitably get turned off or ignored at some point as they make a fairly large difference to the dev experience despite thinking 'what's a few seconds per commit?'. I think your second idea would work, maybe adding it as a GitHub workflow action? It would essentially be a lint check.

Yeah, I guess the commit hook could get annoying as it will make things slower and won't be needed most times. Agree on the Github action 👍

PR #1123, opened 2025-04-24
fix: calculation to monthly usage

VikaCep:
Closes https://github.com/grafana/support-escalations/issues/15913

## Fix: Monthly Usage Calculation
Fixed the calculation of monthly execution counts by removing an unnecessary conversion of frequency from milliseconds to seconds.

## Why
The `getTotalChecksPerMonth` function was being called with a frequency value that was incorrectly converted from milliseconds to seconds. This was causing the monthly execution count to be calculated incorrectly. This was a regression due to a recent [change](https://github.com/grafana/synthetic-monitoring-app/commit/40e84e2d66c67135acb07cc1d433181d3e84ca9a#diff-72f9ee34abb365e308826162ae36d7d6cac2469fa7ba0bad7f4dae6e7a2045c6R34) in said function.

## Changes
- Updated the `useMonthlyTotalExecutionCount` hook to pass the frequency directly without conversion
- Removed the `FREQUENCY_SECONDS` constant in the test file
- Updated the test to use the correct frequency value

## Testing
- Updated the test to use the correct frequency value
- The calculation now correctly reflects the monthly execution count based on the check's frequency in milliseconds

Taking a look at the check depicted in the escalation linked above:

```
  const checks = [
    {
      id: 22153,
      tenantId: 12078,
      frequency: 299000,
      offset: 0,
      timeout: 5000,
      enabled: true,
      labels: [],
      settings: {
        http: {
          ipVersion: 'V4',
          method: 'GET',
          noFollowRedirects: false,
          failIfSSL: false,
          failIfNotSSL: true,
        },
      },
      probes: [19],
      target: 'https://app.droimage.com/',
      job: 'status app.droimage.com',
      basicMetricsOnly: true,
      alertSensitivity: 'high',
      created: 1733998113.209416,
      modified: 1742912030.9396925,
    },
  ];
```

Before the change:
image

After the change:
image

### DIFF
--- a/src/hooks/useMonthlyTotalExecutionCount.test.ts
+++ b/src/hooks/useMonthlyTotalExecutionCount.test.ts
@@ -35,9 +35,8 @@ async function renderExecutionCountHook(checks?: Check[]) {
 }
 
 const FREQUENCY = 60000;
-const FREQUENCY_SECONDS = FREQUENCY / 1000;
 const PROBES = [PRIVATE_PROBE.id!];
-const EXECUTIONS = getTotalChecksPerMonth(PROBES.length, FREQUENCY_SECONDS);
+const EXECUTIONS = getTotalChecksPerMonth(PROBES.length, FREQUENCY);
 
 describe('calculates execution count correctly', () => {
   test('handles calculation for 1 active check correctly', async () => {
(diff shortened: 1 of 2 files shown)


ckbedwell (approved):
LGTM!

Issue #987, opened 2024-11-12
feat: secrets management POC


w1kman:
**What would you like to be added**: A proof of concept implementation/demo of Secrets Management.

**Why is this needed**:
It will allow us to iterate on the UI/UX whilst the core feature is still in development.

**Task list**

- [x] Rework the ConfigPage so that it is tabbed (to prepare for "Secrets" tab).
  - [x] Create a tabbed layout to replace the current `ConfigPage`
  - [x] Split `configPage()` usage and app config into two different components
- [x] Create "Secrets" tab (behind featureToggle)
  - [x] Add a basic layout for creating and viewing secrets (view only shows name of secret, no reveal functionality).

w1kman:
PR for creating tabbed layout for the config page.
https://github.com/grafana/synthetic-monitoring-app/pull/990

PR #1125, opened 2025-04-28
feat: intermediate secrets management pt2

w1kman:
Adds tests and schema validation of secrets. Schema is based on GE spec in order to make migration possible.

An error boundary has been added around the secrets UI to make sure any unexpected errors are captured within it's context.

image

Make sure `synthetic-monitoring-secrets-management=true` is added to feature toggles. 

Resolves https://github.com/grafana/synthetic-monitoring-app/issues/987


src/components/CodeEditor/k6.types.ts
```diff
@@ -41,6 +41,12 @@ import k6ExperimentalWebcrypto from '!raw-loader!@types/k6/experimental/webcrypt
 import k6ExperimentalStreams from '!raw-loader!@types/k6/experimental/streams';
 // @ts-expect-error
 import k6ExperimentalWebsockets from '!raw-loader!@types/k6/experimental/websockets';
+// @ts-expect-error
+import k6Secrets from '!raw-loader!./k6Secrets.d.ts';
+import { isFeatureEnabled } from 'contexts/FeatureFlagContext';
+import { FeatureName } from 'types';
+
+const secretsEnabled = isFeatureEnabled(FeatureName.SecretsManagement);
```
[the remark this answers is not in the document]
w1kman: `isFeatureEnabled` does not rely on any React context in the app.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
```diff
@@ -0,0 +1,122 @@
...
+            variant="call-to-action"
+            message="You don't have any secrets yet."
+            button={
+              <Button onClick={handleAddSecret} icon="plus">
+                Create secret
+              </Button>
+            }
+          >
+            You can use secrets to store sensitive information such as passwords, API keys, and other sensitive
+            information.
```
ckbedwell, line 54:
```suggestion
            You can use secrets to store private information such as passwords, API keys, and other sensitive
            data.
```
Just to avoid some of the repetition and make the copy read a little better.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
```diff
@@ -0,0 +1,122 @@
...
+              <Button size="sm" icon="plus" onClick={handleAddSecret}>
+                Create secret
+              </Button>
+            </div>
+          }
+        >
+          <div>
+            <p>
+              Secrets is a way to store and manage secrets in Grafana Cloud. You can use secrets to store sensitive
+              information such as passwords, API keys, and other sensitive information.
```
ckbedwell, line 71:
```suggestion
              Secrets is a way to store and manage secrets in Grafana Cloud. You can use secrets to store private
              information such as passwords, API keys, and other sensitive data.
```

src/test/handlers/secrets.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const secrets: ApiEntry<SecretsResponse> = {
+  route: `/api/v1alpha1/secrets`,
+  method: `get`,
+  result: () => {
+    return {
+      json: MOCKED_SECRETS_API_RESPONSE,
+    };
+  },
+};
```
ckbedwell, line 14: This needs to be added to the `test/handlers/index.ts` file now. You probably want to add the other secret-related routes to have a default, too, as that makes it easier to mock them specifically during tests as you can use the `server.use(apiRoute(...))` pattern as you can see in some other tests.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementTab.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { SecretsManagementTab } from './SecretsManagementTab';
+
+// Mock a component that throws an error
+jest.mock('./SecretsManagementUI', () => {
+  return {
+    SecretsManagementUI: () => {
+      throw new Error('Test error');
+    },
+  };
+});
```
ckbedwell, line 13: If you hook up your msw secret mocks you can replace this with the following in the test itself, so we are more closely aligned with a real situation.

```
server.use(apiRoute('getSecrets', { result: () => ({ status: 400, body: 'Error message' }) }));
```

ckbedwell (changes requested):
I ❤️ love ❤️  the extensive test coverage but we should look at incorporating MSW to handle mocking the behaviour rather than mocking certain implementation details that might change in the future.

I'd like to aim to get rid of as many of the mocks as possible (I think you can get rid of all of them when the secret routes are all added to msw, to be honest).

Few things I noticed when playing with the UI:

1. The revalidation when updating label name and values doesn't kick in. (e.g. have an empty label / value, when I add it the error doesn't disappear).

2. I noticed that when I try and save a new secret when I haven't touched the textarea it just says: `Required` but if I add something then delete it it then correctly says 'Value is required' 😅 

The invalid state of the secrets from. The value textarea just shows 'Required'.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    render(<SecretCard {...defaultProps} secret={secretWithoutLabels} />);
+
+    expect(screen.queryByText('env: prod')).not.toBeInTheDocument();
+    expect(screen.queryByText('type: api-key')).not.toBeInTheDocument();
+  });
+
+  it('should format the creation date correctly', () => {
+    render(<SecretCard {...defaultProps} />);
+
+    expect(formatDate).toHaveBeenCalledWith(mockSecret.created_at);
```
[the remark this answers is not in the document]
w1kman: @ckbedwell Asserting the DOM would make us double test `formatDate` (assuming that `formatDate` has tests, which it doesn't). 
Ideally, we should be confident enough to assert that `formatDate` was used with the correct args, and have unit tests for `formatDate` to assert that it's functioning as we intend it.

If we assert the DOM, changing the behaviour of `formatDate` would make us have to update the expected value in the `SecretCard.test.tsx`.

IMO we should test that `formatDate` is used to format the date and add tests for `formatDate` instead.
It's your call.

w1kman:
> I'd like to aim to get rid of as many of the mocks as possible (I think you can get rid of all of them when the secret routes are all added to msw, to be honest).

I started by using `msw`, but I felt that adding several files outside of the context of the test was a bit unintuitive, so I bailed on `msw`. I don't share the opinion that using `msw` let's us get rid of mocks, rather it makes us obscure mocks. 🤷🏻 
I could see why `msw` is super useful for e2e tests. I'm not so sure about unit tests (unless testing a data hook).

**TL;DR:** Felt it didn't add much value and that it heavily clouded the readability of the test suite. But I can change it if you think that it is important/better to use `msw` 👍🏻 

> The revalidation when updating label name and values doesn't kick in. (e.g. have an empty label / value, when I add it the error doesn't disappear).
Fixed!

> I noticed that when I try and save a new secret when I haven't touched the textarea it just says: Required but if I add something then delete it it then correctly says 'Value is required' 😅
Fixed

src/test/handlers/secrets.ts
w1kman: Forgot to remove this 😅. Will let it be until you've had a chance to comment on my comment on your comment ✌🏻

ckbedwell:
I need to fully flesh out documenting our testing strategy but I got started [here](https://github.com/grafana/synthetic-monitoring-app/blob/main/docs/development/testing.md). So, everything I am about to write is how most of the tests in the code base act, so that's why I'd prefer consistency.

So, generally speaking we favour integration and e2e tests rather than unit tests (in an ideal world we would have a beautiful mix of them as relevant and needed).

There's a few reasons why I prefer this approach:
1.  It's a bit of a paradigm shift but I think of our testing strategy as there being an overarching framework all the tests operate in which provides some built-in boilerplate and abstractions and once you know them it becomes easier to work with and write durable and long-lasting tests.

2. I strongly believe that integration testing from the user's perspective is more intuitive and creates less developer friction in the long-term than having mocked integration tests or a ton of unit tests that are based on implementation details.

3. The reason why I want our integration tests closer to emulating real server behaviour and asserting on things from the perspective of users is because at some point I think we will be able to start shifting some of our big (slow and somewhat flaky) integration tests to browser-controlled e2e tests and I want it to be ready with as little friction as possible. I want to explore using [vitest's browser mode](https://vitest.dev/guide/browser/) at some point when it is out of being an experimental feature to see if that adoption is viable.

> I started by using msw, but I felt that adding several files outside of the context of the test was a bit unintuitive, so I bailed on msw. I don't share the opinion that using msw let's us get rid of mocks, rather it makes us obscure mocks. 

Part of the framework is assuming every API returns a happy path response and that responsibility is handed off to the testing framework but you could still import the server in the component and explicitly add the API routes and responses so it was less out of context when reading the tests. 

If we take the tests in the `SecretEditModal.test.tsx` my issue is that we have tightly coupled those tests with its implementation details.

The tests rely both on a `useSecret` hook existing and by extension its dependency on us continuing to use `react-query` and the shape of the data it returns. For context, there's a chance plugins might work best in the future by adopting RTK Query or something similar which is provided by the Grafana packages, as that's what `grafana/grafana` uses and if we wanted to take advantage of the global API cache they are considering developing.

If you ever wanted to change the architecture or abstractions (rename / move / change what is returned by `useSecrets`) you have to change the tests, too -- which feels like a bit of a failure why they are there in the first place. You aren't sure if your tests are failing because you genuinely broke the user experience or you just haven't swapped in the new implementation.

The more tests that are coupled with implementation details, the less we encourage devs to solve tech debt or iterate on existing solutions or change existing files. When we get closer to mocking primitives that are out of control (APIs that are likely to remain the same or Product-defined user journeys) the more freedom we create to change and alter our codebase.

I know the API route for secrets is going to change at some point in the future and it would give me more confidence that our UI will handle that change if we update the mocked API to match it and ran our tests rather than hoping the new API is compatible with the existing implementation (which the tests without using MSW are blind to).

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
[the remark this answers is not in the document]
ckbedwell: What about changing it so it is:

```
expect(screen.getByText(formatDate(mockSecret.created_at)).toBeInTheDocument()
```

That is essentially the same as checking formatDate is called but also checking the user can see it?

src/types.ts
```diff
@@ -712,6 +712,7 @@ export enum FeatureName {
   UnifiedAlerting = 'ngalert',
   RBAC = 'synthetic-monitoring-rbac',
   AlertsPerCheck = 'sm-alerts-per-check',
+  SecretsManagement = 'synthetic-monitoring-secrets-management',
```
VikaCep: Right now we have a mixture of names, but should we set on always prefixing feature flags with the `synthetic-monitoring-` prefix from now on? I know we have talked about this before @ckbedwell

src/page/ConfigPageLayout/ConfigContent.tsx
```diff
@@ -23,13 +25,24 @@ interface WithActions {
 
 type Props = BaseProps & (WithoutActions | WithActions);
 
-export function ConfigContent({ title, children, loading = false, actions }: Props) {
+export function ConfigContent({
+  title,
+  children,
+  loading = false,
+  actions,
+  ariaLoadingLabel = 'Loading...',
```
VikaCep, line 33: The `CenteredSpinner` already sets a `Loading` aria-label by default when nothing is provided, so maybe we can omit this

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    created_at: Number(new Date('2024-01-01')),
+    created_by: 'test-user',
+    labels: [
+      { name: 'env', value: 'prod' },
+      { name: 'type', value: 'api-key' },
+    ],
+    modified_at: Number(new Date('2024-01-02')),
+    org_id: 1,
+    stack_id: 1,
+  };
```
VikaCep, line 41: Why not use the a secret from `MOCKED_SECRETS` here instead?

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    fireEvent.click(editButton);
+
+    expect(defaultProps.onEdit).toHaveBeenCalledWith('secret-123');
+  });
+
+  it('should call onDelete when delete button is clicked', () => {
+    render(<SecretCard {...defaultProps} />);
+
+    const deleteButton = screen.getByRole('button', { name: 'Delete Test Secret' });
+    fireEvent.click(deleteButton);
```
VikaCep, line 80: For component interactions, we [should be using](https://ph-fritsche.github.io/blog/post/why-userevent) `userEvent` instead of `fireEvent`

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    // Find the clipboard component
+    const clipboardButton = screen.getByRole('button', { name: /Copy Test Secret ID/i });
+    await act(() => {
+      return user.click(clipboardButton);
+    });
+    // Instead of accessing props directly, test the actual copy behavior
+
+    const clipboardText = await navigator.clipboard.readText();
+    // Verify the copy behavior through the clipboard API
+    // You might need to mock navigator.clipboard
```
VikaCep, line 124: you already did this, right? Is this comment relevant?

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+import { render as testRender } from 'test/render';
+import { server } from 'test/server';
+
+import { SECRETS_EDIT_MODE_ADD } from './constants';
+import { SecretEditModal } from './SecretEditModal';
+
+async function render(element: React.ReactElement) {
+  const result = testRender(<div data-testid={DataTestIds.CONFIG_CONTENT}>{element}</div>);
+  await waitFor(() => expect(screen.getByTestId(DataTestIds.CONFIG_CONTENT)).toBeInTheDocument(), {
+    timeout: 10000,
```
VikaCep, line 16: Seems like a pretty high number for a timeout, is it really needed?

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+    await userEvent.type(screen.getByLabelText(/Name/), inputValues.name); // Should be transformed to 'new-secret'
+    await userEvent.type(screen.getByLabelText(/Description/), inputValues.description);
+    await userEvent.type(screen.getByLabelText(/Value/), inputValues.plaintext);
+
+    const submitButton = screen.getByText('Save');
+    await userEvent.click(submitButton);
+
+    const { body } = await read();
+    // expect(screen.getByText('Saving...')).toBeInTheDocument();
```
VikaCep, line 113:
```suggestion
```

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementTab.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { server } from 'test/server';
+
+import { SecretsManagementTab } from './SecretsManagementTab';
+
+describe('SecretsManagementTab', () => {
+  it('should render the fallback UI when an error occurs', async () => {
+    server.use(apiRoute('listSecrets', { result: () => ({ status: 500, body: 'Error message' }) }));
+    render(<SecretsManagementTab />);
+    await waitFor(() => expect(screen.queryByTestId(DataTestIds.CENTERED_SPINNER)).not.toBeInTheDocument(), {
+      timeout: 10000,
```
VikaCep, line 15: Same question about this timeout value

VikaCep (changes requested):
Looking great! ⚡🚀 👏 
Left a few comments/questions

src/types.ts
w1kman: If it is a SM specific toggle, my vote is on `synthetic-monitoring-*`.
It makes it VERY clear to others.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
w1kman: Old stuff - Will replace with mocks 👍🏻

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
w1kman: Yeah, the comment was obsolete

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx
w1kman: When running only tests for that file, 30ms is enough (on my machine), when running all the tests, 40ms seems to be the sweet spot.

If I put the CPU under load while running all the tests, there are situations where a timeout of 10s is too little (`CheckList.test.tsx` has a timeout of 10s, and it manages to fail under heavy load).

I don't mind setting it at a "reasonable" value, but it becomes more likely that slow running test (CI etc) will fail.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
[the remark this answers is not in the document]
w1kman: MUCH better.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+  it('should provide correct uuid for clipboard copy', async () => {
+    const user = userEvent.setup();
+    render(<SecretCard {...defaultProps} />);
+
+    // Find the clipboard component
+    const clipboardButton = screen.getByRole('button', { name: /Copy test-secret-1 ID/i }); // "Copy <secret name> ID"
+    await act(() => {
+      return user.click(clipboardButton);
+    });
+    // Instead of accessing props directly, test the actual copy behaviorF
```
VikaCep, line 104:
```suggestion
    // Instead of accessing props directly, test the actual copy behavior
```

VikaCep (approved):
Just suggested a small typo fix, other than that LGTM!

ckbedwell (approved):
LGTM! ㊙️🚀

PR #1135, opened 2025-05-07
fix: update return type of update alerts PUT request

VikaCep:
After a conversation with @ka3de, we agreed the response for the `PUT /check/:id/alerts` endpoint should not return the list of alerts in the response body, as this is consumed from the `GET` endpoint already.
This PR addresses the frontend portion of the changes, while it is also going to be removed on the API side.


ka3de (approved):

ckbedwell (approved):
LGTM

Issue #1141, opened 2025-05-12
Probes don't get populated in HTTP scenes dashboards or interpolated correctly for Explore queries.


ckbedwell:
_Somewhat of a placeholder, will fill this out properly._

Raised in an escalation from [Dell](https://github.com/grafana/support-escalations/issues/16231).

ckbedwell:
Fixed in https://github.com/grafana/synthetic-monitoring-app/pull/1142

PR #1142, opened 2025-05-13
fix: added correctSceneVariableInterpolation function with tests

ckbedwell:
Addresses: https://github.com/grafana/synthetic-monitoring-app/issues/1141

## Problem

1. The probe variable shows unrelated probes
2. When selecting multiple values for probes they don't get parsed when opened up in Explore correctly, instead of `probe=~"probe1|probe2"` we receive `probe=~"{probe1,probe2}"` which is incorrect.

## Solution
For HTTP dashboards we use React scenes and use `useVariableInterpolator` to generate the Explore query and for all the other dashboards we use vanilla scenes so use `sceneGraph.interpolate`.

Following the code in scenes we end up at `sceneInterpolator` and `formatRegistry` and it is difficult to discern if and how we pass a custom format for the variable. For the time-being I have created `correctSceneVariableInterpolation` to wrap the interpolate function and added the appropriate tests.

I'll do some investigation and raise this with the scenes team to see if we can get an upstream fix as a longer-term solution.

**Before**

Explore view showing a query including probe=~'{Chris Mac 2,Chris Mac}' and no data being displayed as the visualisation.

**After**

Explore view showing a query including probe=~'Chris Mac 2|Chris Mac' and data in a time series graph being displayed as the visualisation.

### DIFF
--- a/src/scenes/BROWSER/WebVitals/webVitalGaugeScene.tsx
+++ b/src/scenes/BROWSER/WebVitals/webVitalGaugeScene.tsx
@@ -6,6 +6,7 @@ import { css } from '@emotion/css';
 
 import { WebVitalName } from './types';
 import { DataQueryExtended } from 'scenes/ExplorablePanel';
+import { correctSceneVariableInterpolation } from 'scenes/utils';
 
 import { WebVitalGauge } from './WebVitalGauge';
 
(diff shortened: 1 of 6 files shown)


VikaCep (approved):
LGTM!

PR #1145, opened 2025-05-14
Fix/1137 request error triggers zero state view

### LINKED ISSUE #1137: Request error may trigger zero state view
**What happened**:
I navigated to /a/grafana-synthetic-monitoring-app(/home) and expected to see my checks dashboard.
Instead, I was greeted with "You haven't created any checks yet" (see image).

Image

Behind the scenes, the /sm/check/list endpoint had failed with a 503. Which lead the app to incorrectly assume that I hadn't created any checks yet (list was empty because backend failed to return a response).

I expected to see an error message saying that something went wrong when trying to fetch my checks.

1. Go to `a/grafana-synthetic-monitoring-app`
2. Make sure you have at least one check (if not, create one)
3. Block traffic to `/sm/check/list`
4. Refresh the page
5. See "You haven't created any checks yet"

w1kman:
**What this PR does / why we need it**: To avoid zero state ("You haven't created any checks yet...") to be displayed on failed requests, suspense query wrapped with `QueryErrorBoundary` is used to catch request errors and allow the user to retry the request.

Resolves #1137

Block the request to `sm/check/list` to see the boundary in action.

### DIFF
--- a/src/page/CheckList/CheckList.test.tsx
+++ b/src/page/CheckList/CheckList.test.tsx
@@ -52,564 +52,566 @@ const renderCheckList = async (checks = BASIC_CHECK_LIST, searchParams = '') => 
   return res;
 };
 
-test('renders empty state', async () => {
-  server.use(
-    apiRoute(`listChecks`, {
-      result: () => {
-        return {
-          json: [],
-        };
-      },
-    })
-  );
-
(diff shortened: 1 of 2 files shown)


ckbedwell (approved):
LGTM 🚀

Issue #1129, opened 2025-04-30
Per-check alerts: instrument using Rudderstack


VikaCep:
We should instrument the interactions of the new per-check alerts following the analytics tracking framework introduced in this PR https://github.com/grafana/synthetic-monitoring-app/pull/1114

PR #1146, opened 2025-05-14
feat: add tracking for per-check alerts

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1129

Adds tracking for:
- selecting an alert
- unselecting an alert
- changing its period
- changing its threshold
- successfully creating an alert
- successfully deleting an alert

https://github.com/user-attachments/assets/47929d9d-60f6-440c-8068-028d612f9362
https://github.com/user-attachments/assets/d651b292-ff1f-48ec-a76f-b3d9551c2b10


src/features/tracking/perCheckAlertsEvents.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const perCheckAlertEvents = createSMEventFactory('per_check_alerts');
+
+interface PerCheckAlertSelected extends TrackingEventProps {
+  /** The name of the alert */
+  name: CheckAlertType;
+  /** The threshold of the alert */
+  threshold: number;
+  /** The period of the alert */
+  period?: string;
```
ckbedwell, line 13: I wouldn't track these with this event as they are a red herring in the analytics data -- they might enable the event but then change the values and that wouldn't be reflected accurately.

ckbedwell (changes requested):
I am weighing up my opinion on how granular we should go with our analytics data. I'm wondering if it is worth tracking when a user interacts with the threshold and pending UI components separately or whether that is overkill. Generally speaking more data is better than less as we can always ignore additional data but we can't conjure it up if it is missing...

We might learn if people are unclear what values to select if we see these events very high vs how many checks with alerts are created.

So with that said, I think you should some more tracking events which are also tracking when each alert gets successfully created / deleted.

If we have that data we can start to get an indication how many people use the plugin to create alerts with checks vs using Terraform / API directly 😄

VikaCep:
Thanks for the input @ckbedwell. I've updated the tracking events accordingly and also rebased to branch off from [feat/show-alerts-on-checks-list](https://github.com/grafana/synthetic-monitoring-app/tree/feat/show-alerts-on-checks-list) as we need to compare with the existing check alerts in order to calculate when they are successfully created/deleted

src/components/CheckForm/AlertsPerCheck/TLSTargetCertificateCloseToExpiringAlert.tsx
```diff
@@ -70,6 +76,7 @@ export const TLSTargetCertificateCloseToExpiringAlert = ({
                 data-testid={`alert-threshold-${alert.type}`}
                 onChange={(e) => {
                   const value = e.currentTarget.value;
+                  trackChangeThreshold({ name: alert.type, threshold: value !== '' ? Number(value) : 0 });
```
ckbedwell, line 79: Could be same reference for consistency.

src/components/CheckForm/checkForm.hooks.ts
```diff
@@ -67,7 +67,11 @@ export function useCheckForm({ check, checkType, checkState, onTestSuccess }: Us
     setSubmittingToApi(false);
   };
 
-  const { mutate: updateAlertsForCheck } = useUpdateAlertsForCheck({ onSuccess: navigateToChecks, onError });
+  const { mutate: updateAlertsForCheck } = useUpdateAlertsForCheck({
+    onSuccess: navigateToChecks,
+    onError,
+    prevAlerts: check?.Alerts,
```
ckbedwell, line 73: Concern for higher up?

ckbedwell (changes requested):
Pair review comments 2️⃣

src/components/CheckForm/AlertsPerCheck/TLSTargetCertificateCloseToExpiringAlert.tsx
VikaCep: Modified the threshold type as just `string` so now it just receives the debounced value from the input.

src/components/CheckForm/checkForm.hooks.ts
VikaCep: I looked into moving the `trackAlertCreationsAndDeletions` call out of the `useUpdateAlertsForCheck` hook, but I found doing that adds some complexity. To call the tracking function from the outer scope, I would need access to both the `prevAlerts` and the `newAlerts` that are passed to the mutation.

The issue is that while I have `prevAlerts` available at this level, the new alerts only exist inside the mutation, so passing both around just for tracking makes the flow messier than it needs to be.

Also, since the tracking is directly tied to the alert update, I think it makes sense to keep it in the `useCheckAlerts` hook. It keeps the alerts logic and its side effects in one place, which feels easier to maintain.

That said, let me know if you see it differently and we can figure out another approach :)

src/components/CheckForm/useDebouncedCallback.ts
ckbedwell, line 1: I feel a bit guilty for leaving this comment as you so nicely added tests but we have a utility package called `usehooks-ts` that has this function ready for us to use already 😬 : https://usehooks-ts.com/react-hook/use-debounce-callback

src/components/CheckForm/AlertsPerCheck/TLSTargetCertificateCloseToExpiringAlert.tsx
```diff
@@ -70,6 +81,7 @@ export const TLSTargetCertificateCloseToExpiringAlert = ({
                 data-testid={`alert-threshold-${alert.type}`}
                 onChange={(e) => {
                   const value = e.currentTarget.value;
+                  debouncedTrackChangeThreshold(alert.type, value);
```
ckbedwell, line 84: I'd add a fairly significant delay to these that wasn't the default 300. Maybe something closer to 750 - 1000?

As it's tracking and not user-facing having a delay that large doesn't break the conventional wisdom of keeping these things to 300ms or under.

ckbedwell (changes requested):
Last few small changes and we'll get this in today!

src/components/CheckForm/useDebouncedCallback.ts
VikaCep: Oops, totally missed that. I've changed it to use that one instead :)

ckbedwell (approved):
LGTM!

Issue #1131, opened 2025-04-30
Per-check alerts: display visual clue when there are errors


VikaCep:
The SM app should consume the `GET /check/:id/alert` endpoint with the `?alerts=true` param to expose information on the alert creation status and give a visual clue when there is an error. 

Quoting https://github.com/grafana/synthetic-monitoring/issues/266:

> The backend registers the possible errors returned from Hosted Grafana. We should expose these errors to the plugin, so the plugin can take different actions in order to try to provide feedback to the user on that situation.

Note:
This [PR](https://github.com/grafana/synthetic-monitoring-api/pull/1331) implements the new flag that's needed obtain this information.

ka3de:
Notice that there are two ways to retrieve this information:
- With the endpoint for a particular check alerts.
- With the endpoint to list checks, including alerts.

**The first case** corresponds to `GET /check/:id/alert` endpoint, which will always return the alert status such as:
```
{
  "name": "TLSTargetCertificateCloseToExpiring",
  "threshold": 60,
  "period": "",
  "status": "PENDING_CREATE",
  "error": "some error,",
  "created": ...,
  "modified": ...
}
```
See more details, including the specific errors currently supported, in [this comment](https://github.com/grafana/synthetic-monitoring/issues/266#issuecomment-2769263326).

**The second case** corresponds to `GET /check/list?includeAlerts` endpoint, which extends the list checks endpoint to include also the alerts data. I've changed the initially considered `?alerts=true` for `?includeAlerts=true` as I thought this was more "expressive", but we are still in time to get feedback on that.

This endpoint called with the alerts query param will return a response such as:
<details>
<summary>this one</summary>

```json
[
	{
		"id": 1,
		"tenantId": 1,
		"frequency": 60000,
		"offset": 0,
		"timeout": 3000,
		"enabled": true,
		"labels": [],
		"settings": {
			"http": {
				"ipVersion": "V4",
				"method": "GET",
				"noFollowRedirects": false,
				"failIfSSL": false,
				"failIfNotSSL": false
			}
		},
		"probes": [
			1
		],
		"target": "https://test1.com",
		"job": "test1",
		"basicMetricsOnly": true,
		"alertSensitivity": "none",
		"created": 1745317686.5357733,
		"modified": 1745317687.2077816,
		"Alerts": [
			{
				"name": "TLSTargetCertificateCloseToExpiring",
				"threshold": 500,
				"period": "",
				"status": "OK",
[...]

PR #1148, opened 2025-05-16
feat: add indicator when there is an error with per-check alerts

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1131

With the new per-check alerts, the creation request can be successful but there can still be asynchronous errors that need to be displayed to the user. This PR addresses this requirement. 

From https://github.com/grafana/synthetic-monitoring/issues/266:

>Alert rules creation process is performed asynchronously from the HTTP request that triggers its creation. This means that the user might have received a successful response (202 HTTP Accepted), but the rule itself might not end up being created (for example due to quota limits or other possible errors).
>The backend registers the possible errors returned from Hosted Grafana. We should expose these errors to the plugin, so the plugin can take different actions in order to try to provide feedback to the user on that situation. Possible methods discussed are:

- Error indication on the check list:
image

- Error indication in the check form - Alerting section:
image

NOTE: in order to see the error messages, the app can be started using `yarn dev:msw` where there is a check (`Job name for http`) that provides an error for the `TLSTargetCertificateCloseToExpiring` alert

### DIFF
--- a/src/components/AlertStatus/AlertStatus.tsx
+++ b/src/components/AlertStatus/AlertStatus.tsx
@@ -138,6 +138,7 @@ const TooltipContent = ({
         {({ isEnabled }) =>
           isEnabled ? (
             <PerCheckAlertGroups
+              alerts={check.Alerts}
               groups={perCheckGroups}
               loading={perCheckGroupsLoading}
               isError={perCheckGroupsError}
--- a/src/components/AlertStatus/GrafanaNamespaceAlertRuleDisplay.tsx
+++ b/src/components/AlertStatus/GrafanaNamespaceAlertRuleDisplay.tsx
@@ -39,14 +39,14 @@ export const GrafanaNamespaceAlertRuleDisplay = ({ group, alerts }: AlertRuleDis
               <Stack gap={1} alignItems="center" justifyContent="space-between">
                 <Icon name="corner-down-right-alt" />
                 <span>{rule.name}</span>
-                <LinkButton
+                {rule.uid && <LinkButton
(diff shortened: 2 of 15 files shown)


ka3de:
I was a little bit confused because I see that we are showing the error next to the alert in the "alerts pop-up". But IIRC, this alerts pop-up is populated from the alerts that are already created, because these are queried through some alerting endpoint.

So how are we doing that for the case on which an alert has been set up for a check, but this one is not created in the stack, for example due to a _quota limit_ error?
The folder name is static for all SM alerts, and the check alerts data provides the alert name and period, but the group name is still missing... I believe we have a mapping for alert name - group name in the plugin which we use, along with the period (if applies), to build the group name and show all that information even if the rule does not exist?

VikaCep:
You have a good point. I made some changes so now the UI always displays per-check alerts, even if the corresponding alert rule does not exist in the alerting endpoint response. 
Basically, for each alert configured on a check, it now attempts to find the matching rule in the alerting response. If it is missing, a local “dummy” group is constructed with the appropriate category and default values, ensuring the alert (and any error/status) is still shown in the alerts pop-up. The folder UID is static for all SM alerts, and the group/category is determined from the predefined alert mapping.

In the screenshot below, the `TLSTargetCertificateCloseToExpiring` alert is shown from the `alerts` API response as it's not returned from the alerting endpoint. I just had to hide the "view" icon link as there is no place to link it if it doesn't exist. Does this sound good to you?

image

ckbedwell (approved):
LGTM

PR #1151, opened 2025-05-23
fix: correct tracking for update / create check events

ckbedwell:
Comments inline for reasoning.

### DIFF
--- a/src/components/CheckForm/checkForm.hooks.ts
+++ b/src/components/CheckForm/checkForm.hooks.ts
@@ -99,6 +99,10 @@ export function useCheckForm({ check, checkType, checkState, onTestSuccess }: Us
           result = await createCheck(newCheck);
         }
         await handleAlertsAndNavigate(result, alerts);
+      } catch (e) {
+        // swallow the error
+        // it gets handled correctly by the the generic hooks and we have tests to prove that
+        // this isn't strictly necessary but jest complains about this...
       } finally {
         setSubmittingToApi(false);
       }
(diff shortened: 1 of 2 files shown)


src/components/CheckForm/checkForm.hooks.ts
```diff
@@ -64,31 +64,24 @@ export function useCheckForm({ check, checkType, checkState, onTestSuccess }: Us
   const navigateToChecks = useCallback(() => navigate(AppRoutes.Checks), [navigate]);
   const alertsEnabled = useFeatureFlag(FeatureName.AlertsPerCheck).isEnabled;
 
-  const onUpdateCheckComplete = useCallback(() => {
-    trackCheckUpdated({ checkType });
```
ckbedwell, line 68: This function runs when both a check is created AND updated so the tracking event was wrong. I realised the tracking is a concern for the higher order hooks so moved it there instead.

src/components/CheckForm/checkForm.hooks.ts
```diff
@@ -64,31 +64,24 @@ export function useCheckForm({ check, checkType, checkState, onTestSuccess }: Us
...
     prevAlerts: check?.Alerts,
   });
 
   const handleAlertsAndNavigate = useCallback(
     async (result: Check, alerts?: CheckAlertFormRecord) => {
       try {
-        if (!alerts && result.id) {
-          return navigateToChecks();
-        } else {
+        if (alerts) {
```
ckbedwell, line 74: I inverted this so the flow is easier to follow - if there are alerts add this extra step then do the navigation, rather than if there isn't this do navigation, but if there is this do alerts AND navigation.

VikaCep (approved):

Issue #1149, opened 2025-05-19
feat: Support TLS cert expiry alert for TCP checks


VikaCep:
The same metric exported for HTTP checks is also applicable for TCP checks. Notice that the TLS flag must be explicitly enabled in the TCP check settings in order to provide the TLS certificate metrics that are evaluated in the certificate expiry alert rule.

Frontend part of https://github.com/grafana/synthetic-monitoring/issues/298

PR #1153, opened 2025-05-23
feat: add TLS Cert Expiry alert for TCP checks

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1149
Depends on https://github.com/grafana/synthetic-monitoring-api/pull/1357

### Summary

This PR introduces support for the "TLS certificate close to expiring" alert for TCP checks in the synthetic monitoring app.

### Details

- **Feature:**  
  - Users can now enable a per-check alert for when the target’s TLS certificate is close to expiring on TCP checks.
- **UI/UX:**  
  - The alert option is only available if TLS is enabled in the TCP check’s "TLS Config" section.
  - If TLS is not enabled, the alert checkbox is disabled and a warning message is displayed to inform the user that TLS must be enabled for this alert to function.
image

- **Tests:**  
  - Added tests to verify:
    - The alert can be enabled and submitted when TLS is enabled.
    - The alert is disabled and a warning is shown when TLS is not enabled.

https://github.com/user-attachments/assets/5d793440-1348-49a6-a17d-6653fb16ece1


ka3de:
> However, I don't know if we want to include it in the public preview. Do we need to update the docs for it @ka3de?

I would say this is not a must, but whenever it's ready we can merge it. Docs are updated in [this PR](https://github.com/grafana/website/pull/25995) which is ready to merge when the functionality is released.

src/page/NewCheck/__tests__/ApiEndPointChecks/TCPCheck/4-alerting.payload.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { goToSection, renderNewForm, submitForm } from 'page/__testHelpers__/checkForm';
+
+import { fillMandatoryFields } from '../../../../__testHelpers__/apiEndPoint';
+
+const checkType = CheckType.TCP;
+
+describe(`TCPCheck - Section 4 (Alerting) payload`, () => {
+  it(`can add TLS certificate expiry alert when TLS is enabled`, async () => {
+    jest.replaceProperty(config, 'featureToggles', {
+      // @ts-expect-error
```
w1kman: Why?

w1kman (approved):
LGTM - I'm not sure about the orange/warning color for the text.
Seems a bit much. Perhaps double check with @vesnadean ?

src/page/NewCheck/__tests__/ApiEndPointChecks/TCPCheck/4-alerting.payload.test.tsx
VikaCep: I've removed the error suppressions [here](https://github.com/grafana/synthetic-monitoring-app/pull/1153/commits/5947f487812588bb216b9847ba64aca203eeadec) and created a utility function to override feature flags in tests 👍

VikaCep:
Thanks @vesnadean! I took another look, and since we don’t want to allow saving the alert when TLS is off, I went ahead and added a validation rule to handle that case, to be consistent with other invalid scenarios:

Recording:
https://github.com/user-attachments/assets/2806aa6e-8ab3-4945-9659-3e8f364c4f44

image

This should help in cases where someone creates a check with the alert enabled and then later disables TLS by preventing that mismatch from being saved.

If we’d rather not go with validation, we could fall back to just showing the message in a secondary color like you suggested:

image

Let me know which alternative you think is best 🙏

vesnadean:
@VikaCep The recording looks great, let's go with that option.

w1kman (approved):

PR #1161, opened 2025-06-16
fix: change reachability description for probes view

VikaCep:
For probe-related components, the reachability description we are currently displaying can be confusing (see this [Slack](https://raintank-corp.slack.com/archives/C0162C1K9E1/p1750087772831939) conversation):
> The reachability info popup states "The success rate of all the probes. Reachability decreases when any probe fails.", which is a little misleading when viewed in context of the probe list. "The success rate of all synthetic checks run by this probe" might be a little more clear, since the language implies the probe is failing and not the checks.

In this PR, I made the following changes:

- Added a new `PROBE_REACHABILITY_DESCRIPTION` constant with the wording "The success rate of all synthetic checks run by this probe."
- Updated the probe-related components (ProbeCard, ProbeStatus) to use this new description for the reachability gauge, making the meaning clearer in the context of individual probes.

image
image

The original `REACHABILITY_DESCRIPTION` remains unchanged and is still used in other contexts where it refers to all probes.

image
image

This approach ensures the description is accurate and context-specific without affecting other parts of the app that use the original description.

### DIFF
--- a/src/components/constants.ts
+++ b/src/components/constants.ts
@@ -514,7 +514,7 @@ export const UPTIME_DESCRIPTION =
   'Whether any of the probes could reach the target. Uptime decreases when all the probes fail simultaneously.';
 export const REACHABILITY_DESCRIPTION =
   'The success rate of all the probes. Reachability decreases when any probe fails.';
-export const PROBE_REACHABILITY_DESCRIPTION = 'The success rate of all synthetic checks run by this probe.';
+export const PROBE_REACHABILITY_DESCRIPTION = 'The success percentage of all checks performed by this probe.';
 export const LATENCY_DESCRIPTION =
   'The average time to receive an answer across all the checks during the whole time period.';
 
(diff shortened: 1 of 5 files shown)


src/components/constants.ts
```diff
@@ -514,6 +514,7 @@ export const UPTIME_DESCRIPTION =
   'Whether any of the probes could reach the target. Uptime decreases when all the probes fail simultaneously.';
 export const REACHABILITY_DESCRIPTION =
   'The success rate of all the probes. Reachability decreases when any probe fails.';
+export const PROBE_REACHABILITY_DESCRIPTION = 'The success rate of all synthetic checks run by this probe.';
```
w1kman: I would probably change the copy slightly (I'm NOT a native English speaker|writer, so take my advice with a grain of salt).

Something like: "The success percentage of all checks performed by this probe" 🤷🏻

Then again, we use "rate" all over the place, so that it would be a little bit of an outlier

w1kman (dismissed):
LGTM

I feel that the `PROBE_REACHABILITY_DESCRIPTION` copy could be improved. If you don't share that opinion, feel free to keep it as it is 👍🏻

src/components/constants.ts
VikaCep: I like your suggestion so I applied it :)

w1kman (approved):

Issue #1133, opened 2025-05-07
Check creation right aside


w1kman:
**What would you like to be added**:
A UI to the right of the check creation form (aka "right aside"). It will allow us to show useful information next to the check form.

The first tenant of the right aside should be ad-hoc test result.

Key features compared to the current ad-hoc UI:
- Possibility to view ad-hoc result whilst still be able to navigate the check form
- Possibility to view the history of ad-hoc check results

**Why is this needed**:
Shorter feedback loop and ease of use, when it comes to using ad-hoc check result during check authoring.

**Additional context**
Part of https://github.com/grafana/synthetic-monitoring-app/issues/1077

PR #1170, opened 2025-06-30
feat: refactor `<CheckForm/>`

### LINKED ISSUE #1165: Refactor Check creation/edit form component
Separate (split into smaller pieces) the check form from the PluginPage (and other presentational stuff), to prepare for the new check creation/edit flow.

**Why is this needed**:
The check form component is a page component (basically the whole page), which makes it hard to 
maintain and test.

**Additional context**
Needed for https://github.com/grafana/synthetic-monitoring-app/issues/1133

w1kman:
The goal for this PR is to split the actual form (+internals) from the page component, in order to prepare for future improvements of the check authoring flow (See #1133).

The check form component is a page component (basically the whole page), which makes it hard to
maintain and test.

Resolves #1165 

Even though tests are passing, try to spend some time in creating and editing checks, to make sure everything works as it should.

### DIFF
--- a/src/components/CheckForm/CheckForm.tsx
+++ b/src/components/CheckForm/CheckForm.tsx
@@ -40,6 +40,8 @@ interface CheckFormProps extends PropsWithChildren {
 }
 
 export function CheckForm({ check, disabled }: CheckFormProps) {
+  // If the context is not available, we create a new one.
+  // This allows the CheckForm to be used both as a standalone component and within the CheckFormContextProvider.
   const context = useContext(CheckFormContext);
   if (!context) {
     return (
--- a/src/page/CreateCheck/CreateCheck.tsx
+++ b/src/page/CreateCheck/CreateCheck.tsx
@@ -6,7 +6,6 @@ import { css } from '@emotion/css';
 
 import { createNavModel } from 'utils';
 import { CheckForm } from 'components/CheckForm/CheckForm';
-import { CheckFormContextProvider } from 'components/CheckForm/CheckFormContext';
 
 export function CreateCheck() {
   const styles = useStyles2(getStyles);
@@ -20,11 +19,9 @@ export function CreateCheck() {
 
   return (
     <PluginPage pageNav={pageNavModel}>
-      <CheckFormContextProvider>
-        <div className={styles.wrapper}>
-          <CheckForm />
-        </div>
-      </CheckFormContextProvider>
+      <div className={styles.wrapper}>
+        <CheckForm />
+      </div>
     </PluginPage>
   );
 }


src/components/CheckEditor/CheckProbes/ProbesList.tsx
```diff
@@ -12,11 +12,13 @@ export const ProbesList = ({
   probes,
   selectedProbes,
   onSelectionChange,
+  disabled,
 }: {
   title: string;
   probes: ProbeWithMetadata[];
   selectedProbes: number[];
   onSelectionChange: (probes: number[]) => void;
+  disabled?: boolean;
```
VikaCep, line 21: Can't you read the `disabled` value here from `formState.disabled` instead of receiving it as a prop as you do in other places?

src/components/CheckForm/CheckForm.tsx
```diff
@@ -71,39 +34,42 @@ const checkTypeStep1Label = {
...
   check?: Check;
   disabled?: boolean;
-};
+}
+
+export function CheckForm({ check, disabled }: CheckFormProps) {
+  const context = useContext(CheckFormContext);
+  if (!context) {
+    return (
+      <CheckFormContextProvider check={check} disabled={disabled}>
```
VikaCep, line 48: I found this confusing, as there's another `CheckFormContextProvider` defined both in `CreateCheck` and `EditCheck` that wraps the `CheckForm`. Do we need to the provider here as well?
Also, if we set disabled = `true` at this level, it doesn't take effect. We need to set it in the outer context provider in order for it to work.

src/components/CheckForm/CheckForm.hooks.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  // Hook usage
+  const checkType = useFormCheckType(check);
+  const checkTypeGroup = useFormCheckTypeGroup(check);
+  const schema = useCheckFormSchema(check);
+  const options = useCheckTypeOptions();
+  const isOverLimit = useIsOverlimit(!isNew, checkType);
+  const permission = useFormPermissions();
+  const defaultFormValues = useCheckFormDefaultValues(check);
+  const isExistingCheck = getIsExistingCheck(check);
```
VikaCep, line 68: Why not just use `isNew`? (or `!isNew`)

VikaCep (commented):
Nice work! I tested it and works great. Left a few questions/suggestions.

src/components/CheckEditor/CheckProbes/ProbesList.tsx
w1kman: `ProbeList` is a child of `CheckProbes`, which is a child of `ProbeOptions`.
`ProbeOptions` is using context to pass down whether or not the field is disabled.

Making `ProbList` and/or `CheckProbes` depend on the context seems like it would only cause unneeded dependencies.

Unless you have a compelling argument for using context in this case, I'm listening :)

src/components/CheckForm/CheckForm.hooks.ts
w1kman: > `getIsExistingCheck, // use this for type narrowing of check param`

It's so that TypeScript understands that the `check is Check`.

src/components/CheckForm/CheckForm.tsx
w1kman: I've removed the wrapping context for `CreateCheck` for this PR to make more sense.
I've also updated to comment.

> if we set disabled = true at this level, it doesn't take effect. We need to set it in the outer context provider in order for it to work.

This is a conditional context. Trying to set any prop on the provider won't do anything if a context already exists. That's by design.

src/components/CheckEditor/CheckProbes/ProbesList.tsx
VikaCep: Not really, I just saw in other places we did that and my main motive was reducing the number of props, but not a big deal 👍

PR #1079, opened 2025-02-24
feat: migrate HTTP dashboard to react-scenes

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1076

Migrates the HTTP check's dashboard to use [scenes-react](https://github.com/grafana/scenes/pull/734) as POC.

image

Issue #1076, opened 2025-02-24
POC React-Scenes in SM


ckbedwell:
## Problem

In SM we find Scenes quite burdensome to write and often blocks our ability to innovate and iterate on new ideas we have for features. The API is often difficult to reason with as your mental model for how React works does not align very well with the way that scenes is currently written.

## Acceptance criteria

Before exploring alternative ideas (such as potentially dropping Scenes altogether... 🙈 ), we want to see if the proposed [React-Scenes feature](https://github.com/grafana/scenes/issues/761) is a viable solution addressing our developer concerns for managing a scenes-heavy application.

- [x] Create a POC branch rewriting a **single** dashboard (for the POC do not do multiple dashboards unless it feels warranted).
- [x] Write a short report of the findings (e.g. pros and cons) when converting a dashboard from the current Scenes API to the React-based one.

The outcome of this POC should be an honest evaluation of if we should continue to invest in Scenes or explore alternative approaches which align better with our goals.

### A running list of some of the challenges we face using Scenes right now (please update as more issues are found):
 
- Change the query based on the time range selected (e.g. under 768 hours use a Mimir instant query, above 768 hours use a Mimir range query)
- Query management. 
  - e.g. right now I (@ckbedwell) am exploring a custom logs visualisation and keep thinking how the Loki query could be 'lifted' and provide additional context to all of the other panels on the dashboard. I know that if I did this in `react-query` the query provider automatically provides it to any other component that would wish to use it. I don't know what that workflow looks like in Scenes? Lifting the QueryRunner? Should all the QueryRunners be lifted to the outermost parent?

VikaCep:
I've created a migration of the HTTP Dashboards to React Scenes in this draft PR https://github.com/grafana/synthetic-monitoring-app/pull/1079

And added my comments in this document https://docs.google.com/document/d/1XO9DjyAMGmG4H9eUDct6S_jBiMzX3dljqEDV5l45iMM/edit?tab=t.864k6tr14wge

PR #1177, opened 2025-07-07
Feat/scenes react script browser

ckbedwell:
## Problem

Some of our dashboards still use the standard scenes API which after the great work done in the[ React scenes POC](https://github.com/grafana/synthetic-monitoring-app/issues/1076) and the [success of migrating the HTTP dashboard](https://github.com/grafana/synthetic-monitoring-app/pull/1079), we need to do the same for the rest of our check types.

## Solution

This PR converts the Scripted and Browser check dashboards to use React Scenes.

A few things to highlight that this PR adds:

1. Move extensive use of the `src/queries` folder. The aim of this folder is this is a standardised location where all of our collected PromQL will live.
2. Created  `<DashboardContainer />` which extracts all of the common components for a React scenes dashboard
3. Made the Table component compatible with the experimental themes which Grafana now provides:

**Before**
Screenshot of Synthetic Monitoring Scripted Dashboard in Grafana tron theme. Two of the tables have the wrong colour background, which is based on the traditional dark theme.

**After**
Screenshot of Synthetic Monitoring in Grafana tron theme. The tables now have the correct background colour which matches the rest of the Tron theme.


VikaCep:
Amazing work! A few things I noticed from comparing both versions:

**Scripted checks**

- Distinct targets and duration by probe visualizations differ a bit between the old and new versions

Before
image
After
image

- The threshold for uptime seems to be different, it's showing red locally while in dev it's green for the same value

Before
image
After
image

**Browser checks**

- CLS visualizations shouldn't have a time unit
image

- When there's no value in the slider visualizations, we're showing 0 instead of empty

Before
image
After
image

src/scenes/HTTP/HttpDashboard.tsx
```diff
@@ -1,148 +1,60 @@
...
-                  </AnnotationLayer>
-                </AnnotationLayer>
-              </Stack>
-            </PluginPage>
-          </CustomVariable>
-        </CustomVariable>
-      </QueryVariable>
-    </SceneContextProvider>
+        <div className={styles.errorLogs}>
+          <ErrorLogs startingUnsuccessfulOnly={true} />
```
VikaCep, line 54:
```suggestion
          <ErrorLogs />
```

The old version shows all logs and not just unsuccessful ones by default. Not sure if this was on purpose.

ckbedwell: Yeah, this was on purpose to standardise them. I found it annoying that the behaviour was different between the API Endpoint checks and scripted / browser.

VikaCep: For Browser and Scripted checks I see it shows all logs, not just failed ones though. IMO seeing all logs can still be useful so I'd leave it unchecked. 
Not that big of a deal anyway as this is soon going to change with your custom logs work.

VikaCep (changes requested):
Another small thing I noticed, for scripted and browser checks we're missing the contextual menu to explore the query for the `Success rate by probe for is status 200
` visualization (under Assertions):

local:
image

dev:
image

Issue #1159, opened 2025-06-09
Synthetic Check Search breaks with invalid regex


The-9880:
Follows from an escalation raised [here](https://github.com/grafana/support-escalations/issues/16557).

I think what's happening is the search bar is treating the input as regex instead of maybe doing a `contains` string match. The error gets returned for `*` because on it's own that is invalid (you would write `.*`, which does work).

Erroring out the search bar results in locking down the checks page with the `Something went wrong` error message.

peterschretlen:
(Internal-only link) this was raised in https://github.com/grafana/support-escalations/issues/17191

PR #1178, opened 2025-07-08
fix: updated check list search

ckbedwell:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/1159

## Problem

The checklist search currently crashes the page when a special character (specifically ones which are used in regex pattern matching) begins the search.

## Solution

Change the matching implementation so it is literal rather than using regex under the hood. I've added extensive tests to ensure it works correctly and doesn't regress.


VikaCep (approved):
LGTM ✅

PR #1179, opened 2025-07-08
Per-check alerts: add link to notification policies

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring/issues/316

image

Adds a link to notification polices so users can configure where their alerts will be routed.


ckbedwell (approved):
Was interested to see if we could improve the UI for this by piggy-backing off the API in alerting.

Just grafted this branch to ops and viewed the notification policies there and I've changed my mind and like what you've done here as a starting point 😂

I'll chat to @markjmeier about improving this further in the future. 

LGTM

ka3de:
I missed this issue, but just wanted to say that I really like where this is going! :raised_hands:

PR #1185, opened 2025-07-10
fix: restrict environment variable

VikaCep:
Restricts the environment variable for the `call_deploy-plugin.yml` workflow


ckbedwell (approved):
🔥 🔥 🔥

PR #1186, opened 2025-07-10
GHA: a few improvements

VikaCep:
**Changes GAR authentication job to prevent logging an error**

The `login-to-gar` shared action was failing with s HTTP 404 error during service account impersonation. The issue was caused by:
Missing project ID resolution (notice the `-` in `projects/-/serviceAccounts/...`)
The service account impersonation was failing during the first authentication attempt.

The action would eventually succeed via its fallback mechanism, making the job successful, but caused confusing error logs. To prevent it, I replaced the failing `login-to-gar` action with Direct Workload Identity Federation using `google-github-actions/auth` (https://github.com/grafana/deployment_tools/tree/ece2dc51a27a36d68bc75543f1e40f6e2d78667c/terraform/projects/grafanalabs-workload-identity#using-google-github-actionsauth-directly)

Also, in this PR I changed the environment input from the Deploy (dispatch) manual workflow to be a dropdown with 3 possible values: `dev`, `staging` and `production` instead of free text.

### DIFF
--- a/.github/workflows/call_update-deployment-tools.yml
+++ b/.github/workflows/call_update-deployment-tools.yml
@@ -35,15 +35,6 @@ jobs:
             GITHUB_APP_INSTALLATION_ID=updater-app:app-installation-id
             GITHUB_APP_PRIVATE_KEY=updater-app:private-key
 
-      # needed?
-
-      # - name: Export GitHub Actions URL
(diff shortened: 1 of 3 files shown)


ckbedwell (approved):
LGTM 🌟

Issue #1132, opened 2025-04-30
Per-check alerts: display pending/firing status on checks' dashboards


VikaCep:
We should integrate the new per-check alerts with the firing/pending overlay visualizations on the checks' dashboards. 

![Image](https://github.com/user-attachments/assets/a66e8c78-6655-48a7-88f9-380ead97ac72)

ka3de:
AFAIK for this we depend on the alerting team providing us similar metrics to the ones we currently use. This is tracked in [this document](https://docs.google.com/document/d/1QAmSDNtmrDexchXE_DePTod4NSYTWMKbtchVGAAfg-s/edit?tab=t.1uavs3j4y9br#heading=h.bo1nslbzzuis) and it is currently listed as _"In Progress"_.

ka3de:
FYI @VikaCep: There is an [issue](https://github.com/grafana/alerting-squad/issues/1024) and a [design document](https://docs.google.com/document/d/1GuWUSBM6uBrSXSgiWayOlviDUtU-CKAkCiyJFD3PPQA/edit?tab=t.g93f12lygwz3#heading=h.5sybau7waq2q) for this now.

VikaCep:
This is now available in dev & ops, so should be available soon - example dashboard using the metrics:  [https://ops.grafana-ops.net/d/vfUp2mrMz2/alerts?orgId=1&from=now-30m&to=now&timezone=utc&refresh=30s](https://www.google.com/url?q=https://ops.grafana-ops.net/d/vfUp2mrMz2/alerts?orgId%3D1%26from%3Dnow-30m%26to%3Dnow%26timezone%3Dutc%26refresh%3D30s&sa=D&source=docs&ust=1750773151049113&usg=AOvVaw3hP9Gz-Pgt3Y2RL96r-Pxe)

PR #1187, opened 2025-07-11
feat: add GRAFANA_ALERTS to annotations for firing/pending alerts

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1132

Since the alerting team has added the new `GRAFANA_ALERTS` metric to obtain the `alertstatus` of the alerts, we can now use it to display annotations. I've included it to the current `firing`/`pending` annotations in the checks dashboards. 

**Note:** As of today, this metric is only available in prod for _free_ users (see https://github.com/grafana/alerting-squad/issues/1024) but its absence does now cause any issue with current annotations as the queries check for `ALERTS or GRAFANA_ALERTS`

I have tested these annotations using a free account with a custom dashboard and it shows as expected:

image
image
image
image

### DIFF
--- a/src/scenes/Common/alertAnnotations.ts
+++ b/src/scenes/Common/alertAnnotations.ts
@@ -1,6 +1,9 @@
 import { dataLayers, SceneDataLayerSet } from '@grafana/scenes';
 import { DataSourceRef } from '@grafana/schema';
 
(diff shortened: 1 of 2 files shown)


ckbedwell (approved):
LGTM ⭐

PR #1188, opened 2025-07-11
Feat: add new latency alerts

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring/issues/321

**HTTP**
image

**DNS**
image

**Ping**
image

Depends on https://github.com/grafana/synthetic-monitoring-api/pull/1457

### DIFF
--- a/src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
+++ b/src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
@@ -81,7 +81,57 @@ export const GLOBAL_PREDEFINED_ALERTS: PredefinedAlertInterface[] = [
   },
 ];
 
-export const HTTP_PREDEFINED_ALERTS: PredefinedAlertInterface[] = [TLS_TARGET_CERTIFICATE_CLOSE_TO_EXPIRING_ALERT];
+export const HTTP_REQUEST_DURATION_TOO_HIGH_AVG_ALERT: PredefinedAlertInterface = {
+  type: CheckAlertType.HTTPRequestDurationTooHighAvg,
+  name: 'HTTP Request Duration Too High Avg',
+  description: 'The average HTTP request duration for this check',
+  query: `((sum by(instance, job)(rate(probe_all_duration_seconds_sum[%s])) / sum by(instance, job)(rate(probe_all_duration_seconds_count[%s]))) * 1000 >= sum by (instance, job) (sm_alerts_threshold_http_request_duration_too_high_avg{period="%s"})) * on (instance, job) group_right() max without(probe, region, geohash) (sm_check_info)`,
+  supportsPeriod: true,
+  unit: 'ms',
+  category: CheckAlertCategory.Latency,
+  defaultValues: {
+    threshold: 100,
+    period: '5m',
+    isSelected: false,
+  },
+};
+
+export const PING_REQUEST_DURATION_TOO_HIGH_AVG_ALERT: PredefinedAlertInterface = {
+  type: CheckAlertType.PingRequestDurationTooHighAvg,
+  name: 'Ping Request Duration Too High Avg',
+  description: 'The average ping request duration for this check',
+  query: `((sum by(instance, job)(rate(probe_all_duration_seconds_sum[%s])) / sum by(instance, job)(rate(probe_all_duration_seconds_count[%s]))) * 1000 >= sum by (instance, job) (sm_alerts_threshold_ping_request_duration_too_high_avg{period="%s"})) * on (instance, job) group_right() max without(probe, region, geohash) (sm_check_info)`,
+  supportsPeriod: true,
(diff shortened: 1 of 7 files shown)


src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
```diff
@@ -81,7 +81,84 @@ export const GLOBAL_PREDEFINED_ALERTS: PredefinedAlertInterface[] = [
...
+  ) * on (instance, job) 
+  group_right() 
+  max without(probe, region, geohash) (
+    sm_check_info{instance="$instance", job="$job"}
+  )`,
+  supportsPeriod: true,
+  unit: 'ms',
+  category: CheckAlertCategory.Latency,
+  defaultValues: {
+    threshold: 100,
```
ckbedwell:
```suggestion
    threshold: 300,
```

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
```diff
@@ -81,7 +81,84 @@ export const GLOBAL_PREDEFINED_ALERTS: PredefinedAlertInterface[] = [
...
```
ckbedwell:
```suggestion
    threshold: 50,
```

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
```diff
@@ -81,7 +81,84 @@ export const GLOBAL_PREDEFINED_ALERTS: PredefinedAlertInterface[] = [
...
+  query: `(
+    (
+      sum by(instance, job)(rate(probe_all_duration_seconds_sum{instance="$instance", job="$job"}[$period])) / 
+      sum by(instance, job)(rate(probe_all_duration_seconds_count{instance="$instance", job="$job"}[$period]))
+    ) * 1000 >= $threshold
+  ) * on (instance, job) 
+  group_right() 
+  max without(probe, region, geohash) (
+    sm_check_info{instance="$instance", job="$job"}
+  )`,
```
ckbedwell, line 97: @ka3de -- I know this might be a bit late raising this as we use a similar query structure for the other alerts but is the intention of this to reduce the query to a single series? (I'm basing this off stripping the probe, region and geohash labels).

I noticed when I have additional labels for my private probes, as they are on the `sm_check_info` metric that ends up creating additional series in the result which seems undesired?

ckbedwell (changes requested):
Very minor changes for the default values for the thresholds so they are more in line with what I'd expect for each check type.

I've also [raised a question in the Slack channel](https://raintank-corp.slack.com/archives/C075N182HFW/p1753868585510189) concerning the description names but otherwise LGTM and ready to approve shortly :D

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.constants.tsx
ka3de: Our initial alerts try to stick to the "canonical" SM use case, that is alert on the aggregation of the data provided by all the probes on which the check runs. That is applied for the _failed executions_ alert, as well as the _latency_ alerts. Based on that, the left hand side of the alert rule query will always produce a value that is aggregated across all probes, therefore when we join that with the _check_info_ metric in order to include other relevant labels and any custom label that the user set we have to strip any label that is dependent on the probe itself and just use the common ones for a single time series result.
Let me know if I understood it correctly and this answers the question.

ka3de: That does not mean that in the future we might support other type of alerts which can fire based on the metrics for a single probe/per probe. But IMO that sticks less to the "canonical usage" of SM, therefore we've deferred that to a possible future implementation if we have enough users that want this. Notice that this would have to be a different alert rule anyway.

ka3de: > I noticed when I have additional labels for my private probes, as they are on the sm_check_info metric that ends up creating additional series in the result which seems undesired?

Ah, wait, I might be misunderstanding part of your question. So what you are saying is that custom labels in the probe will impact the result of the alert instance? Mmm.. right, that is something that I don't think we are handling perfectly.. I see that, as you say, if there is a private probe with custom labels this will lead to multiple alert instances, which is not ideal...
Unfortunately we can not use `max by` because we would strip the custom check labels also, so I'm not sure there is a way we can keep the check labels, which are important for routing, but strip the probes ones. I'll discuss this with other squad members, but I don't see a way to remove that without removing also the custom check labels.

ckbedwell: Okay, no problem. Just wanted to raise it as I noticed it when I was playing around with the feature today.

PR #1158, opened 2025-06-05
feat: pr creation GH action consolidation test

ckbedwell:
GH actions set-up. Ignore the commits, this will undoubtedly get painful... 😰 

Closes https://github.com/grafana/synthetic-monitoring-app/issues/1164 (together with https://github.com/grafana/deployment_tools/pull/283247)

PR #1192, opened 2025-07-14
fix: publish to gcom on release

VikaCep:
**Add missing publish to grafana.com step for production deployments**

This step was present in the original Drone CI setup but was omitted during the GitHub Actions migration. Without this step, the plugin  isn't published to the public registry and doesn't sync to the CDN.

**References:**
- Original Drone implementation: https://github.com/grafana/synthetic-monitoring-app/pull/1158/files#diff-b54b39f1afced2465e1f3641db9d5bbf4f3a7fcf890996dfedd3c197bcb7f8c7L210
- CDN documentation: https://enghub.grafana-ops.net/docs/default/component/grafana-plugins-platform/plugins-cdn/010-plugins-cdn/#uploading-files-to-the-bucket


ckbedwell (approved):
Great catch. LGTM 🥇

Issue #1014, opened 2024-12-09
Enhance workflow for creating a private probe


ckbedwell:
## Problem

The workflow for setting up a private probe is somewhat nonsensical at the moment, especially when reading it combined with the documentation. The steps feel out of order and certain information is presented that you don't know you need until you get to a later point in the setup process.

## Acceptance criteria

Please rework the private probe creation process to have an enhanced UX and more intuitive workflow. The documentation will also need updating. Proposed workflow:

1. Click new private probe
2. Fill in probe details
3. Indicate the purpose of the map (_it is not clear at the moment and not necessarily important so move where it is in the workflow_)
4. Click 'Add new probe'
5. Either make the modal undismissable until explicitly clicked on a dismiss method (e.g. **not** the overlay but a close button) or add the information on page.
6. Display both the access token and the **api server** (hardcode the mapping to the application) explaining both will be needed when setting up the probe.
7. When viewing the new probe on the edit probe page, display the api server alongside the reset access token
8. Add a test probe button which utilizes the adhoc functionality.

Please ensure whatever solution is implemented has the appropriate integration tests added.

peterschretlen:
Related escalation where HTTP server URL is used instead of GRPC server URL (internal link) https://github.com/grafana/support-escalations/issues/15345

Pokom:
I had ran into this while onboarding as well. I think the most problematic bit that I found that would go a long way in solving is how the api server is displayed. It's only displayed _before_ creating the access token, and I believe it provides the incorrect address(as @peterschretlen linked out in a related escalation).

PR #1194, opened 2025-07-15
feat: add probe server api url alongside backend address for private probe creation

ckbedwell:
## Problem
When creating a private probe it is confusing and unclear that you have to use your backend address and cross-reference it with a table in the documentation so you have the correct Probe API 

## Solution

_As outlined in [1014](https://github.com/grafana/synthetic-monitoring-app/issues/1014) this PR addresses point 6 only._

This PR does two things: add a hard-mapping of the API servers to the plugin (src/page/NewProbe/probeAPIServerMappings.json) and reference the content in a new component called `<ProbeAPIServer />`. This component replaces all instances where the backend address is mentioned:

### Probe creation righthand side:

**Before**
image

**After**
image

### Probe creation / reset modal:

**Before**
image

**After**
image

### General tab:

**Before**
image

**After**
image

## Alleviating the maintenance burden using CI/CD

We have discussed the maintenance burden that a hard-mapping creates so to address this I have created a CI/CD workflow action that checks for every PR if the mappings files matches what is currently on the docs website.

It runs on every PR created / updated so in theory it isn't perfectly up to date but we should get notified fairly quickly when a drift happens.

See the comment on this PR and the edit history to see what failures / successes look like.


src/components/ProbeAPIServer/ProbeAPIServer.tsx
```diff
@@ -0,0 +1,57 @@
...
+        the region of your stack. This is the value you will need to set in the <code>API_SERVER</code> environment
+        variable when setting up a private probe.
+      </Text>
+      {probeAPIServer ? <Preformatted>{probeAPIServer}</Preformatted> : <NoProbeAPIServer />}
+      <Stack direction="column" gap={0.5}>
+        <Text variant="bodySmall" italic>
+          Your backend address is:
+        </Text>
+        <Text color="warning" variant="bodySmall" italic>
+          {backendAddress}
```
VikaCep, line 35: Is there value in showing the `backendAddress` given what the user needs to copy and paste is actually the `probeAPIServer` url?

.github/workflows/call_verify-probe-api-server-mappings.yml
```diff
@@ -0,0 +1,165 @@
...
+                // Remove yarn wrapper messages and final instruction for cleaner output
+                const cleanOutput = output
+                  .split('\n')
+                  .filter(line => 
+                    !line.includes('yarn run') && 
+                    !line.includes('$ yarn') && 
+                    !line.includes('$ /Users/') && 
+                    !line.includes('error Command failed') && 
+                    !line.includes('info Visit https://yarnpkg.com') &&
+                    !line.includes('💡 Please update the probeAPIServerMappings.json file to match the documentation.')
```
VikaCep, line 63: Can we change this to say what command to run to do it?
```
💡 Please update the probeAPIServerMappings.json file to match the documentation by running `yarn build:probe-api-mappings`
```

May also be nice to provide an option to ask and run it automatically upon confirming in case it's being run locally.

.github/workflows/call_verify-probe-api-server-mappings.yml
```diff
@@ -0,0 +1,165 @@
...
+              await github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: message
+              });
+              console.log('Created new comment');
+            }
+
+      - name: Comment on PR if mappings are up to date
```
VikaCep, line 119: Do we really want to comment on the PR if everything is up to date? It is useful to get a comment once it fails, but if everything is OK I don't see the value of getting extra comments in the PR indicating that.

.github/workflows/call_verify-probe-api-server-mappings.yml
```diff
@@ -0,0 +1,165 @@
...
+              await github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: message
+              });
+              console.log('Created new comment');
+            }
+
+      - name: Fail the workflow if verification failed
```
VikaCep, line 154: This won't block merging PRs right?

src/components/ProbeAPIServer/ProbeAPIServer.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <Stack direction="column" gap={2}>
+      <Text element="h3">Probe API Server URL</Text>
+      <Text>
+        The Synthetic Monitoring Agent will need to connect to the instance of the Synthetics API that corresponds with
+        the region of your stack. This is the value you will need to set in the <code>API_SERVER</code> environment
+        variable when setting up a private probe.
+      </Text>
+      {probeAPIServer ? <Preformatted>{probeAPIServer}</Preformatted> : <NoProbeAPIServer />}
```
VikaCep: Can we add a clipboard button to copy this value here?

VikaCep (commented):
Really nice work! Added a few questions/comments.

src/components/ProbeAPIServer/ProbeAPIServer.tsx
ckbedwell: I have some reluctance to completely remove it.

1. We've displayed it for a long time so who knows what side effects it might have for our users to hide it from them. Perhaps some users have a use for it...?
2. We do still have the reference table in the documentation, so it might be confusing to users what the backend address is when looking at that table.

Happy to revisit this in the future but as I've made it fairly small and unsuspecting I think it is okay to carry on displaying it.

.github/workflows/call_verify-probe-api-server-mappings.yml
ckbedwell: Well... define block 😅 ? It will fail the workflow if it is not up to date so you'll have a big red X by your PR until you fix it.

_(YOLO button exists to unblock merging but I can't think why you wouldn't fix it first?)_

src/components/ProbeAPIServer/ProbeAPIServer.tsx
ckbedwell: image

ckbedwell:
Passed CI with new code and didn't post a success message (I deleted the previous one). Will test with error to check for comment, then will push fix to see if it updates the comment correctly.

.github/workflows/call_verify-probe-api-server-mappings.yml
VikaCep: I was thinking of this as an optional check because it might be complete unrelated to the changes you're making in your PR, so I don't think it should necessarily block merging it and I'm not a fan of the `bypass rules` checkbox  😅 
But yeah, most of the times it will be pain-free to just fix it.

src/components/ProbeAPIServer/ProbeAPIServer.tsx
VikaCep: No objections about keeping it 👍

VikaCep (approved):
👏 👏 👏

PR #1201, opened 2025-07-16
fix: prevent loading errorRateMap when data has not finished loading

VikaCep:
## Problem Description

The ErrorRateMap component sometimes throws this error on first load, as reported in [Slack](https://raintank-corp.slack.com/archives/C0162C1K9E1/p1752663586695719):

```
TypeError: Cannot read properties of undefined (reading 'setActive')
  at GeomapPanel.tsx:367:25
 // at this.mouseWheelZoom!.setActive(Boolean(options.mouseWheelZoom));
```

image

This typically happens on the first page load and goes away after refresh.

## Root Cause

The issue is a race condition where the geomap panel tries to initialize before the data queries complete and the expected data structure is available. The geomap attempts to call `setActive()` on undefined objects during its internal initialization.

## Fix Applied

1. **Loading state tracking** - Added `hasLoaded` state variable to track if data has loaded successfully at least once
2. **Prevent early rendering** - Only renders the geomap after data has loaded successfully 
3. **Consistent pattern** - Uses the same approach as other components like `AssertionsTable` (PR #666)

## How to reproduce

The issue can be reproduced by loading a HTTP dashboard with slow network settings (4g throttling did it for me), or forcing a delay locally after the `dataProvider` creation in `ErrorRateMap`: 
```
  useEffect(() => {
    // Force race condition by rendering the geomap before data is ready
    const timer = setTimeout(() => {
      console.log('Force error timer complete - allowing normal rendering');
    }, 5000);
    
    return () => clearTimeout(timer);
  }, []);
  ```
  
  which will output the error:
image

With the fix, this doesn't happen anymore:

image

### DIFF
--- a/src/scenes/HTTP/ErrorRateMap.tsx
+++ b/src/scenes/HTTP/ErrorRateMap.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { LoadingState } from '@grafana/data';
 import { VizConfigBuilders } from '@grafana/scenes';
 import { useDataTransformer, useQueryRunner, useTimeRange, VizPanel } from '@grafana/scenes-react';
@@ -11,6 +11,7 @@ import { useVizPanelMenu } from './useVizPanelMenu';
 
 export const ErrorRateMap = ({ minStep }: { minStep: string }) => {
   const metricsDS = useMetricsDS();
+  const [hasLoaded, setHasLoaded] = useState(false);
 
   function getErrorMapQueries(minStep: string) {
     return [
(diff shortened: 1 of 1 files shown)


ckbedwell (commented):
1. I'm struggling to repro this even with the nice repro steps outlined above.
2. I don't love the solution and am not convinced it will always solve the issue (I understand why it was reached for, though). It's more of a delaying tactic based on the rendering cycle which React uses.
3. My hunch is the issue lies somewhere between the rendering condition is based on the `dataProvider's Loading state` but we pass `dataTransformer` to the viz panel as the `dataProvider`? Looking through the props for `dataTransformer` I see there is a `dataTransformer.isDataReadyToDisplay()` -- could try using that instead?

VikaCep:
I've updated it to use a combination of dataTransformer.isDataReadyToDisplay() and data?.data?.state !== LoadingState.Done. isDataReadyToDisplay() checks whether a data provider has successfully fetched data at least once, but when we're refreshing a visualization we still need to handle the loading state separately.

Originally, we were only checking for data?.data?.state === LoadingState.Loading to show the loading placeholder. That would miss cases where the state was NotStarted, so the check would be skipped. With these conditions in place, the visualization will now wait until we actually have data before trying to render 👍

ckbedwell:
> That would miss cases where the state was `NotStarted`

Quick question: do you know how that state is achieved?

Issue #1031, opened 2024-12-20
Saving a check shouldn't take you back to the check list screen


d0ugal:
**What happened**:

If you edit a check in the UI and save it, you get taken back to the list of checks.

**What you expected to happen**:

I expect to be taken to the check dashboard.
This is annoying because you will often want to see the result of your change.
If you have more than a few checks you need to search for it again on the list view to find it.

**How to reproduce it (as minimally and precisely as possible)**:

1. Click "edit" for a check (either on the list of checks or on the check dashboard)
2. Make some changes (or don't) and click save
3. You are taken back to the list view

peterschretlen:
This is particularly annoying when doing script development or troubleshooting.  I usually avoid saving my changes until I have something working with this loop:

1. Disable the check that isn't working as I want 
2. Go to the "script" step in the edit flow, tweak the script
3. Go to the Execution step
4. Run and adhoc check, review the logs
5. Go back to step 2
6. Only save when I have something working
7. Re-enable the check. 

An alternative to would be to default sort the checks list by 'last modified' (which isn't an option currently).  If you have a lot of checks, saving is a pain because you can't easily find it again if you want to view the dashboard.  Often I will edit a check from the dashboard, so it's surprising to end up on the checks listing page.

PR #1203, opened 2025-07-18
feat: change redirect logic for editing a check

ckbedwell:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/1031

## Problem

Currently when editing any check you get taken back to the check list rather than the check dashboard.

## Solution

Redirect to the check dashboard in all cases. It is worth noting I explored contextually redirecting the user (so if you came from the checklist you would go back to it) but the inconsistent behaviour was confusing and I felt this was more of a concern for redesigning the checklist page to encourage edits to be done on the page directly.


src/components/ProbeSetupModal/ProbeSetupModal.tsx
```diff
@@ -73,7 +72,6 @@ const EnvsTable = ({ token }: { token: string }) => {
       pagination={false}
       id="probe-setup-modal-table"
       name="probe-setup-modal-table"
-      config={config}
```
ckbedwell: This was a typescript error from another PR. Not sure how it snuck in...

VikaCep (changes requested):
I'm getting some weird behavior when saving a new check, I get an error and then it stays in the edit form, but the check is saved:

https://github.com/user-attachments/assets/e4d78e4c-1580-4abe-b943-3bb523f7ea26

VikaCep (approved):
LGTM!

PR #1204, opened 2025-07-18
POC: terraform validation

VikaCep:
WIP


.gitignore
```diff
@@ -39,3 +39,7 @@ temp/
 .DS_Store
 
 dev/license.jwt
+
+# Terraform validation outputs
+test-terraform-output/
+terraform-schema.json
```
ckbedwell: Should we just change the output to the `./artifacts` folder as that's already gitignored and helps standardise where we add these files (_artifacts_)? 😅

package.json
```diff
@@ -25,8 +25,9 @@
     "test:changed": "jest --watch --onlyChanged",
     "test:ci": "jest --passWithNoTests --maxWorkers 4 --bail",
     "test": "jest --all",
-    "typecheck": "tsc --noEmit",
-    "verify:probe-api-mappings": "yarn ts-node scripts/probe-api-mappings/verify-probe-api-mappings.ts"
+    "terraform:generate-test-config": "tsx scripts/terraform-validation/generate-test-configs.mts",
```
ckbedwell: Can we change this script name to `build:generate-terraform-test-config` and can we add another script called `verify:terraform-test-config` which runs the `terraform validate` (might need to be a shell script that steps in and out of the right directory looking at the validate command) 🙏

Both of these are because a) I'm lazy. b) I like standardisation.

ckbedwell (changes requested):
This is awesome. Really glad we are starting to lean into CI/CD more 🙏 

Minor change requests but otherwise looks good!

ckbedwell (approved):
Works great! ✅ 🔥🔥🔥

PR #1205, opened 2025-07-23
fix: use secret name, not UUID, to interact with the API

mem:
The API accepts UUIDs, but the downstream service doesn't. Instead of trying to convert on the fly (which is expensive) make the plugin send names.

Fixes #

### DIFF
--- a/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
+++ b/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
@@ -24,12 +24,12 @@ export function SecretsManagementUI() {
     setEditMode(SECRETS_EDIT_MODE_ADD);
   };
 
-  const handleEditSecret = (id?: string) => {
-    setEditMode(id ?? false);
+  const handleEditSecret = (name?: string) => {
+    setEditMode(name ?? false);
   };
 
-  const handleDeleteSecret = (id: string) => {
-    const secret = secrets?.find((s) => s.uuid === id);
+  const handleDeleteSecret = (name: string) => {
+    const secret = secrets?.find((s) => s.name === name);
     if (secret) {
       setDeleteMode(secret);
     }
@@ -81,7 +81,7 @@ export function SecretsManagementUI() {
       )}
 
       {editMode && (
-        <SecretEditModal id={editMode} existingNames={existingNames} open onDismiss={() => handleEditSecret()} />
+        <SecretEditModal name={editMode} existingNames={existingNames} open onDismiss={() => handleEditSecret()} />
       )}
 
       <ConfirmModal
@@ -101,7 +101,7 @@ export function SecretsManagementUI() {
           </div>
         }
         onConfirm={() => {
-          deleteMode && deleteSecret.mutate(deleteMode.uuid);
+          deleteMode && deleteSecret.mutate(deleteMode.name);
           setDeleteMode(undefined);
         }}
         onDismiss={() => setDeleteMode(undefined)}
--- a/src/test/handlers/index.ts
+++ b/src/test/handlers/index.ts
@@ -11,7 +11,7 @@ import { createAccessToken } from 'test/handlers/tokens';
 import { ApiEntry, RequestRes } from 'test/handlers/types';
 
 import { listAlertsForCheck, updateAlertsForCheck } from './alerts';
-import { createSecret, getSecret, listSecrets, updateSecret } from './secrets';
+import { createSecret, deleteSecret, getSecret, listSecrets, updateSecret } from './secrets';
 
 const apiRoutes = {
   addCheck,
@@ -42,6 +42,7 @@ const apiRoutes = {
   getSecret,
   createSecret,
   updateSecret,
+  deleteSecret,
 };
 
 export type ApiRoutes = typeof apiRoutes;
@@ -74,7 +75,9 @@ export function apiRoute<K extends keyof ApiRoutes>(
 }
 
 function toRestMethod({ route, method, result }: ApiEntry) {
-  const urlPattern = new RegExp(`^http://localhost.*${route}$`);
+  // Convert route patterns like /api/v1alpha1/secrets/:name to regex patterns
+  const patternRoute = route.replace(/:[^/]+/g, '[^/]+');
+  const urlPattern = new RegExp(`^http://localhost.*${patternRoute}$`);
 
   return rest[method](urlPattern, async (req, res, ctx) => {
     const { status = 200, json } = await result(req);
--- a/src/test/handlers/secrets.ts
+++ b/src/test/handlers/secrets.ts
@@ -44,7 +44,7 @@ export const listSecrets: ApiEntry<SecretsResponse> = {
  * - `result`: A function representing the response handler, returning a mock JSON payload containing the secret data.
  */
 export const getSecret: ApiEntry<SecretWithMetadata> = {
-  route: `/api/v1alpha1/secrets/${MOCKED_SECRETS_API_RESPONSE.secrets[0].uuid}`,
+  route: `/api/v1alpha1/secrets/:name`,
   method: `get`,
   result: () => {
     return {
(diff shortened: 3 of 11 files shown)


src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
```diff
@@ -81,7 +81,7 @@ export function SecretsManagementUI() {
       )}
 
       {editMode && (
-        <SecretEditModal id={editMode} existingNames={existingNames} open onDismiss={() => handleEditSecret()} />
+        <SecretEditModal name={editMode} existingNames={existingNames} open onDismiss={() => handleEditSecret()} />
```
mem, line 84: this is one where I'm scratching my head. I don't understand if this `id` here is supposed to be an id parameter, or some kind of id attribute that has nothing to do with the edit operation.

src/test/handlers/secrets.ts
```diff
@@ -44,7 +44,7 @@ export const listSecrets: ApiEntry<SecretsResponse> = {
  * - `result`: A function representing the response handler, returning a mock JSON payload containing the secret data.
  */
 export const getSecret: ApiEntry<SecretWithMetadata> = {
-  route: `/api/v1alpha1/secrets/${MOCKED_SECRETS_API_RESPONSE.secrets[0].uuid}`,
+  route: `/api/v1alpha1/secrets/:name`,
```
mem: this one put up a fight. I'm not sure why the route is set up like this.

src/test/handlers/secrets.ts
```diff
@@ -84,11 +84,31 @@ export const createSecret: ApiEntry<SecretFormValues> = {
...
+ * Represents the API endpoint for deleting a secret.
+ *
+ * This configuration includes the API route, HTTP method, and the expected result.
+ *
+ * Properties:
+ * - `route`: The API endpoint for deleting a secret.
+ * - `method`: The HTTP method used for the request, which is "delete" in this case.
+ * - `result`: A function that returns the expected result of the API call, which is a JSON response with a value of `null`.
+ */
+export const deleteSecret: ApiEntry<unknown> = {
```
mem: this bit here is me basically looking at the rest of the file and asking myself "is this needed? it looks like it's needed"

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
VikaCep: From the interface of `SecretsEditModal`, it looks like the `id` was a prop for the component and you were correct in changing it to `name`

```
interface SecretEditModalProps {
  id: string;
  onDismiss: () => void;
  open?: boolean;
  existingNames?: string[];
}
```

src/test/handlers/secrets.ts
VikaCep: It's so the route matches exactly what the tests use. Instead of `/api/v1alpha1/secrets/:name` you could also use `/api/v1alpha1/secrets/${MOCKED_SECRETS_API_RESPONSE.secrets[0].name}` and it should still work (see https://github.com/grafana/synthetic-monitoring-app/blob/ad93b4dcc91e0d2b9788fa41149817715cae5ba0/src/test/fixtures/secrets.ts#L7)

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
mem: ah, thank you!

I know I have broken this somehow because if I run the code and I hit the edit button, it comes back with an error, and the console says:

```
react-dom.production.min.js:188 TypeError: Cannot read properties of null (reading 'length')
    at oF (SecretEditModal.tsx:145:38)
```

mem: OK, I did fix that error. Still tracking down why the edit dialog doesn't work.

src/test/handlers/secrets.ts
mem: uh... thank you. Now it's working in the original form.

I must have changed something else that broke it.

VikaCep: After some debugging, I think I can see what's going on. The schema is ignoring the `name` prop so the request is not including it in the url (it ends up as `PUT /api/v1alpha1/secrets/undefined`):

https://github.com/user-attachments/assets/deff467a-1213-4515-a1f7-022b0c17bc6d

In order to fix it, you have to add it to the schema. I made a commit with these changes in a new branch: https://github.com/grafana/synthetic-monitoring-app/commit/d8354a5dbbfca6b96090021fa69ac84968c3dd9d

You can see the request url is correctly formed now: 

https://github.com/user-attachments/assets/67537a84-8086-4977-b652-a8ba9773c1ff

VikaCep: There's a follow up commit (https://github.com/grafana/synthetic-monitoring-app/pull/1206/commits/2b8dc8b5a6143644399ee9f9a0a5eea245224ef0) which makes the `name` prop be excluded from the payload when editing the secret. This decision seems to be deliberate to prevent users from changing it, so my second commit keeps it in the form schema (so we can access its to form the request URL) but excludes it from the payload. This also fixes the tests that were failing.  

See the draft PR: https://github.com/grafana/synthetic-monitoring-app/pull/1206

src/test/handlers/index.ts
```diff
@@ -74,7 +75,9 @@ export function apiRoute<K extends keyof ApiRoutes>(
 }
 
 function toRestMethod({ route, method, result }: ApiEntry) {
-  const urlPattern = new RegExp(`^http://localhost.*${route}$`);
+  // Convert route patterns like /api/v1alpha1/secrets/:name to regex patterns
+  const patternRoute = route.replace(/:[^/]+/g, '[^/]+');
+  const urlPattern = new RegExp(`^http://localhost.*${patternRoute}$`);
```
VikaCep: I'd go back to just 
```suggestion
 const urlPattern = new RegExp(`^http://localhost.*${route}$`);
```
since you're not using the `:name` way any longer. 

With the current code I get a type error
image

dev/custom.ini
```diff
@@ -796,6 +796,7 @@ level = debug
 # enable features, separated by spaces
 scripted-checks=true
 dockedMegaMenu=true
+synthetic-monitoring-secrets-management=true
```
VikaCep: Was this change on purpose? 

```suggestion
```

src/test/handlers/secrets.ts
```diff
@@ -84,11 +84,31 @@ export const createSecret: ApiEntry<SecretFormValues> = {
...
+  route: `/api/v1alpha1/secrets/:name`,
+  method: `delete`,
+  result: () => {
+    return {
+      json: null,
+    };
+  },
+};
```
VikaCep: This block is not needed as the delete route is not being directly called in the `SecretEditModal.test` tests (as opposed to others, like https://github.com/grafana/synthetic-monitoring-app/blob/main/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx#L96)
```suggestion
```

src/test/handlers/index.ts
```diff
@@ -42,6 +42,7 @@ const apiRoutes = {
   getSecret,
   createSecret,
   updateSecret,
+  deleteSecret,
```
VikaCep:
```suggestion
```

src/test/handlers/index.ts
```diff
@@ -11,7 +11,7 @@ import { createAccessToken } from 'test/handlers/tokens';
 import { ApiEntry, RequestRes } from 'test/handlers/types';
 
 import { listAlertsForCheck, updateAlertsForCheck } from './alerts';
-import { createSecret, getSecret, listSecrets, updateSecret } from './secrets';
+import { createSecret, deleteSecret, getSecret, listSecrets, updateSecret } from './secrets';
```
VikaCep:
```suggestion
import { createSecret, getSecret, listSecrets, updateSecret } from './secrets';
```

src/test/handlers/secrets.ts
mem: thank you, @VikaCep!

src/page/ConfigPageLayout/tabs/SecretsManagementTab/secretSchema.ts
```diff
@@ -72,7 +72,9 @@ const labels: ZodType<Secret['labels']> = z
       });
       return z.NEVER;
     }
-  });
+  })
+  .nullable()
```
mem: after a lot of back and forth, I figured the reason why the form is not allowing me to submit an updated secret is because the labels field returned from the API (and used to populate the form) is null, and the validation doesn't like that.

VikaCep: Is this change in the labels response being `null` recent? with this change making the labels nullable I'm getting a type error now when running `yarn dev` (or `yarn typecheck`). So we might have to do some extra changes. 
image

VikaCep: I'd go back to making it non nullable, but instead, transform the labels API response to an empty array when it's null. Otherwise, we'd have to change it in several other parts of the code to make it ready for null. I've created a commit with these changes in case you agree with the idea: https://github.com/grafana/synthetic-monitoring-app/commit/521e64dde4641f88a18d87be22ccd262161a8cde

mem: thank you! that works, I'll update the PR and push again

PR #1173, opened 2025-07-04
Feat: per check alerts runbooks UI support

VikaCep:
Frontend part of https://github.com/grafana/synthetic-monitoring/issues/293
Closes https://github.com/grafana/synthetic-monitoring/issues/315

### Add runbook URL support to per-check alerts

This adds the ability to configure runbook URLs for individual alert types on each check. When an alert fires, users can click the runbook link to go directly to troubleshooting documentation.

https://github.com/user-attachments/assets/da921b9c-699c-4330-ac72-e34d97350ed4

**What this does:**
- Adds optional "Runbook URL" field to alert configuration forms
- When alerts fire, the runbook URL gets templated into alert notifications*
- Clicking the runbook link redirects users to the configured documentation
- If no runbook URL is set, falls back to normal dashboard behavior

**Changes:**
- Added `RunbookUrl` component for form input with URL validation
- Enhanced SceneRedirecter to handle runbook redirects
- Updated alert forms (TLS certificate expiry, probe failures) to include runbook field

**How it works:**

- User configures alert with optional runbook URL (e.g., "https://company.com/docs/probe-troubleshooting")
- Alert fires and includes templated runbook link in notification
- User clicks link → goes to the Synthetic Monitoring UI first
- UI finds the specific check and alert, looks up the configured runbook URL
- UI then redirects to the configured documentation
- If something goes wrong (missing URL, check not found, etc.), falls back to the normal check dashboard

**URL Format:***

The backend generates alert notifications with templated runbook URLs like:

`https://your-stack.grafana.net/a/grafana-synthetic-monitoring-app/scene?var-instance={{ urlquery $labels.instance }}&var-job={{ urlquery $labels.job }}&var-alert={{ urlquery $labels.alertname }}&var-runbook=true`

When the alert fires, the template variables are populated with actual values from the alert labels.

**Dependencies:**
- Requires backend PR to be merged first: https://github.com/grafana/synthetic-monitoring-api/pull/1421

**Note:**

Warnings and messages regarding the runbook url not existing when the URL contains the runbook=true parameter were handled here https://github.com/grafana/synthetic-monitoring-app/pull/1207 and it's merged into this PR.

PR #1207, opened 2025-07-24
feat: Runbook URL handing error scenarios

VikaCep:
Frontend part of https://github.com/grafana/synthetic-monitoring/issues/293
Builds on https://github.com/grafana/synthetic-monitoring-app/pull/1173

### Improve runbook redirect UX and handle missing runbook URLs

This enhances the runbook redirect functionality by properly handling cases where runbook URLs are not configured and improving the overall user experience during redirects.

**What this does:**

**- Shows an informative alert when runbook URLs are missing from alert configuration.** 

When a user clicks a runbook link but no URL is configured for that alert type, the UI detects the missing runbook via `runbookMissing` query parameter and shows an informative warning in the alert configuration form.

image

**- Provides a better redirect loading screen with manual fallback option.** 

Added `RunbookRedirectAlert` component with better UX for the redirection state. When redirecting, it now shows a clear "Redirecting to runbook" alert with the target URL and provides "Take me there now" button as manual fallback.

image

**Dependencies:**
- Requires the base runbook functionality from feat/per-check-alerts-runbooks branch
- Backend support from https://github.com/grafana/synthetic-monitoring-api/pull/1421

### DIFF
--- a/src/components/CheckForm/CheckForm.hooks.ts
+++ b/src/components/CheckForm/CheckForm.hooks.ts
@@ -52,6 +52,7 @@ type CheckFormMetaReturn = {
   isDisabled: boolean;
   isLoading: boolean;
   defaultFormValues: CheckFormValues;
+  initialSection?: number;
 };
 
 export function useCheckFormMeta(check?: Check, forceDisabled = false): CheckFormMetaReturn {
--- a/src/components/CheckForm/CheckForm.hooks.ts
+++ b/src/components/CheckForm/CheckForm.hooks.ts
@@ -30,6 +30,7 @@ import { useNavigation } from 'hooks/useNavigation';
 import { toFormValues, toPayload } from '../CheckEditor/checkFormTransformations';
 import { getAlertsPayload } from '../CheckEditor/transformations/toPayload.alerts';
 import { fallbackCheckMap } from '../constants';
+import { SectionName } from './FormLayout/FormLayout.constants';
 import { layoutMap } from './FormLayouts/constants';
 import { broadcastFailedSubmission, findFieldToFocus, getIsExistingCheck } from './CheckForm.utils';
 import { SCHEMA_MAP } from './constants';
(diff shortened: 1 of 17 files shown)


VikaCep:
I've tested this changes with the API deployed in dev, and it works as expected:

- Saving a runbook url for an alert, and loading it later by going through the plugin:

https://github.com/user-attachments/assets/8081fdbf-a83d-49f0-9066-c48f17308879

- Loading a runbook url for an alert that doesn't have the runbook configured, correctly displaying the warning message:

https://github.com/user-attachments/assets/64ed14da-5717-4c48-b51b-c62d60420b6f

src/components/CheckForm/CheckForm.hooks.ts
```diff
@@ -52,6 +52,7 @@ type CheckFormMetaReturn = {
   isDisabled: boolean;
   isLoading: boolean;
   defaultFormValues: CheckFormValues;
+  initialSection?: number;
```
ckbedwell: I'm thinking we should change this so we provide the name of the section as the prop we provide (e.g. `initialSection="alerting"`) then somewhere down the line something converts it to the abstraction number. We have a bit of tech debt to tidy up with the tests because they made the assumption the steps order wouldn't change, so rather we don't fall into that trap again here.

src/page/EditCheck/EditCheck.tsx
```diff
@@ -20,14 +21,18 @@ export const EditCheck = () => {
   const { id } = useParams<CheckPageParams>();
   const { data: checks, isError, isLoading, error, refetch, isFetched } = useChecks();
   const check = checks?.find((c) => c.id === Number(id));
+  const urlSearchParams = useURLSearchParams();
+
+  // Check for runbook missing notification to determine initial section
+  const initialSection = !!urlSearchParams.get('runbookMissing') ? 4 : undefined; // Section 4 is "Alerting"
```
ckbedwell: So this would be:

```suggestion
  const initialSection = !!urlSearchParams.get('runbookMissing') ? "alerting" : undefined;
```

ckbedwell (changes requested):

src/page/EditCheck/EditCheck.tsx
VikaCep: Nice idea 👍

PR #1210, opened 2025-07-29
fix: convert secrets management creation date to millis

peterschretlen:
The secrets list is showing created dates in 1970.  This is because the private preview secrets manager backend returned milliseconds since epoch,  but the public preview backend now returns **seconds** since epoch. 

This PR converts back to millis on the frontend

image

### DIFF
--- a/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.tsx
+++ b/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretCard.tsx
@@ -70,7 +70,7 @@ export function SecretCard({ secret, onEdit, onDelete }: SecretCardProps) {
         <strong>Description:</strong> {secret.description}
       </div>
       <div className={styles.keyValue}>
-        <strong>Created:</strong> {formatDate(secret.created_at)} ({secret.created_by})
+        <strong>Created:</strong> {formatDate(secret.created_at * 1000)} ({secret.created_by})
       </div>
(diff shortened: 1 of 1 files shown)


VikaCep (commented):
If I go to the secrets tab in my org in prod by setting the feature flag to true, I see the date correctly formatted (without the changes in this PR). 

image

How can I reproduce the issue?

peterschretlen:
@VikaCep it will be reproducible in the dev environment, or in any prod environments that other than us-east-0 and us-central-0.  us-east-0 and us-central-0 are still deployed with the private preview and will display the correct date.

d0ugal:
I wondered what will happen for users in the private preview. I ran this in my console.

This confirms the 1970 date and private preview users will be transported to the year 57547 😁 

```
new Date(Date.now() / 1000)
Wed Jan 21 1970 08:11:01 GMT+0100 (Greenwich Mean Time)
new Date(Date.now() * 1000)
Wed Aug 27 57547 02:21:18 GMT+0100 (British Summer Time)
```

VikaCep:
> I wonder if it's better just to omit the display of creation time for now?
@d0ugal has a good point about the private preview (though that will be temporary), and even after this change it's unclear if the backend will return the correct value at launch time (currently it is not).

We can do that. I also created a suggestion [PR](https://github.com/grafana/synthetic-monitoring-app/pull/1214) to handle both cases but I'm not sure we want to cover for that scenario.

PR #1213, opened 2025-07-30
fix: restrict secrets tab usage to admin users

VikaCep:
From https://raintank-corp.slack.com/archives/C07D9QWFTQ8/p1753822656278269?thread_ts=1753720623.506399&cid=C07D9QWFTQ8

As a first stage, we need to restrict the content of the Secrets tab to just Admins.

**Viewer/Editor**
image

**Admin**
image

I reused the `ContactAdminAlert` but I'm open to changing it to other UI if preferred.

At a later stage, we'll focus on adding more granular permissions, but it's out of the scope of this PR.

### DIFF
--- a/src/data/permissions.ts
+++ b/src/data/permissions.ts
@@ -53,4 +53,6 @@ export const getUserPermissions = () => ({
   canWritePlugin: isUserActionAllowed('grafana-synthetic-monitoring-app.plugin:write', OrgRole.Admin),
 
   canWriteSM: isUserActionAllowed('grafana-synthetic-monitoring-app:write', OrgRole.Editor),
+
+  isAdmin: hasMinFallbackRole(OrgRole.Admin),
 });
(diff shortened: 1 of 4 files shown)


ckbedwell (approved):
In the context of the linked thread: LGTM.

_(Otherwise, booooooo. This change sucks 😅  -- let's get a ticket to get those RBAC permissions in sooner rather than later )_

VikaCep:
> (Otherwise, booooooo. This change sucks 😅 -- let's get a ticket to get those RBAC permissions in sooner rather than later )

We could easily consume the fixed roles described in the thread and apply them in the UI:

```
fixed:secret.securevalues:creator
fixed:secret.securevalues:reader
fixed:secret.securevalues:updater
fixed:secret.securevalues:deleter
```

But this would still allow users to access them via the API, so we need to sync with the backend to make sure how to address that.

VikaCep:
Here's the [PR](https://github.com/grafana/synthetic-monitoring-app/pull/1218) with RBAC permissions applied on the UI side

PR #1214, opened 2025-07-30
fix: handle both seconds and milliseconds in created_at response

VikaCep:
Suggestion for https://github.com/grafana/synthetic-monitoring-app/pull/1210 to handle both seconds and milliseconds when displaying the `created_at` time for secrets.

image


ckbedwell:
Closed as this was fixed on the backend.

PR #1215, opened 2025-07-31
feat: more featured redirect modal UI

ckbedwell:
## Suggestion

Just a suggestion for the redirect UI for the runbooks. I was playing with some options when I was reviewing the PR and ended up with something that I thought offered a bit more context to the user, so something to consider!

https://github.com/user-attachments/assets/272ddccc-e596-40ca-94d5-b9406bcd748e

### DIFF
--- a/src/components/RunbookRedirectAlert.tsx
+++ b/src/components/RunbookRedirectAlert.tsx
@@ -1,29 +1,94 @@
-import React from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { GrafanaTheme2 } from '@grafana/data';
-import { Alert, Button, Stack, Text, useStyles2 } from '@grafana/ui';
+import { Alert, Box, Button, Stack, Text, TextLink, useStyles2 } from '@grafana/ui';
 import { css } from '@emotion/css';
+import pluralize from 'pluralize';
+
+import { Check, CheckAlertWithRunbookUrl } from 'types';
+import { AppRoutes } from 'routing/types';
+import { generateRoutePath } from 'routing/utils';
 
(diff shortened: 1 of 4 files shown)


src/components/RunbookRedirectAlert.tsx
```diff
@@ -1,29 +1,94 @@
...
+interface RunbookRedirectAlertProps {
+  check: Check;
+  alertConfig: CheckAlertWithRunbookUrl;
 }
 
-export const RunbookRedirectAlert: React.FC<RunbookRedirectAlertProps> = ({ runbookUrl }) => {
+export const RunbookRedirectAlert = ({ check, alertConfig }: RunbookRedirectAlertProps) => {
+  const runbookUrl = alertConfig.runbookUrl;
   const styles = useStyles2(getStyles);
+  const [timeUntilRedirect, setTimeUntilRedirect] = useState<number | null>(5);
```
VikaCep, line 19: Why does it start in 5? 🤔

VikaCep: oh I guess that's just enough time to read all the content from the modal

ckbedwell: Yeah, I wanted to make it clear what is happening after you click on the alert link from your email / Slack, etc.

My main concern is it will be jarring going from where you received your alert to Grafana, watching multiple spinners and loaders as the Grafana waterfall renders and then once the SM plugin loads it blinks a render then you get redirected to another loading screen where your runbook URL lives.

Despite this is a bit slower to get to your final destination I think it makes it easier to understand for users how they have ended up where they are.

It's also to help debug if their runbook URL is wrong, acting slow or not responding, etc.

PR #1218, opened 2025-08-01
feat: apply granular secrets permissions

VikaCep:
Implements granular restrictions on the secrets UI consuming the following fixed permissions:

```
'secret.securevalues:create',
'secret.securevalues:read'
'secret.securevalues:write'
'secret.securevalues:delete'
  ```
  
**Editor with no secrets permissions assigned:**
image

**Editor with read permissions assigned:**
- with no secrets
image

- with secrets
image

**Editor with read+create permissions assigned:**
- with no secrets
image
- with secrets
image

**Editor with read+update permissions assigned:**
image
image

**Editor with read+delete permissions assigned:**
image

**Editor with read+create+update+delete permissions assigned:**
image
image

### DIFF
--- a/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
+++ b/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
@@ -16,7 +16,7 @@ export function SecretsManagementUI() {
   const [editMode, setEditMode] = useState<string | false>(false);
   const [deleteMode, setDeleteMode] = useState<SecretWithUuid | undefined>();
   const { canCreateSecrets, canReadSecrets } = getUserPermissions();
-  const { data: secrets, isLoading, isFetching } = useSecrets();
+  const { data: secrets, isLoading, isFetching } = useSecrets(canReadSecrets);
   const deleteSecret = useDeleteSecret();
   const emptyState = (canReadSecrets && secrets?.length === 0) || (!canReadSecrets && canCreateSecrets);
 
--- a/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
+++ b/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx
@@ -15,10 +15,10 @@ import { SecretEditModal } from './SecretEditModal';
 export function SecretsManagementUI() {
   const [editMode, setEditMode] = useState<string | false>(false);
   const [deleteMode, setDeleteMode] = useState<SecretWithUuid | undefined>();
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by danielkenlee** [comment]
Hi @VikaCep - excellent work drafting the PR along with screenshots and good attention to detail about hiding the edit / delete buttons based on whether the user has edit / delete permissions. I reviewed the screenshots and all the outlined behaviour is correct. I'm happy to test this when it's ready, just let me know.

**Review by ckbedwell — CHANGES_REQUESTED** [review]
I recognise this is a highly unlikely scenario but when I have a user with just `SECRETS_CREATOR_ACCESS` I can't see the Create secret button. 

I think we need to change the gate a little bit so the create secret button is visible but not the secrets themselves.

Everything else works great 👏

File: src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementUI.tsx [hunk 1]
```diff
@@ -14,9 +15,10 @@ import { SecretEditModal } from './SecretEditModal';
 export function SecretsManagementUI() {
   const [editMode, setEditMode] = useState<string | false>(false);
   const [deleteMode, setDeleteMode] = useState<SecretWithUuid | undefined>();
+  const { canCreateSecrets, canReadSecrets } = getUserPermissions();
   const { data: secrets, isLoading, isFetching } = useSecrets();
```
ckbedwell: I know I'm being a stickler about this but I've seen escalations related to RBAC that are quite particular about respecting their permissions. Despite this is only client-side RBAC we should try our best to enforce it.

We should disable this call to the secrets API if the user doesn't have `canReadSecrets`. We can just pass this value to `useSecrets` and change the hook to:

```
export function useSecrets(enabled: boolean) {
  const smDS = useSMDS();

  return useQuery<SecretsResponse, unknown, SecretWithMetadata[]>({
    ...secretsQuery(smDS),
    enabled,
  });
}
```

That way there is no network request for users that don't have read access. When I was testing this just now I found that if I create a secret that already exists it just overwrites the existing one so it brings into question what the point of having a create vs updating role is but that's a backend consideration 🤷

danielkenlee: > When I was testing this just now I found that if I create a secret that already exists it just overwrites the existing one so it brings into question what the point of having a create vs updating role is but that's a backend consideration

@macabu - is this expected? @PoorlyDefinedBehaviour - is this anything related to the GC work you're doing?

macabu: Fix here https://github.com/grafana/grafana-enterprise/pull/9348

PR #1236, opened 2025-08-21
Feat: add k6 version selection to check form

VikaCep:
Adds the channel selector to browser and scripted checks by consuming the new APIs provided by the backend, which are currently mocked.

These new APIs are:
GET `/api/v1/channels/k6`
GET `/api/v1/channel/k6/<channelID>/current`

See https://docs.google.com/document/d/1ZV70lIZ-xOH9EiAAp7OW2LGVeGwgHYJELdcvsSiH7UU/edit?tab=t.1cne0jb892id for a detailed specification of these endpoints and their purpose.

- Displays a new component below the script box that lets users choose the desired k6 version.
image

- The component consumes the GET `/api/v1/channels/k6` request
image

- When a channel is selected, the GET `/api/v1/channel/k6/<channelID>/current` endpoint is called to obtain the resolved version and be able to display it to the user.
image

- When trying to save when a disabled version is selected
> Checks can continue to run past the deprecatedAfter date. At some point after deprecation, we may want to disable the check entirely. Past a channel's disabledDate, checks cannot be updated unless a different channel is picked as part of the update operation. The frontend will show an alert message if a check is using a disabled channel, and potentially also a warning in the checks page if any check is using a disabled channel. 
image


src/components/CheckEditor/FormComponents/ChannelDetails.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+      {versionError && (
+        <Alert severity="warning" title="Error loading current version">
+          {versionError.message}
+        </Alert>
+      )}
+
+      {isDeprecated && (
+        <Alert severity="warning" title="Deprecated Channel">
+          This channel is deprecated and will be removed after {new Date(channel.deprecatedAfter).toLocaleDateString()}.
```
ckbedwell: After doing the timepoint Explorer I found we have `dateTimeFormat` provided by `@grafana/data` so we don't have to handle this.

src/components/CheckEditor/FormComponents/ChannelDetails.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Text variant="bodySmall" color="secondary">
+          Loading current version...
+        </Text>
+      )}
+
+      {versionError && (
+        <Alert severity="warning" title="Error loading current version">
+          {versionError.message}
+        </Alert>
+      )}
```
ckbedwell: We have a component called `QueryErrorBoundary` which you can wrap this whole component in instead of handling rendering this manually.

That component could do with some feature enhancements (like reading the thrown error) but it works well to give the user some options if it was just a one-off error.

You just want to add `throwOnError: true` to the query.

src/data/useK6Channels.ts
```diff
@@ -0,0 +1,44 @@
...
+import { SMDataSource } from 'datasource/DataSource';
+import { useSMDS } from 'hooks/useSMDS';
+
+export const queryKeys: Record<'list' | 'current', (channelId?: string) => QueryKey> = {
+  list: () => ['k6-channels'],
+  current: (channelId) => ['k6-channel-current', channelId],
+};
+
+const channelsQuery = (api: SMDataSource, enabled = true) => ({
+  queryKey: queryKeys.list(),
```
ckbedwell, line 15: As per the other comment, you would want to add `throwOnError` here to utilise the error boundary component.

src/data/useK6Channels.ts
```diff
@@ -0,0 +1,44 @@
...
+  queryFn: async (): Promise<ListChannelsResponse> => {
```
ckbedwell:
```suggestion
  queryFn: async () => {
```

This is automatically inferred so you don't need to add it explicitly.

src/data/useK6Channels.ts
```diff
@@ -0,0 +1,44 @@
...
+    } catch (e) {
+      throw new Error('Failed to load K6 version channels. Please try again later.');
+    }
+  },
+  enabled,
+});
+
+const currentVersionQuery = (api: SMDataSource, channelId: string, enabled = true) => ({
+  queryKey: queryKeys.current(channelId),
+  queryFn: async (): Promise<string> => {
```
ckbedwell:
```suggestion
  queryFn: async () => {
```

Same here.

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  return (
+    <Field
+      label="k6 Version"
+      description="Select the k6 version channel for this check"
+      htmlFor={id}
+      data-fs-element="k6 channel select"
+    >
+      <Controller
```
ckbedwell: Last time I was doing some form work I discovered the `useController` hook which I found much cleaner and easier to reason with than using the `<Controller />` component and its render props. Worth taking a look at!

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
```diff
@@ -0,0 +1,120 @@
...
+  disabled?: boolean;
+}
+
+export function K6ChannelSelect({ disabled }: K6ChannelSelectProps) {
+  const { isEnabled } = useFeatureFlag(FeatureName.VersionManagement);
+  const { control } = useFormContext<CheckFormValues>();
+  const { check, isExistingCheck } = useCheckFormMetaContext();
+  const id = 'k6-channel-select';
+
+  const { data: channelsResponse, isLoading: isLoadingChannels, error: channelsError } = useK6Channels(isEnabled);
```
ckbedwell: I'm quite surprised they are returning an object for the response. Is there somewhere I can see the discussion for that?

It's not a big deal if they keep it as an object but I think we should parse the response in the query to an array using the `select` method it provides and convert it to something more usable so this component is doing less work.

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const channels = useMemo(() => channelsResponse?.channels || {}, [channelsResponse?.channels]);
+  
+  const previousChannelId = isExistingCheck ? check?.channel : null;
+
+  const defaultChannelId = useMemo(() => {
+    return Object.entries(channels).find(([, channel]) => channel.default)?.[0] || '';
+  }, [channels]);
+
+  const channelOptions = useMemo(() => {
```
ckbedwell, line 66: Isn't this essentially a filter? Should we change it to that?

I quite like the pattern we have at the moment so if the view logic is getting further and further pushed down the file, we break it into a {Component}.(utils|hooks).ts file. Might be worth considering here!

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { useCheckFormMetaContext } from 'components/CheckForm/CheckFormContext';
+
+import { ChannelDetails } from './ChannelDetails';
+
+interface K6ChannelSelectProps {
+  disabled?: boolean;
+}
+
+export function K6ChannelSelect({ disabled }: K6ChannelSelectProps) {
+  const { isEnabled } = useFeatureFlag(FeatureName.VersionManagement);
```
ckbedwell: We have a `<FeatureFlag />` component which you can wrap this in instead, so that saves on all the network calls if this is disabled.

src/components/CheckEditor/FormComponents/ChannelDetails.test.tsx
```diff
@@ -0,0 +1,111 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { K6ChannelWithCurrent } from 'types';
+
+import { ChannelDetails } from './ChannelDetails';
+
+jest.mock('data/useK6Channels', () => ({
+  useCurrentK6Version: jest.fn(),
+}));
```
ckbedwell: Is there any reason to do it this way instead of using the test server and the `render` method we provide in `test/utils`?

My preference is using the test server because then these tests will catch problems with anything in `useCurrentK6Version` by default, especially if the implementation ever changes.

src/components/CheckEditor/FormComponents/K6ChannelSelect.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+jest.mock('@grafana/ui', () => ({
+  ...jest.requireActual('@grafana/ui'),
+  Combobox: jest.fn(({ id, placeholder, options }) => (
+    <select id={id} data-testid={id} title={placeholder}>
+      {options?.map((option: any, index: number) => (
+        <option key={index} value={option.value}>
+          {option.label}
+        </option>
+      ))}
+    </select>
```
ckbedwell: Add this to the global mock in `src/test/mocks/@grafana/ui.tsx`. There's no point in repeating ourselves if this will be a problem for all the combobox rendering.

src/components/CheckEditor/FormComponents/K6ChannelSelect.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+
+// Test wrapper with form context
+function TestWrapper({ children, featureEnabled = true }: { children: React.ReactNode; featureEnabled?: boolean }) {
+  const methods = useForm<CheckFormValues>({
+    defaultValues: {
+      channel: null,
+    }
+  });
+  
+  mockUseFeatureFlag.mockReturnValue({ isEnabled: featureEnabled });
```
ckbedwell: We have a `mockFeatureToggles` helper method to help abstract this away which could be used instead.

src/components/CheckEditor/FormComponents/K6ChannelSelect.test.tsx
ckbedwell, line 1: Same comment as the other test -- is there a reason not to use the test server? That way you get free testing for your hooks by default.

ckbedwell (changes requested):
Got a few questions and a few things to take a look at 😃 

I don't want to address loading the editor's k6 types in this PR but I am questioning the placement of the channel select. Which channel they select will affect what k6 types we load into the editor, so to me it feels like it would make more sense above the editor rather than below it.

WDYT?

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
VikaCep: From the backend [design doc](https://docs.google.com/document/d/1ZV70lIZ-xOH9EiAAp7OW2LGVeGwgHYJELdcvsSiH7UU/edit?tab=t.1cne0jb892id):

```
GET /api/v1/channels/k6
{
  "channels": {
    "v0": {
      "name": "v0.x",
      "default": false,
      "deprecatedAfter": "2025-04-01",
      "disabledAfter": "2026-04-01",
      "manifest": "k6>1,k6>=0.53"
    },
    "v1": {
      "name": "v1.x",
      "default": true,
      "deprecatedAfter": "3000-01-01",
      "disabledAfter": "3000-01-01",
      "manifest": "k6>=1"
    },
    "v2": {
      "name": "v2.x",
      "default": false,
      "deprecatedAfter": "3000-01-01",
      "manifest": "k6>=2"
    }
  }
}
```

This API is still in discussion so we could propose to change it to an array if you believe that's more appropriate, so we avoid having to do extra manipulations in the frontend.

ckbedwell (dismissed):
Love it 👏

VikaCep:
@ckbedwell I added some changes related to validation based on some changes on the backend design doc:
Now, besides `deprecatedAfter`, channels provide the `disabledAfter` prop. If a check is assigned a channel that is disabled, the dropdown will include it in the list, it but we don't allow saving the form until that's changed. From the doc:

> Checks can continue to run past the deprecatedAfter date. At some point after deprecation, we may want to disable the check entirely. Past a channel's disabledDate, checks cannot be updated unless a different channel is picked as part of the update operation. The frontend will show an alert message if a check is using a disabled channel

If this happens, we now show this validation error message:
image

I also added some payload tests that were missing.

As a sidenote, I started thinking about loading the correct k6 types for the editor but I think it will be better to address in a different PR. I have an issue for it here: https://github.com/grafana/synthetic-monitoring-app/issues/1252

src/components/CheckEditor/FormComponents/ChannelDetails.tsx
```diff
@@ -0,0 +1,73 @@
...
+        k6 version constraint:{' '}
+        <code
+          style={{
+            backgroundColor: theme.colors.background.secondary,
+            color: theme.colors.text.primary,
+            padding: '2px 6px',
+            borderRadius: theme.shape.radius.default,
+            fontSize: theme.typography.bodySmall.fontSize,
+            fontFamily: theme.typography.fontFamilyMonospace,
+          }}
```
ckbedwell: Nit: this is minor but can we use a `className` here instead?

src/types.ts
```diff
@@ -337,6 +337,7 @@ export type CheckAlertFormRecord = Partial<Record<CheckAlertType, CheckAlertForm
 export type CheckFormValuesBase = Omit<Check, 'settings' | 'basicMetricsOnly' | 'alerts'> & {
   publishAdvancedMetrics: boolean;
   alerts?: CheckAlertFormRecord;
+  channelDisabled?: boolean;
```
ckbedwell: I don't love this. Really this is metadata about the form that doesn't have any direct translation to the check payload. To be honest, we could do a better job at keeping `CheckFormValuesBase` and the expected `Check` payload more closely in sync and the more metadata values like this we add to the form the more complicated it becomes, so I think we should stay away from this pattern.

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
```diff
@@ -0,0 +1,142 @@
...
+      .map(([channelId, channel]) => {
+        const labelSuffix = channel.default ? ' (default)' : '';
+
+        return {
+          label: `${channel.name}.x${labelSuffix}`,
+          value: channelId,
+          description: `k6 version range: ${channel.manifest}`,
+        };
+      });
+  }, [channels, isExistingCheck, previousChannelId]);
```
ckbedwell: I still think we should do most of this work when we make the k6 channels API call. The only real consideration that is specific here is if its an existing check it might have access to some of the values that a new check doesn't.

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
```diff
@@ -0,0 +1,142 @@
...
+  useEffect(() => {
+    const currentChannelId = field.value || defaultChannelId;
+    if (currentChannelId && channels[currentChannelId]) {
+      const selectedChannel = channels[currentChannelId];
+      const isDisabled = new Date(selectedChannel.disabledAfter) < new Date();
+      setValue('channelDisabled', isDisabled);
+    } else {
+      setValue('channelDisabled', false);
+    }
+  }, [field.value, defaultChannelId, channels, setValue]);
```
ckbedwell: I mentioned this in another comment but I think this can be handled differently. This is masking a validation concern as a property of the check so creates some mental overhead of what is going on.

Thinking about scripted and browser check's holistically, I think it is fair to say that the k6 channels API call is now a dependency (when the feature flag is on) and should be lifted higher in the tree so once the channels have been fetched they are injected into the schema and can become part of the validation.

There's a good pattern you can extrapolate from for this which @w1kman has added for secrets validation, as the secrets validation depends on what secrets already exist -- the same way the validation here depends on what is in the check already.

src/components/CheckEditor/FormComponents/K6ChannelSelect.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const defaultChannelId = useMemo(() => {
+    return Object.entries(channels).find(([, channel]) => channel.default)?.[0] || '';
+  }, [channels]);
+
+  // Set the field value to the default channel if no value is currently set and we have a default
+  useEffect(() => {
+    if (!field.value && defaultChannelId && channels[defaultChannelId]) {
+      field.onChange(defaultChannelId);
+    }
+  }, [defaultChannelId, channels, field]);
```
ckbedwell: This feels like it should be on the `toFormValues` transformation rather than being added here.

ckbedwell (changes requested):
With these new changes and their validation considerations it has become apparent that the k6 channels affect the whole scripted and browser form experience so I think we should invest some effort into making them more integrated into the schema from a higher level rather than overloading the `<K6ChannelSelect />` component with too many considerations.

src/components/CheckEditor/FormComponents/ChannelDetails.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <ChannelDetails
+        channelId={null}
+        channels={mockChannels}
+      />
+    );
+
+    await waitFor(() => {
+      expect(screen.getByText(/each probe will use its default k6 version/i)).toBeInTheDocument();
+    });
+  });
```
ckbedwell: I couldn't work out in the UI how to repro this state?

src/components/CheckEditor/FormComponents/K6ChannelSelect.test.tsx
```diff
@@ -0,0 +1,405 @@
...
+    
+    render(
+      <FormWrapper>
+        <K6ChannelSelect />
+      </FormWrapper>
+    );
+
+    await waitFor(() => {
+      const combobox = screen.getByLabelText(/k6 version/i);
+      // v1 is the default but deprecated, so v2 should be selected as the first available
```
ckbedwell, line 118: I'll leave this comment here as I found some odd behaviour because of this in the UI when using the MSW dev mode and the current values returned. I'm assuming this state shouldn't ever exist in a live environment but it is interesting we have it in the mocks.

If when I first go to the new check form, it selects v2.x initially. I am unable to see v1.

If I leave the check form and then come back to the form (the only difference being the channel response is in the query cache) it chooses v1 by default but it doesn't show any additional information.

https://github.com/user-attachments/assets/a0446ac2-27b3-42bf-a5e0-c1a844cede22

When I log the response for the submit vlaues I don't see v1 in the payload.

src/components/CheckEditor/checkFormTransformations.ts
```diff
@@ -35,7 +35,7 @@ import { getTraceroutePayload } from 'components/CheckEditor/transformations/toP
 // export function getFormValuesFromCheck(check: PingCheck): CheckFormValuesPing;
 // export function getFormValuesFromCheck(check: TCPCheck): CheckFormValuesTcp;
 // export function getFormValuesFromCheck(check: TracerouteCheck): CheckFormValuesTraceroute;
-export function toFormValues(check: Check, checkType: CheckType): CheckFormValues {
+export function toFormValues(check: Check, checkType: CheckType, defaultChannelId?: string): CheckFormValues {
```
ckbedwell: Let's remove the unused `checkType` from this function.

src/components/CheckForm/FormLayouts/CheckBrowserLayout.tsx
```diff
@@ -19,10 +20,11 @@ export const BROWSER_CHECK_FIELDS: BrowserFields = {
 
 export const BrowserCheckLayout: Partial<Record<LayoutSection, Section<CheckFormValuesBrowser>>> = {
   [LayoutSection.Check]: {
-    fields: Object.values(BROWSER_CHECK_FIELDS).map((field) => field.name),
+    fields: [...Object.values(BROWSER_CHECK_FIELDS).map((field) => field.name), 'settings.browser.channel'],
```
ckbedwell: Is there a reason we are adding this inline and not adding it to the `BROWSER_CHECK_FIELDS` object? (same as scripted below).

src/schemas/forms/BaseCheckSchema.ts
```diff
@@ -21,6 +21,7 @@ export const baseCheckSchema = z.object({
   labels: labelsSchema,
   publishAdvancedMetrics: z.boolean(),
   alerts: checkAlertsSchema.optional(),
+  channel: z.string().nullable().optional(),
```
ckbedwell: We can remove this now.

ckbedwell (changes requested):

src/components/CheckEditor/FormComponents/ChannelDetails.test.tsx
VikaCep: This happens when the channels array is either empty (from the backend's response) or all channels are deprecated/disabled, which means they will be filtered out. I added some comments and extra tests in order to make it more clear.

src/components/CheckEditor/FormComponents/K6ChannelSelect.test.tsx
VikaCep: This is a great catch! This was caused by inconsistent filtering logic between form initialization and runtime rendering.
The first visit selected v2 (correctly filtered out deprecated v1), but the second visit selected v1 because form initialization was choosing the default one from unfiltered channels while the runtime component used filtered channels.
I fixed this by consolidating the filtering logic into the single `useFilteredK6Channels` hook used at the top level, ensuring both code paths use the same filtered channel list for consistent default selection.

https://github.com/user-attachments/assets/26ca000a-d8eb-4781-896b-2e62245ae36e

src/data/useK6Channels.ts
```diff
@@ -0,0 +1,128 @@
...
+  queryKey: queryKeys.list(),
+  queryFn: async () => {
+    try {
+      return await api.listK6Channels();
+    } catch (e) {
+      throw new Error('Failed to load K6 version channels. Please try again later.');
+    }
+  },
+  enabled,
+  throwOnError: false,
```
ckbedwell, line 24: I noticed this has changed to `false` and started emulating it failing to see what happened and I noticed we are localising the error to the k6 channels UI component.

Because it now is a dependency higher up do we think it makes sense raising the QueryErrorBoundary and not rendering the form at all if it fails to load?

VikaCep: do you think it is a blocker for the check to be created/saved? if no channel is set it will just use the default version but the check would be functional

ckbedwell: _If this PR wasn't in a holding pattern because of the BE dependency I would be more pragmatic about just approving this and getting it out but seeing we've got the time, might as well discuss the finer points of error handling in the UI 😬_ 

So I blocked just the `/current` endpoint and I was shown this error in the UI and it sent `null` as the payload. I guess there's no reason they still couldn't select a version, though, right, we just can't tell them the precisely resolved version?

image

image

When I block the channels/k6 endpoint, it resolves to `null` in the payload, too. I think it makes sense to block saving the form until the user has resolved that endpoint's response so they can submit a value. If they are editing a check and we already know their selection we could just reuse that?

VikaCep: > So I blocked just the /current endpoint and I was shown this error in the UI and it sent null as the payload. I guess there's no reason they still couldn't select a version, though, right, we just can't tell them the precisely resolved version?

Good point. I made some changes so that when the `/current` endpoint fails, we can still show and select the desired channel. A message is shown indicating that the version couldn't be resolved.
image

> When I block the channels/k6 endpoint, it resolves to null in the payload, too. I think it makes sense to block saving the form until the user has resolved that endpoint's response so they can submit a value. If they are editing a check and we already know their selection we could just reuse that?

When `/channels/k6` fails, I'm now avoiding setting the `channel` property in the payload (instead of `null` as in your screenshot). Since this is an optional property, I opted for going with graceful degradation and let users still submit their checks. If the check has an existing channel already set, it will be preserved.
I'm not 100% convinced on blocking the whole form when that endpoint fails. If no channel is selected, the backend should resolve to a default version, same thing that will happen with old checks that don't have any channel assigned by a user.
What do you think of this approach? If you still rather block the form and show an error message, I can do that instead.

image

From the backend design doc:

> When running a check with a channel field set to null, the probe will use a default channel, which may change across probe versions but will be static for a given probe version. In practice, this channel will be the one that maps to the k6 v1.

PR #1237, opened 2025-08-25
feat: refactor `<CheckForm />`  part 2

w1kman:
image

**Part 1:** The goal for this PR is to split the actual form (+internals) from the page component, in order to prepare for future improvements of the check authoring flow (See #1133).

**Part 2 (this PR):** Make components and state be accessible outside the actual `form` (`CheckForm` and internals), so that we can interact with the `form` from the "outside" (e.g have buttons in `PluginPage` without wrapping the `PluginPage` with the `form`. 

**Updated design**
- Boxed in design with a top nav (rather than a left side nav).
- "Test" and "Submit" now visible in `PluginPage` actions section
- A "Submit" button will be visible within the `CheckForm`, when on the last step
- Improved overflow behaviour
- Nav has been changed slightly in the way it renders active/visited as well as errors
- Nav has been changed in how it looks when on smaller screens

There are a bunch of contexts and hooks that still need to be refactored to clean up the code.
**Expect those to be addressed in the next PR, which adds the right-aside**
Even though tests are passing, try to spend some time creating and editing checks to make sure everything works as it should.

image

image

image


src/components/CheckForm/CheckForm.hooks.ts
```diff
@@ -205,15 +193,46 @@ export function useCheckForm({ check, checkType, checkState, onTestSuccess }: Us
...
+    // @ts-expect-error Typings does not align with the actual code
+    // @see https://github.com/react-hook-form/resolvers/blob/master/zod/src/zod.ts
+
+    const { errors } = await resolver(values, undefined, {});
+
+    if (Object.keys(errors).length === 0) {
+      return null;
+    }
+
+    return errors as FieldErrors<CheckFormValues>;
```
VikaCep:
```suggestion
    return errors;
```

src/components/CheckForm/FormLayout/FormLayoutContext.tsx
```diff
@@ -0,0 +1,159 @@
...
+  stepOrder: Record<number, { label: string; fields?: Section['fields'] }>;
+  submitDisabled: boolean;
+  visitedSections: number[];
+}
+
+export const FormLayoutContext = createContext<FormLayoutContextValue | null>(null);
+
+export function FormLayoutContextProvider({ children }: PropsWithChildren) {
+  const [visitedSections, setVisitedSections] = useState<number[]>([]);
+  const [activeSection, setActiveSection] = useState(0);
```
VikaCep: Here we should initialize the active section taking into account the `initialSection` that should be passed to the provider. Basically when the `runbookMissing` parameter is set in the URL, it should immediately go to the Alerting step to show a warning message (see https://github.com/grafana/synthetic-monitoring-app/pull/1207). I believe with the current code it won't do this, but will always show the first step first initially. 

Something like this should work:
```javascript
export function FormLayoutContextProvider({ children, initialSection }: FormLayoutContextProviderProps) {
  const [visitedSections, setVisitedSections] = useState<number[]>([]);
  
  const getInitialSectionIndex = (sectionName?: SectionName): number => {
    if (!sectionName) {
      return 0;
    }
    const index = FORM_SECTION_STEPS.indexOf(sectionName);
    return index >= 0 ? index : 0;
  };
  
  const [activeSection, setActiveSection] = useState(getInitialSectionIndex(initialSection));
```

src/components/CheckForm/useCheckType.tsx
```diff
@@ -28,5 +28,10 @@ export function useFormCheckTypeGroup(check?: Check) {
     return CHECK_TYPE_OPTIONS.find(({ value }) => value === checkType)?.group;
   }
 
+  // A default group is required for the form to render correctly, so we default to bbe/api-endpoint if no group is provided
+  if (!checkTypeGroup) {
```
VikaCep: In what case it wouldn't be provided?

src/data/useChecks.ts
```diff
@@ -240,6 +241,45 @@ export function useBulkDeleteChecks({ eventInfo, onSuccess, onError }: MutationP
...
 
+export function useAdHocCheck() {
+  // const [, setAdHocChecks] = useSessionStorage<AdHocCheckCache[]>('adHocChecks', []);
+  const dataSource = useSMDS();
+  return useMutation<AdHocCheckResponse, Error, Check, UseMutationResult>({
+    mutationFn: async ({ id, ...check }) => {
+      const data = await dataSource.testCheck(check);
+      if (!data) {
+        throw new Error(
+          'AdHocCheckError: Sever returned an empty response. Most likely due to malformed check settings.'
```
VikaCep:
```suggestion
          'AdHocCheckError: Server returned an empty response. Most likely due to malformed check settings.'
```

src/page/EditCheck/EditCheck.tsx
```diff
@@ -17,13 +17,15 @@ import { CenteredSpinner } from 'components/CenteredSpinner';
...
+import { PageActions } from '../../components/CheckForm/PageActions';
+
 export const EditCheck = () => {
   const { id } = useParams<CheckPageParams>();
   const { data: checks, isError, isLoading, error, refetch, isFetched } = useChecks();
   const check = checks?.find((c) => c.id === Number(id));
   const urlSearchParams = useURLSearchParams();
 
-  // Check for runbook missing notification to determine initial section
+  // Check for runbook missing notification to determine an initial section
```
VikaCep:
```suggestion
  // Check for runbook missing notification to determine the initial section
```

VikaCep (changes requested):
There's an issue when the `runbookMissing` url param is present, it does not automatically go to the `Alerting` step (see comment). 

Besides that it works great!
My only comment on the UI is that on big screens it looks a bit weird with the column centered and so much space on the sides  (maybe I just need to get used to it) but the save button seems to get lost on the right top corner among all the blank space:
image

src/components/CheckForm/FormLayout/FormLayoutContext.tsx
w1kman: Nice catch! I've performed some refactoring on the section order configuration and its corresponding types/maps, etc. 
Note: There were two sets of configs for section order, only one affected the actual order. As a result, incorrect "step" information would potentially been tracked, depending on feature flags.

Everything moved to a `constants.ts` file for now.

src/components/CheckForm/useCheckType.tsx
w1kman: Fallback covers unhappy path.
It's provided by router params, meaning that the hook itself could be used when there is no route param setup.

Note: refactored `useFormCheckType` as well, as changing the search param to an incorrect checkType would make the app explode.

w1kman:
Thanks @VikaCep for the swift review! I've patched all comments.

> There's an issue when the runbookMissing url param is present, it does not automatically go to the Alerting step (see https://github.com/grafana/synthetic-monitoring-app/pull/1237#discussion_r2298955524).

This has been fixed (with some bugs and double configuration of form section step order)

> My only comment on the UI is that on big screens it looks a bit weird with the column centered and so much space on the sides (maybe I just need to get used to it)

Agree. I've talked to @vesnadean about this. What we decided to do, was to center the inner UI (as the outer UI would get a companied by the right aside anyways). So hopefully you don't have to get used to it too much.

> but the save button seems to get lost on the right top corner among all the blank space

This is inherited by using the `PluginPage` component (which is also used all over grafana).
If you use the step buttons, the last step is a "Save" button (that is more in the context of the form).

Note: It's possible to wrap the `PluginPage`, but I haven't found another page in Grafana that does that. Perhaps something to consider when the right aside is in place?

VikaCep (commented):
Shouldn't the top Save button have `primary` color all the time (when it's enabled)?  I noticed it's `secondary` in all steps except the last one. Is this intentional?

image

VikaCep (dismissed):
I made a question above but it's not blocking. Works great! 🎉 ✅

src/components/CheckForm/FormLayout/FormLayout.tsx
```diff
@@ -94,158 +77,169 @@ export const FormLayout = <T extends FieldValues>({
...
-    gridTemplateColumns: `160px minmax(0, ${FORM_MAX_WIDTH}) minmax(50px, auto)`,
-    height: '100%',
-  };
+  // const breakpoint = theme.breakpoints.values.md;
+  // const query = `(min-width: ${breakpoint + 1}px)`;
+  // const containerQuery = `@container ${containerName} ${query}`;
+  //
+  // const containerRules = {
+  //   height: '100%',
+  // };
```
ckbedwell: ✂️

src/components/CheckForm/FormLayout/FormLayout.tsx
```diff
@@ -94,158 +77,169 @@ export const FormLayout = <T extends FieldValues>({
...
-      contain: 'layout',
-    }),
-    container: css({
-      display: 'grid',
-      gap: theme.spacing(4),
-      [containerQuery]: containerRules,
-      [mediaQuery]: containerRules,
+    wrapper: css`
+      container-name: ${containerName};
+      container-type: inline-size;
```
ckbedwell, line 189: Is this used now?

src/data/useChecks.ts
```diff
@@ -240,6 +241,45 @@ export function useBulkDeleteChecks({ eventInfo, onSuccess, onError }: MutationP
...
+}
+
+export function useAdHocChecks(id?: number) {
+  const [adHocChecks] = useSessionStorage<AdHocCheckCache[]>('adHocChecks', []);
+
+  const data = adHocChecks.filter((item) => item.id === id);
+  const isPending = data.some((item) => item.pending);
+
+  return { data, isPending };
+}
```
ckbedwell: Is this used?

src/page/NewCheck/__tests__/BrowserChecks/Scripted/1-script.ui.test.tsx
```diff
@@ -47,17 +50,21 @@ describe(`BrowserCheck - 1 (Script) UI`, () => {
...
+      
+      await fillMandatoryFields({ user, fieldsToOmit: [], checkType });
 
       await submitForm(user);
       const err = await screen.findByText('Script does not export any options.');
       expect(err).toBeInTheDocument();
     });
 
-    it(`will display an error when it does set the browser type to 'chromium'`, async () => {
+    it(`will display an error when it does NOT set the browser type to 'chromium'`, async () => {
```
ckbedwell, line 61: 👏

src/test/dataTestIds.ts
```diff
@@ -27,5 +29,5 @@ export enum DataTestIds {
   TEST_ROUTER_INFO = 'test-router-info',
   TEST_ROUTER_INFO_PATHNAME = 'test-router-info-pathname',
   TEST_ROUTER_INFO_SEARCH = 'test-router-info-search',
-  FREQUENCY_COMPONENT = 'frequency-component',
+  FORM_SIDEBAR = 'form-sidebar',
```
ckbedwell: I love that you moved `FREQUENCY_COMPONENT` to be alphabetical but didn't bother with this one 😛

ckbedwell (dismissed):
I'm approving but I have some super super small nitty comments.

Great work. Love these changes 👏

w1kman:
While working on the 3rd step (right aside and `CheckEditor`), I noticed some bugs that I believe is in this PR as well. I will change it into draft so that it doesn't get merged.

src/components/CheckForm/FormLayout/FormLayout.tsx
w1kman: Yes, it's required for the "stepper"

src/data/useChecks.ts
w1kman: Nope - It is leftovers from the initial "right aside" branch

src/test/dataTestIds.ts
w1kman: I may have sorted them before I started to add stuff 🙄 
It's sorted again.

VikaCep (changes requested):
I'm not able to test scripted or browser checks, when I click the Test button I get sent to the script step but the modal doesn't show up

https://github.com/user-attachments/assets/39696581-3d8f-4d52-80f6-8160432031c4

PR #1239, opened 2025-08-26
chore: converted ping dashboard to react scenes

ckbedwell:
## Problem

In order to put the Timepoint Explorer on the ping dashboard, it needs to use `@grafana/scenes-react`.

## Solution

Convert the dashboard over to use `@grafana/scenes-react`. It was quite simple as it has a lot of similarities to the `HttpDashboard` 🎉


src/scenes/Common/ResponseLatency.tsx
```diff
@@ -6,14 +6,18 @@ import { GraphDrawStyle, StackingMode, ThresholdsMode } from '@grafana/schema';
 import { useMetricsDS } from 'hooks/useMetricsDS';
 import { useVizPanelMenu } from 'scenes/Common/useVizPanelMenu';
 
-export const ResponseLatency = () => {
+interface ResponseLatencyProps {
+  metric: `probe_http_duration_seconds` | `probe_icmp_duration_seconds`;
+}
+
+export const ResponseLatency = ({ metric }: ResponseLatencyProps) => {
```
VikaCep, line 13: There are a few differences in the Ping dashboard for this panel:

with old scenes: 
image

with scenes react: 
image

We should set the unit to `s` for this Ping viz (might not apply to HTTP)

```javascript
  const viz = VizConfigBuilders.timeseries()
    .setUnit('s')
```
And if we want to make it look the same, we should move the legend to the right

```javascript
    .setOption('legend', {
      showLegend: true,
      displayMode: LegendDisplayMode.Table,
      placement: 'right',
    })
```

VikaCep (changes requested):
We should change the `Response latency by phase` unit to seconds.

src/scenes/Common/ResponseLatency.tsx
ckbedwell: Good catch! I've set it as seconds and it is sensible enough that when it is a low number it automatically shows milliseconds. 

I was thinking the only reason to have the right-aligned legend is when you add the additional values besides the legend so I've added them, too! Good catch 🧤

PR #1241, opened 2025-08-26
chore: convert tcp dashboard over

ckbedwell:
## Problem

In order to put the Timepoint Explorer on the TCP dashboard, it needs to use `@grafana/scenes-react`.

## Solution

Convert the dashboard over to use `@grafana/scenes-react`. It was quite simple as it has a lot of similarities to the `HttpDashboard` 🎉


VikaCep (approved):
LGTM 🎉

PR #1246, opened 2025-08-27
feat: add lightweight feedback feature

ckbedwell:
## Problem

We need a way to collect quantitative and qualitative feedback from users in a way that is as non-blocking to their experience as possible.

## Solution

Implemented a lightweight-feedback component that utilises Rudderstack events to track feedback.

Users can just click the thumbs up / down buttons (as many times as they like...) and afterwards they are offered the opportunity to provide additional comments if they so wish.

For this first implementation I have added it to the Alerts per-check feature during the check edit / creation process.

The next candidate will be:
- Timepoint Explorer
- RHS bar in check creation

**Video demonstration of the RudderStack events:**

https://github.com/user-attachments/assets/eeb1ede6-cf07-4c83-8856-e25367db0ba2


VikaCep (approved):
Nice!! 💯

PR #1249, opened 2025-08-28
chore: add gRPC dashboard for react scenes and remove unused scenes components

ckbedwell:
## Problem

In order to put the Timepoint Explorer on the gRPC dashboard, it needs to use `@grafana/scenes-react`.

## Solution

Convert the dashboard over to use `@grafana/scenes-react`. It was quite simple as it has a lot of similarities to the `TcpDashboard` 🎉

As this is the last dashboard that needed converting, I took the opportunity to remove all of the unused components that were left over from the old dashboard implementation.


VikaCep (approved):
LGTM!

PR #1250, opened 2025-08-28
Feat/high level metrics raised

ckbedwell:
## Problem

With the introduction of the Timepoint Explorer the check dashboards need to be modified so there is a high-level view of the metrics above it before entering a more detailed view.

## Solution

This PR adds Uptime, Reachability, Average Latency, Frequency and in some check type cases other important metrics above the Timepoint Explorer to provide a high-level overview of the check's health.

This PR begins to unify and standardise the check dashboards. It is meant to be a step towards making the dashboards easier to parse and engage with.

It also includes:
- A responsive dashboard header bar
- It begins to lay the foundations so that the metric panels beneath the Timepoint Explorer will be responsive (hence the change from `Stack` / `Grid` components to divs with styles, so we can invoke container queries.
- It improves the annotation filtering experience so non-relevant alerting annotations are no longer rendered. It also lays the groundwork so we can put a CTA for setting up alerts if they are not enabled.

image

A **HTTP** dashboard showing the high-level metrics at the top and with the responsive dashboard controls kicking in.

image

**Ping** dashboard with no alerts set up.

image

For the **Traceroute** dashboard I have added the high-level metrics (they weren't there originally) and decided the node graph should go above the Timepoint Explorer.

image

**DNS** dashboard. Moved the answer records stat to the top.

image

**Scripted** and **MultiHttp** dashboard with standardised high-level metrics at the top.

image
Similar with **Browser**.


w1kman (approved):
Luv it! 👨🏻‍🍳

PR #1251, opened 2025-08-29
fix: update probe mappings

ckbedwell:
## Problem

Switzerland was recently added as a new region and the app is lacking the hard-coded mapping: https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/set-up/set-up-private-probes/#add-a-new-probe-in-your-grafana-instance

## Solution

Ran the script to update the mapping 😄


VikaCep (approved):
🏆 🎉 🥇

Issue #1243, opened 2025-08-26
Reject scripts that import extensions or include pragmas


VikaCep:
From the design doc:

>Both the frontend and the API, the latter by means of [k6deps](https://github.com/grafana/k6deps), should analyze scripts and reject those that import extensions, or that include version pragmas that would change the k6 version being used.

PR #1253, opened 2025-09-01
feat: add validation for k6 extensions and version pragmas

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1243

As we're implementing Version Management, check versions will be selected by choosing the corresponding channel from the k6 channel [selector](https://github.com/grafana/synthetic-monitoring-app/pull/1236) (see [design doc](https://docs.google.com/document/d/1ZV70lIZ-xOH9EiAAp7OW2LGVeGwgHYJELdcvsSiH7UU/edit?pli=1&tab=t.ns59gkttkrnx#heading=h.vfc0jwc7ku0x) and slack [conversation](https://raintank-corp.slack.com/archives/C08RWB55049/p1753193562552899?thread_ts=1752864991.996529&cid=C08RWB55049))
This means setting the version restriction in the script is not encouraged, in fact it's not valid (at least for now) as the backend won't honor it. For this reason I'm adding the respective UI validation:

- Error message when importing k6 extensions
image

- Error message when using k6 pragmas
image


src/schemas/forms/script/rules.ts
```diff
@@ -0,0 +1,50 @@
...
+  'Script imports k6 extensions which are not allowed. Please remove imports from k6/x/ paths.';
+
+// Match patterns like: "use k6 >= v1.0.0", "use k6 > 0.52", `use k6 >= v1.0.0`, etc.
+export const K6_PRAGMA_REGEX = /["'`]use\s+k6\s*[><=!]+\s*v?\d+/gi;
+
+// Match import statements that include k6/x/ paths
+export const K6_EXTENSION_IMPORT_REGEX = /import\s+.*from\s*[\'"`][^'"`]*k6\/x\/[^'"`]*[\'"`]/gi;
+
+export function hasK6Pragma(script: string): boolean {
+  K6_PRAGMA_REGEX.lastIndex = 0;
```
ckbedwell: TIL that regex has global state.

Could be worth looking at how to refactor this file to be more functional in nature? I had to look up why we needed to do this.

ckbedwell (dismissed):
I've left a comment but it's quite small so I'm going to approve and leave it to you to decide if you want to address it or not 🙂

src/schemas/forms/script/rules.ts
```diff
@@ -0,0 +1,50 @@
+// Shared script validation rules for both Zod validation and Monaco markers
+
+export const K6_PRAGMA_MESSAGE =
+  'Version directives cannot be used within scripts. Please remove any "use k6" statements.';
+
+export const K6_EXTENSION_MESSAGE =
+  'Script imports k6 extensions which are not allowed. Please remove imports from k6/x/ paths.';
+
+// Match patterns like: "use k6 >= v1.0.0", "use k6 > 0.52", `use k6 >= v1.0.0`, etc.
+export const K6_PRAGMA_REGEX = /["'`]use\s+k6\s*[><=!]+\s*v?\d+/gi;
```
ckbedwell: Can we change this regex to this so it captures specific version numbers? Currently I found the squigly ends prematurely.

```
const K6_PRAGMA_PATTERN = /["'`]use\s+k6\s*[><=!]+\s*v?\d+(?:\.\d*)*(?:[-+][\w.]+)*["'`]/i;
```

Before

Image

After
Image

ckbedwell: Also, it doesn't seem to be used in any other file so doesn't need exporting.

VikaCep:
I've addressed your comments @ckbedwell

w1kman (changes requested):
I don't feel that this change actually checks for JavaScript directives, rather it uses a regular expression to find a match.

Image

`monaco` is way too advanced for not allowing to check specifically for directives, rather than matches, anywhere.

Also, I'd expect the example directives in `k6deps` to trigger the error.

Image

w1kman:
Seems possible. Did a simple validation rule for "use k6" (It does not apply to version req).

image

src/components/CodeEditor/monacoValidation.test.ts
ckbedwell, line 1: ❤️  these tests ❤️

PR #1259, opened 2025-09-03
feat: add custom validation to monaco

ckbedwell:
@VikaCep as discussed.

Please review the code because I got Cursor to write it all 😂


VikaCep:
Thanks @ckbedwell! I like these changes and I think we can use this approach in several other validation scenarios. I made some small changes and will merge it the base branch :)

PR #1271, opened 2025-09-23
fix: display latency visualization units in milliseconds

VikaCep:
Closes https://github.com/grafana/support-escalations/issues/18505

### Problem
Latency visualizations were displaying inconsistent units:
- `AvgLatencyViz` showed seconds with unrealistic 1s max and 2s red threshold
- `ResponseLatencyByProbe` showed seconds with 80s red threshold 
- Other components use milliseconds

### Solution
- Changed both components to use `ms` unit 
- Updated thresholds to realistic millisecond values:
  - AvgLatency: 1000ms yellow, 2000ms red, 5000ms max
  - ResponseLatencyByProbe: 800ms red (aligns with TTFB standards)

### DIFF
--- a/src/scenes/Common/AvgLatencyViz.tsx
+++ b/src/scenes/Common/AvgLatencyViz.tsx
@@ -87,10 +87,10 @@ export const AvgLatency = () => {
 
   const viz = VizConfigBuilders.stat()
     .setOption('graphMode', BigValueGraphMode.None)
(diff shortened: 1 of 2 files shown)


VikaCep:
Closing as this is not needed. The `s` unit is correct. See https://github.com/grafana/support-escalations/issues/18505#issuecomment-3339918953

PR #1272, opened 2025-09-24
fix: add probe api mapping for eu-west-5

nadiamoe:

Adds mapping for prod-eu-west-5

Fixes #


nadiamoe:
Seems like this file is built automatically _from_ the docs, so manual changes don't make sense here. Closing and opening an issue instead.

PR #1273, opened 2025-09-24
fix: upgrade Grafana dependencies to resolve API compatibility check

VikaCep:
## What this PR does
Fixes the Grafana API compatibility check failure by upgrading Grafana dependencies to their latest compatible versions.

## Problem
The levitate compatibility check was failing because:
- Local dependencies were locked to `12.1.0` in yarn.lock
- The compatibility check tests against latest versions (`12.2.0`)
- Found 2 incompatibilities in `@grafana/runtime` between versions 12.1.0 → 12.2.0
- Created a mismatch between what we're developing against vs. what we're testing compatibility with

## Solution
Upgraded Grafana dependencies from 12.1.0 to 12.2.0:
- `@grafana/data`: `12.1.0` → `12.2.0`
- `@grafana/runtime`: `12.1.0` → `12.2.0` 
- `@grafana/schema`: `12.1.0` → `12.2.0`
- `@grafana/ui`: `12.1.0` → `12.2.0`

### DIFF
--- a/src/components/Clipboard/CopyToClipboard.tsx
+++ b/src/components/Clipboard/CopyToClipboard.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { Button, ButtonVariant, IconButton } from '@grafana/ui';
 
-interface CopyToClipboardProps extends React.ComponentProps<typeof Button> {
+interface CopyToClipboardProps {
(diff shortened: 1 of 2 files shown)


ckbedwell (approved):
LGTM

PR #1276, opened 2025-09-25
fix: adapt terraform config to runbook_url

VikaCep:
After the latest terraform provider release (https://registry.terraform.io/providers/grafana/grafana/4.10.0/docs/resources/synthetic_monitoring_check_alerts) we now need to adjust the app's terraform configuration to add the `runbook_url` field to per-check alerts resources. 

image
image


ckbedwell (approved):
✅

PR #1278, opened 2025-10-02
Feat: add Terraform HCL format

VikaCep:
### Add HCL Export Support to Terraform Configuration

This PR adds support for exporting Terraform configurations in HCL format alongside the existing JSON format, providing users with more flexibility in their Terraform workflows.

Users can now export their synthetic monitoring checks, probes and alerts in native HCL format through the Terraform configuration tab. The interface includes a tabbed view where users can switch between HCL and JSON formats, with HCL as the default option.

https://github.com/user-attachments/assets/377766eb-e973-49b4-aae0-c20d1fc555d7


src/page/ConfigPageLayout/tabs/TerraformTab.tsx
```diff
@@ -60,34 +64,84 @@ export function TerraformTab() {
...
+                </TextLink>
+                , with their respective value.
+              </Text>
+              <Clipboard
+                highlight={['<GRAFANA_SERVICE_TOKEN>', '<SM_ACCESS_TOKEN>']}
+                content={JSON.stringify(config, null, 2)}
+                className={styles.clipboard}
+                isCode
+              />
+            </>
```
ckbedwell: Isn't this identical to the one above except the title, link, extension and clipboard content? Probably worth abstracting to a component with those props.

PR #1279, opened 2025-10-03
chore: update default script examples

peterschretlen:

This PR updates the default scripted and browser examples to:
- Use quickpizza login as the example for both scripts
- Advertise k6 studio for recording browser scripts
- Advertise secrets management
- Modernize by replacing `fail` with `expect` (assertions library)

Fixes https://github.com/grafana/synthetic-monitoring/issues/358

My main concern is this makes the default scripts longer and harder to read.

Scripted Before
image

Scripted After
image

Browser Before
image

Browser After
image

### DIFF
--- a/src/components/constants.ts
+++ b/src/components/constants.ts
@@ -226,7 +226,9 @@ export default async function () {
 
     // TIP: Use check() to report test results in the 'assertions' dashboard panel
     // Scripts continue to run even if a check fails. Failed checks don't impact uptime and reachability
-    await check(heading, async (h) => (await h.textContent()) == "Latest pizza recommendations");
+    await check(heading, {
+      ["Header is present"]: async (h) => (await h.textContent()) == "Latest pizza recommendations"
+    });
 
   } catch (e) {
     console.log('Error during execution:', e);


src/components/constants.ts
```diff
@@ -148,64 +148,82 @@ export const TIME_UNIT_OPTIONS = [
 export const TEN_MINUTES_IN_MS = 1000 * 60 * 10;
 export const FIVE_MINUTES_IN_MS = 1000 * 60 * 5;
 
-const EXAMPLE_SCRIPT_SCRIPTED = btoa(`import { check, fail } from 'k6'
+const EXAMPLE_SCRIPT_SCRIPTED = btoa(`import { expect } from 'https://jslib.k6.io/k6-testing/0.5.0/index.js';
+import { check } from 'k6'
 import http from 'k6/http'
+import secrets from 'k6/secrets';
```
w1kman: Feels like this should be commented as it is not used in the script.

```suggestion
// import secrets from 'k6/secrets';
```

src/components/constants.ts
```diff
@@ -148,64 +148,82 @@ export const TIME_UNIT_OPTIONS = [
...
+  const authURL = 'https://quickpizza.grafana.com/api/users/token/login';
+
+  // TIP: Secure your credentials using secrets.get()
+  // https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/create-checks/manage-secrets/
+  const authInfo = {
+    username: 'default', // username: await secrets.get('quickpizza-username');
+    password: '12345678' // password: await secrets.get('quickpizza-password');
   }
+
+  const resp = http.post( authURL, JSON.stringify(authInfo), { headers: { 'Content-Type': 'application/json' }});
```
w1kman: IMO: For readability, I'd move the headers to a `const`

```suggestion
  const headers = {
    headers: { 'Content-Type': 'application/json' },
  };

  const resp = http.post(authURL, JSON.stringify(authInfo), headers);
```

src/components/constants.ts
peterschretlen: Updated - thank you!

src/components/constants.ts
peterschretlen: Updated - thank you!

Issue #1263, opened 2025-09-08
Add `label_` prefix explanation during check creation


ka3de:
**What would you like to be added**:
A brief explanation on the `label_` prefix being applied to custom labels for Synthetic Monitoring.
This should probably be added in the current text that describes implications of custom labels in the `Labels` tab during the check creation flow.

<img width="767" height="476" alt="Image" src="https://github.com/user-attachments/assets/eb8e1b4f-ed9b-432b-b9c6-a463c9654e5d" />

**Why is this needed**:
Provides users with better context of the implications of custom labels usage directly from the app.

**Additional context**
See related [thread](https://raintank-corp.slack.com/archives/C0162C1K9E1/p1757294471974569?thread_ts=1756884649.807499&cid=C0162C1K9E1).

Issue #1174, opened 2025-07-04
Ad-hoc runs that do not return any logs are reported as failed


nadiamoe:
A script that does not log anything is inconsistently reported as failed in the UI:

```js
import { sleep } from 'k6'

export default function main() {
  sleep(1);
}
```

<img width="1440" height="877" alt="Image" src="https://github.com/user-attachments/assets/3e1d025a-549a-4628-bd68-06658dc0d824" />

<img width="1058" height="500" alt="Image" src="https://github.com/user-attachments/assets/17495bfc-e571-4fab-9b64-a7a90f9d91ea" />

We probably haven't encountered this in the wild much as it is hard to craft a valid script that does not log anything, but I did encounter this with a valid browser script. Took me a while to figure out why the script was "failing".

To be 100% fair, I'm not sure if this is a plugin issue or an agent issue. I'm leaning towards the former due to the green check, but feel free to correct me here.

peterschretlen:
See related issue https://github.com/grafana/synthetic-monitoring-app/issues/953

Issue #953, opened 2024-10-04
Browser synthetics - test button shows "error" when results don't contain any logs


peterschretlen:
**What happened**:

When using the test button, the test result will show "there was an error running the test" if the test does not generate any logs

![Image](https://github.com/user-attachments/assets/0ed70069-c4d8-4451-9eb7-f10349c4b449)

If I add a `console.log` to the script, the error goes away:

![Image](https://github.com/user-attachments/assets/62c995b2-a608-445c-aa39-9eb08eb75bba)

**What you expected to happen**:

There should be no error

**How to reproduce it (as minimally and precisely as possible)**:

This test (a no-op script) will produce an error
```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
};

export default async function () {

}
```

This test (adds a console.log) has no error
```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
};

export default async function () {
  console.log("adding a console log fixes the error");
}
```

We seem to have an assumption that logs will be returned:
https://github.com/grafana/synthetic-monitoring-app/blob/26d0c5b8889b70bb98e2b9e2725a563e9afda74b/src/components/CheckTestResultsModal.tsx#L106-L109

PR #1280, opened 2025-10-06
feat: new check editor

w1kman:
# Reworked check editor for synthetic checks

## Complete re-creation of the current check Form.

Why? As we require more features and customization of the Check form, it has become increasingly complex to do the simplest changes. In an attempt to simplify and surface inner workings, "everything" has been remade.

The name "Checkster" is a placeholder name that will later be refactored into "CheckEditor", as soon as we have removed the old "CheckEditor".

### Features

This PR also brings new features to the check authoring view:
- New adhoc check experience
- Docs 
- Secrets (CRUD) while authoring scripted/browser checks

### How to test

- Enable feature flag `synthetic-monitoring-check-editor`
- Use the app as normal

image

## Adhoc check example (blackbox exporter)
image

## Adhoc check example (k6/multihttp)
image

## Adhoc check example (k6)
image

### Additional context

Resolves https://github.com/grafana/synthetic-monitoring-app/issues/1263
Resolves https://github.com/grafana/synthetic-monitoring-app/issues/1174
Resolves https://github.com/grafana/synthetic-monitoring-app/issues/953
Resolves https://github.com/grafana/synthetic-monitoring-app/issues/1133

### DIFF
--- a/src/hooks/useHandleSubmitCheckster.ts
+++ b/src/hooks/useHandleSubmitCheckster.ts
@@ -45,7 +45,7 @@ export function useHandleSubmitCheckster(initialCheck?: Check) {
       await handleAlerts(result, alertsEnabled ? formValues.alerts : undefined);
       await queryClient.invalidateQueries({ queryKey: queryKeys.list });
 
-      return () => navigateToCheckDashboard(result, payload?.id !== undefined);
+      return () => navigateToCheckDashboard(result, payload?.id === undefined);
     },
     [alertsEnabled, initialCheck, createCheck, handleAlerts, navigateToCheckDashboard, updateCheck]
   );


ckbedwell:
https://github.com/user-attachments/assets/871d9d79-b88f-4ebd-bc4f-886c416ba3da

Also found that for MultiStep for query parameters it didn't go to the right tab with the error.

https://github.com/user-attachments/assets/668a7c15-09e5-4518-b440-4cda44424700

When the error was in the variable dropdown:
1. The wizard form step didn't register it was invalid
2. It didn't go to the step and/or open the variables dropdown.

https://github.com/user-attachments/assets/be382193-948d-447e-a6f4-ecbdef5c9930

In MultiStep when I add several assertions, it doesn't scroll down to the one with the error either.

src/components/CheckEditor/ProbeOptions.tsx
```diff
@@ -10,9 +10,10 @@ import { CheckProbes } from './CheckProbes/CheckProbes';
 
 interface ProbeOptionsProps {
   checkType: CheckType;
+  onlyProbes?: boolean;
```
ckbedwell: Can we make a note that we can remove this prop after CheckEditorV1 is removed?

src/components/Checkster/Checkster.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { AppContainer } from './components/ui/AppContainer';
+import { PrimaryLayoutSection } from './components/ui/PrimaryLayoutSection';
+import { SecondaryLayoutSection } from './components/ui/SecondaryLayoutSection';
+import { InternalConditionalProvider, useChecksterContext } from './contexts/ChecksterContext';
+import { FeatureTabsContextProvider } from './contexts/FeatureTabsContext';
+import { FeatureContent } from './feature/FeatureContent';
+import { FeatureTabs } from './feature/FeatureTabs';
+
+interface ChecksterProps {
+  checkOrCheckType?: Check | CheckType;
```
ckbedwell: Is there any reason we couldn't have this as two separate props? I find it somewhat difficult when I follow it through to the context provider what it is doing.

src/components/Checkster/components/CollapsibleRequestEntry.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { HttpMethod } from 'types';
+import { getMethodColor } from 'utils';
+
+import { ErrorIcon } from './ErrorIcon';
+
+interface CollapsibleRequestEntryProps extends PropsWithChildren {
+  method: HttpMethod;
+  target?: string;
+  placeholder?: string;
+  open: boolean;
```
ckbedwell: nit: is there any reason not to just call this prop`isOpen` to skip the renaming in the destructuring step below?

src/components/Checkster/components/CollapsibleRequestEntry.tsx
```diff
@@ -0,0 +1,86 @@
...
+      align-items: center;
+      justify-content: space-between;
+      padding-right: ${theme.spacing(1)};
+
+      &:hover {
+        background-color: ${theme.colors.background.secondary};
+      }
+
+      // Style by proxy
+      & > h4 {
```
ckbedwell, line 65: 😃 🔫 

Raised my eyebrows at this, reread the code then realised it was just a rehash of the conversation we had this morning 🤦 

I'm putting that eslint rule in because being forced to do this kind of thing is so dumb.

src/components/Checkster/components/FormSectionNavigation/FormSectionNavigation.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const {
+    isNew,
+    checkType,
+    formNavigation: { sectionOrder, setSectionActive, isSectionActive, getSectionFields, isSeenStep, getSectionLabel },
+  } = useChecksterContext();
+  const styles = useStyles2(getStyles);
+  const allErrors = useSilentErrors();
+
+  return (
+    <ol aria-label="Check form navigation" role="navigation" className={styles.container}>
```
ckbedwell: Because they aren't links this wouldn't be navigation but probably a tablist?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tablist_role

src/components/Checkster/components/FormSectionNavigation/FormSectionNavigation.tsx
```diff
@@ -0,0 +1,154 @@
...
+    `,
+  };
+}
+
+function getStyles(theme: GrafanaTheme2) {
+  const breakpoint = theme.breakpoints.values.md;
+  const query = `(max-width: ${breakpoint}px)`;
+  const containerQuery = `@container ${CSS_PRIMARY_CONTAINER_NAME} ${query}`;
+
+  const isActive = css``;
```
ckbedwell: ✂️ ?

src/components/Checkster/components/QueryParamsEditor.tsx
```diff
@@ -0,0 +1,261 @@
...
+  } catch (e) {
+    try {
+      const firstIndexOfQuestionMark = url.indexOf('?');
+      const search = firstIndexOfQuestionMark !== -1 ? url.substring(firstIndexOfQuestionMark + 1) : '';
+      return getSearchParamEntries(`https://www.grafana.com?${search}`);
+    } catch (e) {
+      return [];
+    }
+  }
+}
```
ckbedwell: nit: Can we move these to `QueryParamsEditor.utils.ts`? I used that pattern quite extensively in the Time Point Explorer and found it quite a nice pattern.

src/components/Checkster/components/QueryParamsEditor.tsx
ckbedwell, line 1: Scope creeeeeeep feature 😛 (works great, though)

src/components/Checkster/components/form/FormFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+          gap: ${theme.spacing(1)};
+        `}
+      >
+        {next && (
+          <Button
+            type="button"
+            variant={isStepsComplete ? 'secondary' : 'primary'}
+            onClick={() => setSectionActive(next.name)}
+          >
+            {/* since Button.iconPosition=right does nothing */}
```
ckbedwell: Works for me... 🤔 

image

src/components/Checkster/components/form/FormFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+    formNavigation: {
+      stepActions: { previous, next },
+      setSectionActive,
+      isStepsComplete,
+    },
+  } = useChecksterContext();
+
+  return (
+    <div
+      className={css`
```
ckbedwell, line 27: This gets quite nitty but my preference here is we stay consistent with creating a styles object and always following the pattern of `className={styles.myLovelyClassName}`.

I find it noisy when I have to read inline styles when  my focus is on the semantic structure when reading the file (HTML/JSX). Could do this as a separate follow-up PR and see if we could get some sort of AI automation to clean it up across the whole codebase? Maybe even see if there is a linting rule we could add?

src/components/Checkster/components/form/FormHttpRequestMethodTargetFields.tsx
ckbedwell, line 1: I found the request methods is inaccessible when using keyboard for this component. My expectation would be to be able to start on the request methods and then be able to tab to the request target field.

src/components/Checkster/components/form/FormRoot.tsx
```diff
@@ -0,0 +1,101 @@
...
+  // Wait for formState to be reset before calling onSave callback
+  useEffect(() => {
+    if (!isDirty && typeof onSaveCallback === 'function') {
+      onSaveCallback();
+    }
+  }, [onSaveCallback, isDirty]);
+
+  return (
+    <form
+      onSubmit={handleSubmit(
```
ckbedwell: nit: pull out the two arguments here to nicely named `onValid` and `onInValid` callbacks to pass in? At a glance it looks like one long inline function.

src/components/Checkster/components/form/FormTabs.tsx
```diff
@@ -0,0 +1,152 @@
...
+                css`
+                  flex: 1 1 0;
+                  overflow: auto;
+                  & > div {
+                    flex-grow: 1;
+                    margin-bottom: unset;
+                    overflow: unset;
+                  }
+                `,
+              child.props.className
```
ckbedwell, line 105: The inline css makes this so hard to read. Just reiterating my comment above that we need to standardize on using the styles object and use the `cx` method from `@emotion/css` for the conditional classes.

src/components/Checkster/components/form/layouts/DnsCheckContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function DnsCheckContent() {
+  const hasRequestOptionError = useHasFieldsError(DNS_CHECK_REQUEST_OPTIONS_FIELDS);
+  const tabIndexErrors = useGetIndexFieldError(DNS_REQUEST_OPTIONS_TAB_FIELDS);
+  return (
+    <SectionContent>
+      <FormJobField field="job" />
+      <ChooseCheckType />
+      <GenericInputField
+        field="target"
+        label="Reuqest target"
```
ckbedwell:
```suggestion
        label="Request target"
```

src/components/Checkster/components/form/layouts/TracerouteCheckContent.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { GenericCheckboxField } from '../generic/GenericCheckboxField';
+import { GenericInputField } from '../generic/GenericInputField';
+
+const REQUEST_OPTIONS_TAB_FIELDS = [
+  [/\.headers/], // Options
+  undefined, // Body
+  [/\.basicAuth\./], // Authentication
+  [/\.tlsConfig\./], // TSL
+  [/\.proxyConnectHeaders\.\d+\./], // Proxy
+];
```
ckbedwell, line 16: This looks wrong to me for Traceroute.

src/components/Checkster/components/modals/ChooseCheckTypeModal.tsx
ckbedwell, line 1: Not used, so let's kill it.

src/components/Checkster/components/ui/AppContainer.tsx
```diff
@@ -0,0 +1,81 @@
...
+  `;
+
+  return {
+    wrapper: css`
+      display: flex;
+      flex-direction: column;
+      height: 100%;
+      position: relative;
+    `,
+    // Thoughts?
```
ckbedwell: Subtle enough not to be a distraction so looks good to me. I only noticed the fade in after I saw the code here!

src/components/Checkster/components/ui/Column.tsx
```diff
@@ -0,0 +1,72 @@
...
+          `,
+        overflow !== undefined &&
+          css`
+            overflow: ${overflow};
+          `,
+        padding !== undefined &&
+          css`
+            padding: ${typeof padding === 'number' ? theme.spacing(padding) : padding};
+          `,
+        className
```
ckbedwell: import { cx } from `@emotion/css`

🥺

src/components/Checkster/components/ui/SectionContent.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { css } from '@emotion/css';
+
+import { FIELD_SPACING } from '../../constants';
+import { useChecksterContext } from '../../contexts/ChecksterContext';
+import { Column } from './Column';
+
+interface SectionContentProps extends PropsWithChildren {
+  label?: ReactNode;
+  vanilla?: boolean; // No content styling
+  showHeading?: boolean;
```
ckbedwell: We aren't showing the heading any more so it feels like this component can be simplified a lot. I'd go as far as removing the `vanilla` prop as it is only used in one place and just have a custom div and styles in that case.

I found it a bit confusing that the component changes to a Fragment but still receives props which are now irrelevant. I also have a feeling that will create noise in our tests, too.

src/components/Checkster/components/ui/StyledField.tsx
ckbedwell, line 1: I'd quite like us to add documentation on these components as it's not immediately clear what they are for. I can see it causing confusion in the future so would be nice to have something explicit written down.

src/components/Checkster/components/ui/ValidationWarning.tsx
ckbedwell, line 1: Can't this component and `ValidationError` be a single component which takes a prop whether it is an error or warning level?

src/components/Checkster/components/ui/ValidationError.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface ValidationErrorProps extends PropsWithChildren {
+  className?: string;
+}
+
+/*
+                      <div className={cx(styles.errorRow, isHeaderMatchType && styles.expressionWithHeaderErrorColumn)}>
+                        <Icon name="exclamation-triangle" />{errorMap.expression.error}
+                      </div>
+ */
```
ckbedwell: ✂️ ?

src/components/Checkster/constants.ts
```diff
@@ -0,0 +1,615 @@
...
+    await check(page.locator("h2"), {
+      header: async (locator) => (await locator.textContent()) == "Welcome, admin!",
+    });
+  } catch (e) {
+    console.log('Error during execution:', e);
+    throw e;
+  } finally {
+    await page.close();
+  }
+}`);
```
ckbedwell: These need updating.

src/components/Checkster/constants.ts
```diff
@@ -0,0 +1,615 @@
...
+
+export const DEFAULT_CHECK_TYPE: CheckType = CheckType.HTTP;
+
+const CHECK_BASE_CONFIG: Omit<Check, 'settings'> = {
+  job: '',
+  target: '',
+  frequency: 60 * 1000,
+  timeout: 3 * 1000,
```
ckbedwell: My preference would be to make these clearer by extracting them to their own constants or (well, and...) have tests that are explicit about what values they should be. 

We've had issues in the past where they haven't been kept in sync so the more clarity we can provide in the code the easier it is to identify updating them and ensuring they are maintained correctly.

e.g.

```
  frequency: ONE_MINUTE_IN_MS,
  timeout: THREE_SECONDS_IN_MS,
```

src/components/Checkster/contexts/FeatureTabsContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+interface FeatureTabsContextValue {
+  setActive: (label: FeatureTabLabel) => void;
+  tabs: FeatureTabConfig[];
+  activeTab: FeatureTabConfig;
+}
+
+export const FeatureTabsContext = createContext<FeatureTabsContextValue | undefined>(undefined);
+
+// In case nothing adds up
+const panicTab = ['', null, []];
```
ckbedwell:
```suggestion
const panicTab: FeatureTabConfig = ['', null, []];
```

☝️  It would be nice to add the explicit type here as I wanted to know what format it was following but worth noting this actually causes a type error because `null` is not compatible with `ComponentType`. It doesn't get picked up in the component further down because of the interaction between the typed array and the nulish coalescing operator means it always evaluates the type to what is in the array.

src/components/Checkster/feature/adhoc-check/AdhocCheckPanel.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  if (isLoadingProbes) {
+    return <div>Loading probes...</div>;
+  }
+
+  if (!items.length) {
+    return (
+      <div
+        className={css`
+          padding: 8px 8px 8px 0;
```
ckbedwell: `theme.spacing` instead here, ploise (and move it a styles object because 👎  to the inline css).

src/components/Checkster/feature/adhoc-check/AdhocCheckPanel.tsx
```diff
@@ -0,0 +1,225 @@
...
+        </EmptyState>
+      </div>
+    );
+  }
+
+  return (
+    <Column
+      gap={2}
+      className={css`
+        padding: 8px 8px 8px 0;
```
ckbedwell: Same here.

src/components/Checkster/feature/adhoc-check/AdhocCheckPanel.tsx
ckbedwell, line 1: Similar to another component I commented on, I'd be tempted to follow the pattern of putting all the hooks logic in `AdhockCheckPanel.hooks.ts`. It's quite a complicated component so it's nice to have the separation.

Maybe for another PR because I think we are going to move where these results are stored?

src/components/Checkster/feature/adhoc-check/LogDetails.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Preformatted>
+      {content && typeof content === 'object'
+        ? Object.entries(content).map(([key, value]) => {
+            return (
+              <div key={key}>
+                {key}:{' '}
+                <span
+                  className={css`
+                    opacity: 0.6;
```
ckbedwell: Feels like this should be using the secondary text colours from the theme rather than adding the opacity. Mostly concerned that because there are multiple themes now we add a much bigger burden to ourselves if we stray away from the provided theming system.

src/components/Checkster/feature/adhoc-check/LogsPanel.tsx
```diff
@@ -0,0 +1,130 @@
...
+        </div>
+      )}
+    </div>
+  );
+}
+
+function getStyles(theme: GrafanaTheme2) {
+  const breathAnimation = keyframes`
+    0% {
+      box-shadow: 0 0 0 4px rgba(255, 255, 255, 0.3)
```
ckbedwell: Because of the hard-coded colour this doesn't appear correctly in the light themes. Can either question if the theme is light or not or look through the theme and see if there is an appropriate value to change it to?

src/components/Checkster/hooks/useCheckMeta.ts
ckbedwell, line 1: Not used. Delete?

src/components/Checkster/hooks/useLiveErrors.ts
ckbedwell, line 1: When this is used it is mostly referred to as `liveErrors`. Can we rename this to `useLiveErrors.ts` - that makes more sense to me.

src/components/Checkster/components/InputSelect.tsx
```diff
@@ -0,0 +1,164 @@
...
+      &[disabled] {
+        cursor: not-allowed;
+        background-color: ${theme.colors.action.disabledBackground};
+        color: ${theme.colors.action.disabledText};
+        border-color: ${theme.colors.action.disabledBackground};
+      }
+      &:focus {
+        ${getInputFocusStyles(theme)};
+      }
+      ,
```
ckbedwell: According to my IDE these commas are invalid syntax.

src/components/Checkster/styles.ts
```diff
@@ -0,0 +1,48 @@
...
+import { css } from '@emotion/css';
+
+export function getBodySmallStyles(theme: GrafanaTheme2) {
+  return css`
+    font-size: ${theme.typography.bodySmall.fontSize};
+    line-height: ${theme.typography.bodySmall.lineHeight};
+  `;
+}
+
+export function getInputFocusStyles(theme: GrafanaTheme2) {
```
ckbedwell: `@grafana/ui` provides `styleMixins.getFocusStyles(theme)` -- can't that be used instead?

I also ended up removing this and couldn't tell the difference, it looked like it was all provided by the runtime cascade any way?

src/components/Checkster/styles.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    // Change code snippet menu background
+    & section > div > div {
+      background-color: ${theme.colors.background.primary};
+    }
+  `;
+}
+
+// a11y class for hiding elements visually, whilst keeping the element for screen readers.
+export const visuallyHidden = css`
```
ckbedwell: Not used so ✂️

src/components/Checkster/styles.ts
```diff
@@ -0,0 +1,48 @@
+import { GrafanaTheme2 } from '@grafana/data';
+import { css } from '@emotion/css';
+
+export function getBodySmallStyles(theme: GrafanaTheme2) {
```
ckbedwell: Most of the places I saw using this could be replaced by `<Text variant="bodySmall">` and those that couldn't I'd just add the two lines to their custom class.

src/components/Checkster/styles.ts
```diff
@@ -0,0 +1,48 @@
...
+      outline-offset: 2px;
+      box-shadow: 0 0 0 2px ${theme.colors.background.canvas}, 0 0 0px 4px ${theme.colors.primary.main};
+      transition-timing-function: cubic-bezier(0.19, 1, 0.22, 1);
+      transition-duration: 0.2s;
+      transition-property: outline, outline-offset, box-shadow;
+    }
+  `;
+}
+
+export function codeSnippetWrapper(theme: GrafanaTheme2) {
```
ckbedwell: This is only used by `ScriptedCheckContent` so let's move it there.

src/components/Checkster/styles.ts
ckbedwell, line 1: So, the comments in this file are essentially asking to kill it. I don't want to get into the habit of having something like this as it's going to make our styling consistency with the rest of Grafana much more difficult as we move forwards.

src/components/Checkster/types.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export enum HTTPAuthType {
+  None = 'none',
+  BasicAuth = 'basic-auth',
+  BearerToken = 'bearer-token',
+}
+
+// End of `CheckEditor.types.ts`
+
+export enum RightAsideFeatures {
```
ckbedwell: ✂️ (not used)

src/components/Checkster/types.ts
```diff
@@ -0,0 +1,89 @@
...
+  errors: string[] | undefined;
+}
+
+export enum HTTPAuthType {
+  None = 'none',
+  BasicAuth = 'basic-auth',
+  BearerToken = 'bearer-token',
+}
+
+// End of `CheckEditor.types.ts`
```
ckbedwell: ✂️

src/components/Checkster/utils/adaptors.test.ts
```diff
@@ -0,0 +1,446 @@
...
+  describe('DNS check', () => {
+    it('should call getDNSCheckFormValues for DNS check type', () => {
+      mockGetCheckType.mockReturnValue(CheckType.DNS);
+      mockGetDNSCheckFormValues.mockReturnValue(mockFormValues);
+
+      const result = toFormValues(mockCheck);
+
+      expect(mockGetCheckType).toHaveBeenCalledWith(mockCheck.settings);
+      expect(mockGetDNSCheckFormValues).toHaveBeenCalledWith(mockCheck);
+      expect(result).toBe(mockFormValues);
```
ckbedwell, line 125: Honestly, I see very little value in these tests due to the heavy use of mocking and type casting. It's almost 500 lines which is essentially checking if the switch statements work but not the actual transformations themselves. My expectation when reading this file was that it would be testing the transformations for each check type.

In the first half of the tests every test returns the HTTP check type in the result so it's difficult to have confidence in a test that is focussed on DNS when the input and result has nothing to do with DNS. If you were to keep these tests I'd make it much clearer they are just testing the switch statements and looking at parameterising them so it's just a for loop that is lining up checkType with the transformation function.

src/components/Checkster/utils/check/getDefaultCheckConfig.ts
ckbedwell, line 1: Can we rename this `createDefaultCheck.ts` -- I was expecting this file to be much different to what it actually does 😅

src/components/Checkster/Checkster.tsx
w1kman: The trails ends after the two first lines in the provider scope.

```
  const check = isCheck(checkOrCheckType) ? checkOrCheckType : undefined;
  const [checkType, setCheckType] = useState<CheckType>(
    isCheck(checkOrCheckType) ? getCheckType(checkOrCheckType.settings) : checkOrCheckType ?? DEFAULT_CHECK_TYPE
  );
```

Either way, I will change it into two different props.

src/components/Checkster/components/CollapsibleRequestEntry.tsx
w1kman: Mixed up the convention of `show` and `isOpen`.

src/components/Checkster/components/FormSectionNavigation/FormSectionNavigation.tsx
w1kman: Had to change a lot of stuff for the `tablist` 😩 
Please take a look at the `tablist`, `tab` and `tabpanel` roles associated with the nav.

src/components/Checkster/components/ui/SectionContent.tsx
w1kman: Removed `heading` and `label`.
Renamed `vanilla` to `noWrapper`.

Since `SectionContent` is the `tabpanel` for each section, I'd like to not create an exact copy just to remove a wrapper element. One component to rule them all.

Take a look and see if you like it better now.

w1kman: I changed it into a ternary expression instead of passing props to the `void` (aka `Fragment`).

src/components/Checkster/components/QueryParamsEditor.tsx
w1kman: It's all the jazz 💯 
I tend to write them inline with the component and then move them out. Me cutting corners here.

src/components/Checkster/components/form/FormFooter.tsx
w1kman: Note that I wrote `iconPosition`, not `iconPlacement`. Dunno where I got `iconPosition` from? There is a fair chance my IDE just made up a prop and it didn't show an error when I tried it out. 

Glad I commented, so that you could correct me :D

src/components/Checkster/components/form/FormFooter.tsx
w1kman: Initially the plan was to move the inline `css` usage to a styles object.
After a while, I found it super useful since it gave didn't obscure the styling (guess it's more useful if you need to change something).

I'm not super impressed by the styles object. Perhaps it's just my IDE, but I feel like we have a ton of unused styles because of the fact that every prop on the style object is seemingly used, when in fact, some props are not used at all.

image

**TL;DR: The idea was never to enforce the new fancy way of using `css` inline. We (I) should move them to an object (I'd even go so far to say that styles should follow the same pattern as .test.ts, .utils.ts).** 

> Could do this as a separate follow-up PR

Yes, please

> and see if we could get some sort of AI automation to clean it up across the whole codebase?

Let's ask AI to make css objects actual css, so that we get the benefit of being able to lint css (typescript for styles is kind of meh). 🤞🏻

src/components/Checkster/components/form/FormHttpRequestMethodTargetFields.tsx
w1kman: Was missing tabIndex.

src/components/Checkster/components/ui/Column.tsx
w1kman: `cx` is imported. But I get your point ;)

src/components/Checkster/components/ui/StyledField.tsx
w1kman: There was a `README.md` before, but I scrapped it.
I think we should add a `README.md` to explain how to add/edit new check, what to think about when doing it as well as an overview on the CheckEditor flow.

Not saying we shouldn't add documentation to code, but as an addition to that.

src/components/Checkster/components/ui/ValidationWarning.tsx
w1kman: Errors are handled by Field (or rather `StyledField`).
This is used for a very particular use-case, where a field is not actually invalid, but at the same time it doesn't make any sense.

So far, the only place we use it is for a HTTP check with Body regexp validation for a request without a body (`HEAD`).

image

w1kman: Was a bit fast there. Yes, we could consolidate them :D

src/components/Checkster/constants.ts
w1kman: Just exported the "current" ones from `components/constants.ts` instead.

src/components/Checkster/feature/adhoc-check/LogsPanel.tsx
w1kman: This is an oopsi. Nice catch!

src/components/Checkster/styles.ts
w1kman: There was a difference (if you used a mouse). Tabbed focus was fine.
Either way, the intended goal was for it to look like a MultiSelect, that goal was not met by the styles I had added. I've updated the component to actually a have focus-like look now.

src/components/Checkster/styles.ts
w1kman: I've removed the helper.
Was unable to use `Text` as a replacement though. So now all files that were using the helper have added styles.

You can show me sometime, how to make `Text` be bold and have a specific `line-height`.

Bonus: Found that the wrong `font-weight` was used in a couple of places, that is now fixed.

src/components/Checkster/utils/adaptors.test.ts
w1kman: > It's almost 500 lines which is essentially checking if the switch statements work

That is also exactly what the module does. But I agree, it's not pretty and it can definitely be improved.

Ideally each individual `toPayload.` and `toFormValues.` should have their own unit tests and not be tested via `adaptors.tests.ts`.

src/components/Checkster/utils/check/getDefaultCheckConfig.ts
w1kman: Changed it to `getDefaultCheckConfig`, as that is what it does.

src/components/Checkster/components/form/FormHttpRequestMethodTargetFields.tsx
```diff
@@ -0,0 +1,188 @@
...
+                  url={targetValue}
+                  onDismissWarning={() => {
+                    setShowQueryParams(false);
+                  }}
+                  onChange={handleQueryParamsOnChange}
+                />
+              </StyledField>
+              <Text variant="bodySmall" color="secondary">
+                Looking for <Text color="primary">&#34;Cache busting query parameter&#34;</Text>? It can be found under
+                <Text color="primary">&#34;Request options&#34;</Text>;
```
VikaCep:
```suggestion
                <Text color="primary">&#34;Request options&#34;</Text>
```

src/components/Checkster/components/ui/SectionContent.tsx
```diff
@@ -0,0 +1,42 @@
...
+export function SectionContent({ children, noWrapper }: SectionContentProps) {
+  const theme = useTheme2();
+
+  const {
+    formNavigation: { active },
+  } = useChecksterContext();
+
+  return (
+    <div
+      tabIndex={0}
```
ckbedwell: Because this is focusable I'd expect to be able to use my keyboard to scroll up and down it but because the overflow: auto is on its child that doesn't work.

I think we can safely move the overflow: auto to this component?

src/hooks/useHandleSubmitCheckster.ts
```diff
@@ -0,0 +1,52 @@
...
+          tenantId: initialCheck.tenantId,
+          ...payload,
+        });
+      } else {
+        result = await createCheck(payload);
+      }
+      await handleAlerts(result, alertsEnabled ? formValues.alerts : undefined);
+      await queryClient.invalidateQueries({ queryKey: queryKeys.list });
+
+      return () => navigateToCheckDashboard(result, payload?.id !== undefined);
```
ckbedwell:
```
payload?.id !== undefined
```

I think this should be inverted to `payload?.id === undefined` because it is passing to `isNew`.

Issue #1258, opened 2025-09-02
Per-check alerts: implement notification policies preview


VikaCep:
Currently users have no way to see details about where alerts instances are going to be routed other than a link to the Alerting section. We should implement a solution so that we can show the routing and contact point associated to these alerts. 

The alerting team is [adding](https://github.com/grafana/grafana/pull/108982) the route matching functionality to the `alerting` package so that we and other interested plugins can consume it. 

We should consume this code and implement a solution similar to what's described in this [comment](https://github.com/grafana/synthetic-monitoring-app/pull/1179#issuecomment-3074986086):

> 🔔 Alert Routing Summary
> 
> This alert includes the following labels:
> 
> team_name: frontend
> label_name: label_value
> ✅ This alert will be routed to: Slack – #some-channel
> 
> 📬 Notification Contact Points:
> 
> Slack: #some-channel
> Email: [alerts@example.com](mailto:alerts@example.com)
> 🔧 [Edit notification policies]
> 📚 [Learn how alert routing works]
> 
> Or in the case where theres no matching policy:
> 
> 🔔 Alert Routing Summary
> 
> This alert includes the following labels:
> 
> team_name: frontend
> label_name: label_value
> ⚠️ This alert does not match any notification policy.
> You may not receive notifications unless a matching policy is created. click here to create a policy
> 
> ➡️ [Create a policy with these labels]
> 📚 [Learn how alert routing works]

PR #1282, opened 2025-10-07
Feat: Per check alerts/notification policies

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1258

Consumes https://github.com/grafana/grafana/pull/108982 in order to display a notification policy preview for each of our alert instances:

image
image
image

It shows the journey each alert instance will have for its notification based on the notification policies configured in the Alerting section:

image

### DIFF
--- a/src/components/CheckForm/AlertsPerCheck/AlertItem.tsx
+++ b/src/components/CheckForm/AlertsPerCheck/AlertItem.tsx
@@ -4,9 +4,10 @@ import { GrafanaTheme2, urlUtil } from '@grafana/data';
 import { Button, TextLink, useStyles2 } from '@grafana/ui';
 import { css } from '@emotion/css';
 
-import { CheckAlertType, CheckFormValues } from 'types';
+import { CheckAlertType, CheckFormValues, FeatureName } from 'types';
 import { useMetricsDS } from 'hooks/useMetricsDS';
 import { NotOkStatusInfo } from 'components/AlertStatus/NotOkStatusInfo';
+import { FeatureFlag } from 'components/FeatureFlag';
 
 import { AlertRoutingPreview } from './AlertRoutingPreview';
 import { PredefinedAlertInterface } from './AlertsPerCheck.constants';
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -112,6 +112,7 @@
   },
   "dependencies": {
     "@emotion/css": "11.10.6",
+    "@grafana/alerting": "canary",
```
ckbedwell: Before we merge this we probably should work out what this needs to be pinned to. As much as TIL you could do this in this PR, I can see it causing longer term problems if we left it.

src/components/CheckForm/AlertsPerCheck/AlertItem.tsx
```diff
@@ -29,6 +30,7 @@ export const AlertItem = ({
   onSelectionChange: (type: CheckAlertType) => void;
 }) => {
   const styles = useStyles2(getAlertItemStyles);
+  const [showRouting, setShowRouting] = useState(false);
```
ckbedwell, line 34: Not for this PR because I think the best way to fix it will be when 1280 goes in (cc @w1kman 😉) but I wanted the wizard form to remember my routing visible / hidden state when I was jumping between the labels step and alerting step and making sure I had added the labels correctly and they would be matched.

src/components/CheckForm/AlertsPerCheck/RouteTreeDisplay.tsx
```diff
@@ -0,0 +1,177 @@
...
+  level: number;
+  routingStops?: boolean;
+  isEffective?: boolean;
+}> = ({ route, level, routingStops, isEffective = false }) => {
+  const styles = useStyles2(getStyles);
+  const hasMatchers = route.matchers && route.matchers.length > 0;
+  const isDefaultPolicy = level === 0 && !hasMatchers;
+  const policyInfo = getPolicyIdentifier(route, isDefaultPolicy);
+
+  const badgeClass = policyInfo.type === 'matchesAll' ? styles.matchesAllBadge : styles.matchersBadge;
```
ckbedwell: nit: you can use the the `cx` method from `@emotion/css` for the conditional logic. `styles.matchersBadge` and `styles.matchesAllBadge` have a fair few shared styles so could have a shared style and then just have the additional styles as appropriate:

```
className={cx(styles.badge, {
 [styles.matchers]: policyInfo.type === 'matchers',
 [styles.matchesAll]: policyInfo.type === 'matchers',
})}
```

src/components/CheckForm/AlertsPerCheck/RouteTreeDisplay.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  return (
+    <div key={route.id || `route-${level}`} className={styles.routeNode} style={{ marginLeft: level * 16 }}>
```
ckbedwell: nit: Does this need the OR for the key? Shouldn't they all have a unique ID?

ckbedwell (changes requested):
My CR are really tiny.

This feature is great and makes a real difference to the alerting experience! 👏

package.json
VikaCep: I've pinned it to the specific `12.3.0-18685446948` version that corresponds to `canary`. The `@grafana/alerting` package will follow the core grafana versioning and release schedules for now as @gillesdemey told me. That means that we should be able to point to the next minor release when that is cut (12.3.0). We should also change the `@grafana/alerting/unstable` imports to `@grafana/alerting` when they're available.

Issue #1252, opened 2025-09-01
Make monaco editor use the correct k6 types


VikaCep:
Currently the k6 version for the script editor is hardcoded in `package.json`.
This version is used to provide autocomplete and provide errors as users write their scripts. Since the k6 channel can be selected with Version Management, this typings should be dynamically loaded so that users can write scripts with the editor assuming the correct version.

PR #1289, opened 2025-10-16
Feat: make monaco load types dynamically

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1252

# Dynamic k6 Type Loading for Monaco Editor

## Overview

This PR implements dynamic loading of k6 TypeScript definitions in Monaco editor based on user-selected k6 channels. Users now get accurate IntelliSense and type checking that matches the actual k6 version used at execution time.

Previously, we bundled a single k6 version (currently `0.57.1`) regardless of the channel selection. This means with the version selection new features, users would load the editor with types that won't often match the actual execution environment, leading to confusion about available features and missing autocomplete for channel-specific APIs.

## How It Works

When a user selects a k6 channel in the dropdown, the CodeEditor receives a `k6Channel` prop and triggers the dynamic type loading process:

`Channel Selection →  CDN Fetch (with React Query Cache) using the channel version → Monaco Update → Accurate IntelliSense`

~~The system fetches the current version using the endpoint `sm/channel/k6/${channelId}/current` (which is to be implemented), then loads the corresponding k6 type definitions from `unpkg.com/@types/k6@{version}`. All k6 modules are fetched including http, metrics, and experimental features.
~~ --> no longer valid, as there is no support of `sm/channel/k6/${channelId}/current`.
With the new approach, we rely on `unpkg` to [resolve](https://unpkg.com/) to the latest version, as it accepts semver ranges and npm tags:

> You can also use any valid [semver](https://docs.npmjs.com/about-semantic-versioning) range or [npm tag](https://docs.npmjs.com/adding-dist-tags-to-packages):
> [unpkg.com/preact@latest/dist/preact.min.js](https://unpkg.com/preact@latest/dist/preact.min.js)
> [unpkg.com/react@^18/umd/react.production.min.js](https://unpkg.com/react@%5E18/umd/react.production.min.js)

When the type fetching fails, we use the default bundled types as a fallback.

https://github.com/user-attachments/assets/acf73091-d29b-435a-9632-e22297d7095b

https://github.com/user-attachments/assets/ae282a1f-1d7f-4053-b2cc-c264c8a3f29a

### DIFF
--- a/src/components/CodeEditor/k6TypesLoader/k6TypesCdnLoader.ts
+++ b/src/components/CodeEditor/k6TypesLoader/k6TypesCdnLoader.ts
@@ -41,8 +41,6 @@ export async function fetchK6TypesFromCDN(channelId: string): Promise<Record<str
       
       const content = await response.text();
       types[module.name] = content;
-
-      console.log(url, content);
     } catch (error) {
       failedModules.push(module.name);
     }
--- a/src/page/__testHelpers__/channel.ts
+++ b/src/page/__testHelpers__/channel.ts
@@ -1,3 +1,9 @@
+// Mock k6 types CDN loader to avoid real external fetch calls to unpkg.com
+// Channel tests enable VersionManagement which renders CodeEditor that fetches k6 types
+jest.mock('components/CodeEditor/k6TypesLoader/k6TypesCdnLoader', () => ({
+  fetchK6TypesFromCDN: jest.fn().mockResolvedValue({}),
+}));
+
 import { screen, waitFor } from '@testing-library/react';
 import { apiRoute } from 'test/handlers';
 import { server } from 'test/server';


src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -62,8 +86,26 @@ export const CodeEditor = forwardRef(function CodeEditor(
 ) {
   const [editorRef, setEditorRef] = useState<null | monacoType.editor.IStandaloneCodeEditor>(null);
   const [constrainedInstance, setConstrainedInstance] = useState<null | ConstrainedEditorInstance>(null);
+
+  const isJs = language === 'javascript';
   const [prevValue, setPrevValue] = useState(value);
 
+  const { types: dynamicK6Types, loading: k6TypesLoading, error: k6TypesError } = useK6TypesForChannel(k6Channel, isJs);
+
+  console.log('hola',k6Channel, isJs, k6TypesLoading, k6TypesError);
```
ckbedwell: Hola! ✂️

src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -161,6 +204,25 @@ export const CodeEditor = forwardRef(function CodeEditor(
...
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'center',
+            color: '#888',
+          }}
+        >
+          Loading k6 types for {k6Channel}...
+        </div>
+      </div>
+    );
```
ckbedwell: I think this should render as an overlay on the editor rather than replacing it being rendered, it feels quite jarring that it snaps in and out.

Do the types re-register correctly if that is the case? Often adding the `key={k6ChannelValueWhichHasNowSuccessefullyLoadedOrSomething}` prop forces a subtle re-render if it needs a re render to register them?

VikaCep: Good idea! Implemented the overlay as suggested:

https://github.com/user-attachments/assets/38b9417f-4f3b-4604-b1f9-385e9a90fcdf

ckbedwell (dismissed):
LGTM! ⭐

Issue #1261, opened 2025-09-05
Show UI hint for checks disabled due to free tier overage


The-9880:
The `disableReason` field is now available on the fetched `Check` object to denote checks which were disabled due to overage in the free tier 🎉 

We should show a UI hint so these are easier to track.

PR #1290, opened 2025-10-20
feat: add indicator of disabled check

VikaCep:
# Add UI hint for checks disabled due to free tier overage

Closes issue #1261

## Summary

Implements a visual indicator for checks that have been disabled due to free tier overage. When the `disableReason` field is populated on a check object, users will now see a clear orange "Disabled" badge with a tooltip explaining the reason.

image

### DIFF
--- a/src/page/CheckList/components/DisableReasonHint.test.tsx
+++ b/src/page/CheckList/components/DisableReasonHint.test.tsx
@@ -1,16 +1,43 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
 import { DisableReasonHint } from './DisableReasonHint';
 
 describe('DisableReasonHint', () => {
-  it('renders badge when disableReason is provided', () => {
-    render(<DisableReasonHint disableReason="This check was disabled due to free tier overage" />);
-    expect(screen.getByText('Disabled')).toBeInTheDocument();
+  it('renders badge and shows user-friendly message for FREE_LIMIT_EXCEEDED', async () => {
+    const user = userEvent.setup();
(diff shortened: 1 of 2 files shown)


ckbedwell (changes requested):
@The-9880 -- I'm cross-referencing the [OAS Check schema](https://synthetic-monitoring-api-dev.grafana-dev.net/api/v1/swagger#/schemas/Check) (also 👏)  and I noticed `disableReason` is always present in the schema which then showed me that reasonless checks have an empty string. My preference would be `null`, would that be a big deal to change?

What are the current strings we have that this field can be?

I'd also like for us to agree that the reason field is some sort of enum we can map to.  Reasons for this are:

1. i18n. The less English language strings we have coming from the API the easier it is for us to fully translate the app when we get round to it (which will be on the roadmap in FY27)
2. Depending on the reason we can change the UI and potential messaging. In particular if we wanted to try and upsell if someone's checks were disabled because they hit the free tier limit.

@VikaCep -- works great. Maybe just add a simple test that if `disableReason` is present it gets rendered in the UI and it it is an empty string it doesn't?

ckbedwell:
> The current only value we are using for this is FREE_LIMIT_EXCEEDED, which is a string but formatted like an enum identifier. Does this work for the app?

Beautiful. Exactly what we want ❤️

@VikaCep can you make the changes so that we show a user-friendly message using something like a key/value map?

src/page/CheckList/components/DisableReasonHint.test.tsx
ckbedwell, line 1: Can we also add a test if the value is `null` in response to when they update it in the API?

PR #1305, opened 2025-10-24
feat: allow to duplicate a check

### LINKED ISSUE #1034: Ability to duplicate an existing check

Many checks have nearly identical configuration, and differ only by a single value.  For example I might have 10 HTTP checks that use the same set of probes, HTTP configurations, and labels, but differ only by the target (instance value). 

Having an option to "duplicate" or "copy" an existing check would make it easier to quickly create a set of checks with uniform configuration. 

**Why is this needed**:

In the UI today, if you want create two checks with identical settings, you must go through the entire check creation flow twice and set all the same configurations. 

Internal reference to this enhancement request:  https://grafana.zendesk.com/agent/tickets/166402

VikaCep:
## Add check duplication functionality

Closes https://github.com/grafana/synthetic-monitoring-app/issues/1034

This PR adds a "duplicate check" feature that allows users to quickly create new checks based on existing ones with identical configuration.

- Added a duplicate button (copy icon) to check action buttons in the check list
- Created `useDuplicateCheck` hook that navigates to the create check page with the original check ID
- Modified `NewCheck` component to detect duplicate requests via URL params and pre-populate the form with the original check data, removing the ID and appending "(Copy)" to the job name

https://github.com/user-attachments/assets/e12fc953-339d-4520-8964-cd6f620b21e6


w1kman (commented):
With the old check editor:
When duplicating and changing checkType, all "duplicated" data is lost. The question is whether or not we spend time on fixing it?

ckbedwell (commented):
I'll review this in more detail tomorrow but my immediate thought is this is a lot of overhead for something that is supposedly a 'papercut' issue. I found a few oddities, too but I'll document them better when I've got full repro steps.

I'm wondering if a dedicated route would work better now I'm playing with it. As a user I find it confusing that I've hit 'Duplicate' and see the breadcrumbs for `Choose a check type > API Endpoint`. I'm wondering if I'd rather see: `Checks > Duplicate {jobName}` or even `Checks > {jobName} > Duplicate`?

cc @vesnadean @markjmeier -- any thoughts on the experience for duplicating a check? Should it take you to the 'new check' route or a dedicated duplicate route?

We are also discussing if you should be able to change the check type in Slack.

src/page/CheckList/components/CheckItemActionButtons.tsx
```diff
@@ -69,6 +73,17 @@ export const CheckItemActionButtons = ({ check, viewDashboardAsIcon }: CheckItem
...
+        data-testid="duplicate-check-button"
+        tooltip="Duplicate check"
+        name="copy"
+        onClick={() => {
+          const checkType = getCheckType(check.settings);
+          trackDuplicateCheckButtonClicked({ checkType });
+          duplicateCheck(check);
+        }}
+        disabled={!canWriteChecks}
+      />
```
ckbedwell: Can we change this to a LinkButton so it is more accessible and can be opened in another tab?

src/types.ts
```diff
@@ -334,6 +334,7 @@ export type CheckAlertFormRecord = Partial<Record<CheckAlertType, CheckAlertForm
 export type CheckFormValuesBase = Omit<Check, 'settings' | 'basicMetricsOnly' | 'alerts'> & {
   publishAdvancedMetrics: boolean;
   alerts?: CheckAlertFormRecord;
+  duplicate_check?: boolean; // Only used to make the check form "dirty" when duplicating a check
```
ckbedwell: nit: is there anyway we can move this to some separate meta data state? I understand we want to track it but it is meta data and has no bearing on the form values we are sending to the API. The more meta data we add to `CheckFormValues` the harder it becomes to reason with what a check looks like and what we should be sending to the API.

ckbedwell (changes requested):
When I duplicate a scripted or browser check I get an error but it does save fine:


https://github.com/user-attachments/assets/908d760e-bb4a-4f27-b9d1-2c298381b248

src/types.ts
VikaCep: I managed to remove it from `CheckFormValues` by deferring when the `(Copy)` suffix is added to the job name.

Before we were adding`(Copy)` to the job name early and that was the default data for the new check. As this became the default value, the form wasn't dirty and we needed  the `duplicate_check` field to artificially make it dirty and allow saving.

Now, the original job name is not initially changed when creating the duplicate.  `ChecksterContext` adds `(Copy)` programmatically after form initialization, and this makes it dirty without needing to store that value in the model. 

What do you think of this approach?

src/page/CheckList/components/CheckItemActionButtons.tsx
VikaCep: Done!

https://github.com/user-attachments/assets/31761706-6f29-4a83-bb5f-2b2a18db741b

PR #1316, opened 2025-10-27
feat: add bespoke documentation and tracking per check type

markjmeier:
## Problem

1. The documentation panel in the check editor v2 is generic and not tailored to any particular check types.
2. If you are new to writing scripted or browser checks, it is difficult to get started and know where to begin

## Solution

This PR addresses the first problem by expanding the documentation panel feature so it looks up the check type to decide what to render.

**API Endpoint and Multi Step checks documentation panel**
image

**Scripted check documentation panel**

image

**Browser check documentation panel**

image

The second problem is addressed by adding a small text button above the code editor which links to opening the docs panel and includes promoting k6 Studio as a means to authoring your scripted checks.

https://github.com/user-attachments/assets/e1fa4c15-6126-4a0a-a85b-090e51cbf246

I've taken this opportunity to refactor the `DocsLink` component and add a tracking function so we can send telemetry and gain insights into how users are interacting with our linked documentation.

### DIFF
--- a/src/components/DocsLink/DocsLink.constants.ts
+++ b/src/components/DocsLink/DocsLink.constants.ts
@@ -1,2 +1,2 @@
 export const TRACKING_LINK_SRC = `grafana-cloud`;
-export const TRACKING_LINK_CNT = `synthetic-monitoring`;
+export const TRACKING_LINK_CNT = `grafana-synthetic-monitoring-app`;


src/components/Checkster/components/ui/SecondaryLayoutSection.tsx
```diff
@@ -40,5 +44,11 @@ function getStyles(theme: GrafanaTheme2) {
         border-right: 1px solid ${theme.colors.border.medium};
       }
     `,
+    sectionContent: css`
+      &:focus {
+        outline: 2px solid ${theme.colors.primary.border};
+        outline-offset: -2px;
+      }
+    `,
```
w1kman: We usually don't put an outline on every container div. Why do we do it here?
I can't say that I think it looks particularly good.

w1kman: Image

src/components/Checkster/feature/docs/AboutBrowserChecks.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Stack, Text } from '@grafana/ui';
+
+export function AboutBrowserChecks() {
+  return (
+    <Stack direction="column" gap={2}>
+      <Text element="p">
```
w1kman: Feels like we're missing a heading before the "about text", aren't we?

Image

src/components/Checkster/feature/docs/Aboutk6Studio.tsx
```diff
@@ -0,0 +1,44 @@
...
+      </Text>
+      <Text element="p">
+        k6 Studio is a free, open source desktop application designed to help you create k6 test scripts using a visual
+        interface. Download it for free and get started with your first script in minutes.
+      </Text>
+      <div>
+        <LinkButton
+          href="https://grafana.com/docs/k6-studio/set-up/install/"
+          target="_blank"
+          onClick={() => onDocsLinkClick('https://grafana.com/docs/k6-studio/set-up/install/', source)}
```
w1kman: Since it is an external link, adding the external link icon seems fitting.

```suggestion
          target="_blank"
          icon="external-link-alt"
          onClick={() => onDocsLinkClick('https://grafana.com/docs/k6-studio/set-up/install/', source)}
```

Image

src/components/Checkster/feature/docs/Aboutk6Studio.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { GrafanaTheme2 } from '@grafana/data';
+import { LinkButton, Stack, Text, useStyles2, useTheme2 } from '@grafana/ui';
+import { css } from '@emotion/css';
+
+import { onDocsLinkClick } from 'components/DocsLink/DocsLink.utils';
+
+import { k6StudioLogoDarkTheme, k6StudioLogoLightTheme } from 'img';
+
+export function Aboutk6Stuido({ source }: { source: string }) {
```
w1kman:
```suggestion
export function Aboutk6Studio({ source }: { source: string }) {
```

src/components/Checkster/feature/docs/AboutScriptedChecks.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Stack, Text } from '@grafana/ui';
+
+export function AboutScriptedChecks() {
+  return (
+    <Stack direction="column" gap={2}>
+      <Text element="p">
```
w1kman: Since this is shown before a section with a `h3`, it seems fitting to have a `h3` here as well.

src/components/Checkster/feature/docs/constants.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const DOC_LINK_K6_SCRIPTED_CHECKS = {
+  title: 'Learn more about k6 scripted checks',
+  href: 'https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/create-checks/checks/k6/',
+};
+
+export const DOC_LINK_K6_SM = {
+  title: 'Check types and what they do',
+  href: 'https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/create-checks/checks/',
+};
```
w1kman: Unused ✂️ 

```suggestion
```

src/components/Checkster/feature/docs/DocsPanel.tsx
w1kman, line 1: I don't like the fact that the "feature" has been split into several features pretending to be the same feature.

src/components/Checkster/feature/docs/AboutSMChecks.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { Stack, Text } from '@grafana/ui';
+
+export function AboutSMChecks() {
```
w1kman:
```suggestion
export function AboutApiEndpointChecks() {
```

w1kman: All checks are SM-checks, also, we're referring to it as "...APIEndpoint..." for the actual parent component.

src/components/Checkster/feature/docs/index.ts
w1kman, line 1: IMO, this should export ONE component.
The inner workings of the feature (docs in this case), should not have to be spread out and handled by other components. Especially not for a trivial thing as showing context aware content.

src/components/Checkster/feature/config.ts
```diff
@@ -1,12 +1,24 @@
...
 import { SECRETS_CHECK_COMPATIBILITY, SecretsPanel } from './secrets';
 
 export const FEATURE_TABS: FeatureTabConfig[] = [
   ['Test', AdhocCheckPanel, ADHOC_CHECK_COMPATABILITY],
   ['Secrets', SecretsPanel, SECRETS_CHECK_COMPATIBILITY, FeatureName.SecretsManagement],
-  ['Docs', DocsPanel, DOCS_CHECK_COMPATABILITY],
+  ['Docs', DocsPanelAPIEndpoint, API_ENDPOINT_DOCS_CHECK_COMPATABILITY],
+  ['Docs', DocsPanelBrowserCheck, BROWSER_CHECK_DOCS_CHECK_COMPATABILITY],
+  ['Docs', DocsPanelScriptedCheck, SCRIPTED_DOCS_CHECK_COMPATABILITY],
+  ['Docs', DocsPanelMultiStep, MULTI_STEP_DOCS_CHECK_COMPATABILITY],
```
w1kman: This was not the way it was intended to be used.

The docs feature is supported by ALL check types. The fact that it shows different content for different check types should not be a consideration from outside the exported panel component.

src/components/DocsLink/DocsLink.utils.ts
```diff
@@ -0,0 +1,30 @@
...
+function parseUrl(url: string) {
+  try {
+    const parsedUrl = new URL(url);
+    return {
+      hostname: parsedUrl.hostname,
+      pathname: parsedUrl.pathname,
+      search: parsedUrl.search,
+    };
+  } catch (error) {
+    return null;
```
w1kman, line 34: Perhaps adding some debugging information for non production environments would be in its place?

src/components/DocsLink/DocsLink.tsx
```diff
@@ -1,46 +1,24 @@
...
+import { onDocsLinkClick } from 'components/DocsLink/DocsLink.utils';
 
 type DocsLinkProps = {
-  article: keyof typeof docs;
   children: ReactNode;
-  iconPosition?: 'prefix' | 'suffix';
-  className?: string;
+  href: string;
+  inline?: boolean;
+  source: string;
```
w1kman, line 11: I think that the "source" prop warrants a description to what it's expected to be

src/features/tracking/linkEvents.ts
```diff
@@ -0,0 +1,19 @@
...
+
+interface LinkEvent extends TrackingEventProps {
+  /** The href of the clicked link */
+  href: string;
+  /** The hostname of the clicked link */
+  hostname: string;
+  /** The path of the clicked link */
+  path: string;
+  /** The search of the clicked link */
+  search: string;
```
w1kman: Nit/hint: even though the type won't change, we'd be able to piggyback on the descriptions from the URL typings.

```suggestion
  /** The href of the clicked link */
  href: URL['href'];
  /** The hostname of the clicked link */
  hostname: URL['hostname'];
  /** The path of the clicked link */
  path: URL['pathname'];
  /** The search of the clicked link */
  search: URL['search'];
```

src/img/k6-studio-logo-dark-theme.svg
w1kman, line 1: Same effect (and better) can be achieved with actually adding "color" props (or just use `currentColor`, which would also work for non "dark|light" themes.

src/img/k6-studio-logo-light-theme.svg
w1kman, line 1: Remove this one, use a generic one that works with any theme.

src/components/Checkster/components/form/layouts/ScriptedCheckContent.tsx
```diff
@@ -31,6 +33,7 @@ export function ScriptedCheckContent({
   const theme = useTheme2();
   const hasExamples = examples && examples?.length > 0;
   const styles = useStyles2(getStyles);
+  const source = scriptField === 'settings.scripted.script' ? 'scripted_check' : 'browser_check';
```
w1kman: This assumes that there will be no other check type that uses scripted sections.
The better way would be to use the `checkType` (also, there is no need to pass the source prop, since `HelpButton` is not exported, it might as well just use the `ChecksterContext` and create the source itself.

src/components/Checkster/components/form/layouts/ScriptedCheckContent.tsx
```diff
@@ -54,6 +57,27 @@ export function ScriptedCheckContent({
   );
 }
 
+const HelpButton = ({ source }: { source: string }) => {
```
w1kman:
```suggestion
function HelpButton() {
const { checkType } = useChecksterContext();
const source = `${checkType}_check`;
```

src/components/Checkster/components/ui/SecondaryLayoutSection.tsx
w1kman: After looking at all the code, I figure that what we need is something that highlights that we opened up a tab for the customer, not a constant outline.

Perhaps it can be done in a more elegant way, that doesn't mean that the UI constantly has an outline?

w1kman (changes requested):
The general idea is super and it helps in polishing the check authoring flow.

The order of the scripted docs seems a bit off to me. Some general text (no heading), then a (actually two) SVG inside a heading, promoting k6 studio, then some actual docs links.

I think we should add "secrets" to all doc lists, since it highlights a pretty fine feature. Can secrets be used with protocol checks, no. But at least the user would get informed that they'd need to change check type to be able to use secrets.

I'm fine with whatever we decide to do with secrets, just want to pitch that it can still be handy for our users.

**Testing**
With the new fancy SVGs, the test runner consumes waaaay more RAM. I'd consider making a test aware HoC for the SVG (which could also deal with the theming issue).

For a speedy re-review, please poke me on Slack.

src/img/k6-studio-logo-dark-theme.svg
ckbedwell: I'm going to leave this one and circle back to it. We need a better strategy for handling SVGs.

src/components/Checkster/components/ui/SecondaryLayoutSection.tsx
ckbedwell: https://github.com/grafana/synthetic-monitoring-app/pull/1444

Separate PR so not to block this one and so we can evaluate methods in greater depth.

PR #1342, opened 2025-10-29
fix: put the container-type on the actual container

ckbedwell:
## Problem

The original fix for the Feedback widget was applied incorrectly. 

## Solution

Tear out the unnecessary containers and add it to the Time Point Explorer **container**.


w1kman (approved):
LGTM 👌🏻 
- Timepoint explorer ✔️ 
- Check editor ✔️

PR #1346, opened 2025-10-29
feat: make check editor features be configured with FeatureName

w1kman:
image

## The problem

Even if Secrets management isn't enabled, the secrets tab will be rendered.

## The solution

Make it possible to add FeatureName (that needs to be enabled) to tab config.


ckbedwell (approved):
LGTM

PR #1411, opened 2025-11-04
Fix/pathfinder guide1 testids

ckbedwell:
## Problem

In our effort to support Pathfinder and creating durable and long lasting interactive guides, we want to instrument our app with `data-testid` attributes.

Currently our `data-testid` naming convention is fragmented and inconsistent within the repo and with the rest of Grafana.

## Solution

This PR adds `data-testid` attributes to the essential parts of activating Synthetic Monitoring and setting up a check with alerts. As Pathfinder will be multilingual and we will be looking at supporting i18n in the near-term, we want to ensure our Pathfinder guides are long-lasting by encouraging the use of `data-testid`s from the get-go.

This PR also looks at starting to deprecate the old dataTestId enum and update it with an object constant that follows the same naming convention as Grafana.

### DIFF
--- a/src/components/Checkster/components/form/generic/GenericInputSelectField.tsx
+++ b/src/components/Checkster/components/form/generic/GenericInputSelectField.tsx
@@ -40,7 +40,7 @@ export function GenericInputSelectField({
   const inputValue = watch(field) as any;
 
   const handleOnChange = ({ value }: SelectableValue) => {
-    setValue(field, value);
+    setValue(field, value, { shouldDirty: true });
   };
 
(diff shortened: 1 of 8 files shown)


src/components/Checkster/components/form/generic/GenericInputSelectField.tsx
```diff
@@ -27,7 +27,6 @@ export function GenericInputSelectField({
   description,
   options = [],
   className,
-  width = 20, // 0 means undefined
```
ckbedwell, line 30: This seemed unused and was confusing me? 😅

Issue #1013, opened 2024-12-09
Poll to update probe statuses


ckbedwell:
## Problem

When setting up a private probe the offline response gets cached until you manually refresh the page. The experience would be improved significantly if the status updated without doing a page refresh.

## Acceptance criteria

1. We should poll for probe status at a set interval. Please explore what would be the appropriate amount of time.
2. The status should automatically update in all the places we display probe status across the app (checks creation, checks dashboard, probe card)
3. Please ensure the appropriate integration tests have been added

https://github.com/user-attachments/assets/5282e8e0-83f8-4eee-894e-344e049656b4

PR #1415, opened 2025-11-05
feat: probe polling across application

ckbedwell:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/1013

## Problem

In many places across the app where we show probe online/offline status it doesn't update automatically. We fetch it once then store it in the cache so users only get an updated status if they refresh their browser or navigate away from SM then come back.

## Solution

This PR adds a `refetchInterval` of ten seconds to the underlying `useProbes`hook, ensuring that the probe status is always up to date. I chose ten seconds because it feels like the right balance between getting feedback when you may be turning on and off probes versus being too noisy.

I added a sync icon next to the probe's status on the probe editing screen as I figured you may want to explicitly test if your probe is online when setting it up for the first time.

### DIFF
--- a/.github/release-please/release-please-config.json
+++ b/.github/release-please/release-please-config.json
@@ -10,11 +10,11 @@
         { "type": "docs", "section": "Documentation" },
         { "type": "chore", "section": "Miscellaneous Chores" },
         { "type": "refactor", "section": "Code Refactoring" },
-        { "type": "release", "section": "Release"}
+        { "type": "release", "section": "Release" }
       ]
     }
   },
   "draft-pull-request": true,
-  "pull-request-title-pattern": "chore: release ${version}",
+  "pull-request-title-pattern": "release: Grafana Synthetic Monitoring App v${version}",
   "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json"
 }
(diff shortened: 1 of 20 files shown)


src/page/EditProbe/EditProbe.test.tsx
```diff
@@ -79,6 +80,78 @@ describe(`Private probes`, () => {
...
+              },
+            ],
+          };
+        },
+      })
+    );
+
+    jest.advanceTimersByTime(PROBE_REFETCH_INTERVAL);
+
+    await screen.findByText(/Online/);
```
VikaCep: personally I like it better to be more explicit and use expect assertions, like `expect(await screen.findByText(/Online/)).toBeInTheDocument();`

src/page/EditProbe/EditProbe.test.tsx
```diff
@@ -79,6 +80,78 @@ describe(`Private probes`, () => {
...
+          return {
+            json: [OFFLINE_PROBE],
+          };
+        },
+      })
+    );
+
+    renderEditProbe(OFFLINE_PROBE);
+
+    // Initial render - should call once
```
VikaCep: instead of the comment, don't you want to add an assertion making sure it's just called 1 times?

VikaCep (dismissed):
LGTM, small non-blocking comments on some tests.

PR #1416, opened 2025-11-05
feat: add enable/disable check button with live updates

markjmeier:

On the checks page `/a/grafana-synthetic-monitoring-app/checks` I wanted to solve to problems:
- make it eaier to enable/disable checks, without going through the bulk edit
- improve the zero state for disabled check

This PR adds:

- Add play/pause button to check list items for quick enable/disable
- Show pause icon when check is enabled (to disable)
- Show play icon when check is disabled (to enable)
- Add query invalidation to useUpdateCheck for live page updates after enable/disable

n/a

Fixes #

Created with Cursor/Graft

image
image


VikaCep (changes requested):
> Add N/A gauges for uptime, reachability, and latency when check is paused

This is currently not showing:
https://github.com/user-attachments/assets/2e65520e-7b7b-4f8e-84f2-13f539102e37

I also think we should change the icon to be a loading indicator while the query is loading as otherwise it looks like it freezes for a few secs.

Issue #1212, opened 2025-07-30
Start reporting `X-Client-ID`/`X-Client-Version`


The-9880:
The API now has support for clients to include the following headers:
* `X-Client-ID`
* `X-Client-Version`

We should start including these in requests to the HTTP API so that we can better understand which requests are coming from the app.

PR #1419, opened 2025-11-06
feat: add X-Client headers to datasource methods

ckbedwell:
Closes: https://github.com/grafana/synthetic-monitoring-app/issues/1212

## Problem

Customers may be managing their resources outside of the Synthetic Monitoring application but we have no way to identify what these are.

## Solution

The API now accepts two headers: `X-Client-ID` and `X-Client-Version` which indicate where the interaction with the API came from. Our Terraform provider has been sending these headers for a while now and this PR adds the same headers with the Synthetic Monitoring app's values.

The long-term goal here is to add a feature which identifies how resources are managed so customers can easily find the source of truth for managing them, as well as blocking interactions -- such as checks which are managed by Terraform can't be updated in the app.

## Additional context

I found this ticket was a good excuse to add tests to our datasource module as I can easily see this regressing as we work on the datasource methods further.

### DIFF
--- a/src/datasource/DataSource.ts
+++ b/src/datasource/DataSource.ts
@@ -35,7 +35,6 @@ import {
 } from './responses.types';
 import { QueryType, SMOptions, SMQuery } from './types';
 import { findLinkedDatasource, getRandomProbes, queryLogs } from 'utils';
-import { SM_CLIENT_HEADERS } from 'datasource/constants';
 import { ExtendedBulkUpdateCheckResult } from 'data/useChecks';
 import { SecretsResponse } from 'data/useSecrets';
 import { SecretFormValues } from 'page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementTab.utils';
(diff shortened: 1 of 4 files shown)


ckbedwell:
> Are you intentionally not adding the headers to `GET` endpoints?
> 
> e.g. `listProbes`, `listChecks`, `getCheck`? Otherwise, LGTM

Added them now! It made the code much cleaner as I just put it on our `fetchAPI` wrapper method.

src/datasource/DataSource.ts
```diff
@@ -301,12 +302,12 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
...
   //--------------------------------------------------------------------------------
 
-  async listChecks(includeAlerts = false) {
-    return this.fetchAPI<ListCheckResult>(`${this.instanceSettings.url}/sm/check/list?includeAlerts=${includeAlerts}`);
+  async getCheckInfo() {
+    return this.fetchAPI<CheckInfoResult>(`${this.instanceSettings.url}/sm/checks/info`);
   }
 
-  async getCheck(checkId: number) {
-    return this.fetchAPI<Check>(`${this.instanceSettings.url}/sm/check/${checkId}`);
```
The-9880, line 309: I guess removing this is fine because it seems unused, I was kind of confused by it. Do we always just call `listChecks` then?

I refreshed individual check pages and that seems to be the case.

The-9880 (approved):

src/datasource/DataSource.ts
ckbedwell: Yeah, we might use it at some point and if so I'll then add it back in but right now we just get everything. There's quite a few features where we don't have a dedicated API endpoint to do something so we always revert to using the listChecks method and parse the response ourselves. (e.g. what checks are using what probes, common labels, etc.)

PR #1420, opened 2025-11-06
fix: add check timeout to adhoc check timeout

w1kman:
# adhoc check timeout too short

The new check editor is not including the `check.timeout` when waiting for adhoc check to timeout. Instead it uses a hardcoded value of `30s`.

image

This PR will mimic the old way of timing out adhoc checks (`check.timeout` + `30s`).

[Issue described here](https://grafanalabs.enterprise.slack.com/archives/C06T9S2ST63/p1762431560818229)


ckbedwell (approved):
LGTM

PR #1433, opened 2025-11-10
feat: add auto-labelling via github script

ckbedwell:
## Problem

I want to utilise some of our internal dashboards to monitor team velocity for the next FY. I want to measure what our release rate is for each of our release taxonomies but that is difficult to do with our current set-up.

## Solution

As we have a `release-please` set up which also coincides with our `commitlint` set-up we can add a CI/CD step with a github-script that utilises the naming convention to automatically add labels. These labels can then be easily consumed in our internal dashboards to help measure releases.

### DIFF
--- a/.github/workflows/call_auto-label.yml
+++ b/.github/workflows/call_auto-label.yml
@@ -3,8 +3,7 @@ on:
   workflow_call:
 
 permissions:
-  issues: write # needed to add labels
-  pull-requests: read
+  pull-requests: write # needed to add labels to PRs
 
 jobs:
   label-by-title:
--- a/.github/workflows/on-pr-creation.yml
+++ b/.github/workflows/on-pr-creation.yml
@@ -14,8 +14,7 @@ jobs:
     name: Auto Label
     needs: setup-env-and-cache
     permissions:
-      issues: write
-      pull-requests: read
+      pull-requests: write
     uses: ./.github/workflows/call_auto-label.yml
 
   lint:


.github/release-please/release-please-config.json
```diff
@@ -10,11 +10,11 @@
...
         { "type": "chore", "section": "Miscellaneous Chores" },
         { "type": "refactor", "section": "Code Refactoring" },
-        { "type": "release", "section": "Release"}
+        { "type": "release", "section": "Release" }
       ]
     }
   },
   "draft-pull-request": true,
-  "pull-request-title-pattern": "chore: release ${version}",
+  "pull-request-title-pattern": "release: Grafana Synthetic Monitoring App v${version}",
```
ckbedwell, line 18: Not sure why our releases were chores 😅

.github/workflows/call_auto-label.yml
```diff
@@ -0,0 +1,43 @@
...
+          script: |
+            const title = context.payload.pull_request && context.payload.pull_request.title || '';
+            // Mapping: regex -> label name (aligned with conventional commits & release-please)
+            const mappings = [
+              { regex: /\b(feat|feature|enhancement|improvement)\b/i, label: 'feature' },
+              { regex: /\b(fix|bug|bugfix|regression)\b/i, label: 'fix' },
+              { regex: /\b(doc|docs|documentation)\b/i, label: 'docs' },
+              { regex: /\b(chore|ci|pipeline|workflow|deps|dependencies)\b/i, label: 'chore' },
+              { regex: /\b(refactor|refactoring|restructure)\b/i, label: 'refactor' },
+              { regex: /\b(release)\b/i, label: 'release' }
```
ckbedwell, line 24: Do we think the labels should have a prefix to indicate their connection to the release labelling PRs or are they good as they are?

Do we think we should keep these and clean up all our other labels in the repo instead?

VikaCep: Could you give me an example of that? I think it looks good the way it is now.

ckbedwell: Alright, let's leave them. I was wondering whether the labels should only be used for ci / cd at first but I can't think of a good reason to segment them like that.

PR #1440, opened 2025-11-11
fix: pass in logs datasource

ckbedwell:
Related escalation: https://github.com/grafana/support-escalations/issues/19348

## Problem

The `queryLogsV2` method on the datasource currently uses `this.instance.jsonData.logs` as its datasource reference for the query. This is a somewhat opaque method as the app isn't responsible for how it gets initiated. In the escalation above, the customer's datasource reference isn't fully formed so when looking for the logs that the ad hoc check generated it receives a 400 in the response: 

```
message: "Query does not contain a valid data source identifier"
messageId: "query.invalidDatasourceId"
statusCode: 400
```

The payload for the datasource reference in the query is:

```
grafanaName: "grafanacloud-{customer}-logs"
hostedId: <redacted>
```

When we compare this to any of the customer's log requests on their dashboards (who use our useLogsDS hook), we can see the requests complete successfully:

```
{
    "id":<redacted>,
    "uid": "grafanacloud-logs",
    "type": "loki",
    "name": "grafanacloud-{customer}-logs",
}
```

## Solution

As this is intended to be a quick solution to unblock the customer, I've called the same `useLogsDS` hook and passed the result in as an argument to `queryLogsV2`.

The longer term fix is to migrate the ad hoc checks to using the `queryLoki` method which the check dashboards use.

### DIFF
--- a/src/components/Checkster/feature/adhoc-check/useAdHocLogs.ts
+++ b/src/components/Checkster/feature/adhoc-check/useAdHocLogs.ts
@@ -2,7 +2,6 @@ import { useQuery } from '@tanstack/react-query';
 import { DataFrameJSON } from '@grafana/data';
 
 import { AdHocResponseResults } from './types.adhoc-check';
-import { useLogsDS } from 'hooks/useLogsDS';
 import { useSMDS } from 'hooks/useSMDS';
 
 export interface LokiQueryResults<RefId extends keyof any = 'A'> {
@@ -86,12 +85,11 @@ export function useAdHocLogs(
   poll = true
 ) {
   const dataSource = useSMDS();
-  const logsDS = useLogsDS();
 
   return useQuery<AdHocResponseResults>({
-    queryKey: ['logs', 'ad-hoc', { expr, from, to, logsDS }],
+    queryKey: ['logs', 'ad-hoc', { expr, from, to }],
     queryFn: async () => {
-      return loggify(await dataSource.queryLogsV2(expr!, from, to, `adhoc-logs`, logsDS));
+      return loggify(await dataSource.queryLogsV2(expr!, from, to));
     },
     enabled: !!expr,
     refetchInterval: poll ? 3000 : 0,
--- a/src/datasource/DataSource.ts
+++ b/src/datasource/DataSource.ts
@@ -228,7 +228,7 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
             refId: 'A',
             expr,
             queryType: 'range',
-            datasource: this.instanceSettings.jsonData.logs,
+            datasource: this.getLogsDS(),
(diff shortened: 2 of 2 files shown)


src/datasource/DataSource.ts
w1kman, line 1: seems like `queryLogs` suffers from the same issue + it isnt used. ✂️ `queryLogs` and  `useLogs`?

src/components/Checkster/feature/adhoc-check/useAdHocLogs.ts
```diff
@@ -85,11 +86,12 @@ export function useAdHocLogs(
   poll = true
 ) {
   const dataSource = useSMDS();
+  const logsDS = useLogsDS();
 
   return useQuery<AdHocResponseResults>({
-    queryKey: ['logs', 'ad-hoc', { expr, from, to }],
+    queryKey: ['logs', 'ad-hoc', { expr, from, to, logsDS }],
```
w1kman: can `logsDS` change? Seems like an odd queryKey

src/components/Checkster/feature/adhoc-check/useAdHocLogs.ts
```diff
@@ -85,11 +86,12 @@ export function useAdHocLogs(
...
     queryFn: async () => {
-      return loggify(await dataSource.queryLogsV2(expr!, from, to));
+      return loggify(await dataSource.queryLogsV2(expr!, from, to, `adhoc-logs`, logsDS));
```
w1kman: Instead of passing a datasource to the SM datasource, why not let the datasource handle the logs datasource as well?

src/datasource/DataSource.ts
```diff
@@ -254,7 +255,7 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
             expr,
             queryType: 'range',
             // direction: 'backwards',
-            datasource: this.instanceSettings.jsonData.logs,
+            datasource,
```
w1kman: Feels a bit wonky to get the datasource in a hook that calls the same instance of the actual method we're passing it to 🤷🏻 

```suggestion
 queries: [
          {
            refId, // undefined === 'A'
            expr,
            queryType: 'range',
            // direction: 'backwards',
            datasource: this.getLogsDS(),
            intervalMs: 2000,
            maxDataPoints: 1779,
          },
        ],
```

w1kman (changes requested):
The root to all evil is still left `queryLogs` (and `useLogs`), they should probably be removed as they contain the invalid example code.

I don't see why we'd use `useLogsDS` and pass the value as a prop to the same instance as we're getting the logsDS from?

w1kman (approved):
LGTM! 🥇

PR #1441, opened 2025-11-11
fix: changes to notification policies routing view

VikaCep:
## Change the way we show alerting routing policies

- Makes the contact point section more visible
- Adds a link and information to the alerting Notification Policies view to change the matching logic

image

https://github.com/user-attachments/assets/227878a6-9c0c-4340-94be-811ef1d95b97

### DIFF
--- a/src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
+++ b/src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
@@ -135,9 +135,7 @@ export const AlertRoutingPreview: React.FC<AlertRoutingPreviewProps> = ({ alertT
 
         <div className={styles.routeTreeSection}>
           {routeMatches.length > 0 && routeMatches[0] ? (
-            <>
-              <RouteTreeDisplay routeMatch={routeMatches[0]} />
-            </>
+            <RouteTreeDisplay routeMatch={routeMatches[0]} />
           ) : (
             <div className={styles.contactPointsSection}>
               <Alert severity="info" title="Default notification policy will be used">


src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
```diff
@@ -116,13 +116,28 @@ export const AlertRoutingPreview: React.FC<AlertRoutingPreviewProps> = ({ alertT
...
 
         <div className={styles.routeTreeSection}>
           {routeMatches.length > 0 && routeMatches[0] ? (
-            <RouteTreeDisplay routeMatch={routeMatches[0]} />
+            <>
+              <RouteTreeDisplay routeMatch={routeMatches[0]} />
+            </>
           ) : (
             <div className={styles.contactPointsSection}>
               <Alert severity="info" title="Default notification policy will be used">
```
ckbedwell, line 141: I can't work out how to get this else clause (when there are no route matches) to show in the UI.

I deleted all my notification policies and then I blocked the URL (which gave me a nice graceful error), have they updated the implementation so it will always return the default policy now?

src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
```diff
@@ -116,13 +116,28 @@ export const AlertRoutingPreview: React.FC<AlertRoutingPreviewProps> = ({ alertT
...
+              </TextLink>{' '}
+              in the Alerting section.
+            </Text>
+          </div>
         </div>
 
         <div className={styles.routeTreeSection}>
           {routeMatches.length > 0 && routeMatches[0] ? (
-            <RouteTreeDisplay routeMatch={routeMatches[0]} />
+            <>
```
ckbedwell: Super nit: don't need this fragment.

src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
VikaCep: Right, this was added for the case when the routes prop from the `/apis/notifications.alerting.grafana.app/v0alpha1/namespaces/default/routingtrees` endpoint returned `null` when no notification policies were defined:

image

That was an invalid response (see [here](https://github.com/grafana/synthetic-monitoring-app/pull/1282#issuecomment-3398116058)) - it made the hook throw an error. It has been fixed [here](https://github.com/grafana/grafana/pull/112880) by the alerting team so that it now returns an empty array in that case:

image

I think we can keep it as a defensive safeguard in case it ever returns `null` again.

ckbedwell: Alright, once we are on Grafana `12.4` and update the `@grafana/alerting` dependency let's re-evaluate if this is needed.

I don't see the point in having unreachable code if the API no longer supports that outcome. I worry that we are potentially making an assumption that is incorrect if they go the other way and update their API to support an empty array because in the future (for whatever reason) someone has somehow deleted their default notification policy and they truly don't have one and then we are telling them they do 😅

ckbedwell (approved):

src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
VikaCep: I believe it's not possible to delete the default notification policy, at least from the UI, but I agree let's revisit this when we change the dependency to `12.4` 👍

PR #1445, opened 2025-11-11
chore: remove old check editor

w1kman:
# Remove old check editor
This branch removes old check editor
This branch DOES NOT refactor location of files that are still used from the old editor.

### Next steps

Refactor location of files that are still in-use


src/routing/InitialisedRouter.tsx
```diff
@@ -84,12 +82,7 @@ export const InitialisedRouter = () => {
               )
             }
           />
-          {isFeatureEnabled(FeatureName.CheckEditor) ? (
```
VikaCep, line 87: Since the feature flag was removed we can also remove the comment above:

```// Alpha (requires `synthetic-monitoring-check-editor=true`)```

PR #1461, opened 2025-11-14
Feat: add k6 version selection to check form (Checkster version)

VikaCep:
## Add k6 version selection to check form (Checkster version)

Closes https://github.com/grafana/synthetic-monitoring-app/issues/1197

This is based on https://github.com/grafana/synthetic-monitoring-app/pull/1236 but ported to the new `Checkster` form editor.

Adds the channel selector to browser and scripted checks by consuming the new API provided by the backend, which is currently mocked.

See https://docs.google.com/document/d/1ZV70lIZ-xOH9EiAAp7OW2LGVeGwgHYJELdcvsSiH7UU/edit?tab=t.1cne0jb892id for a detailed specification of the `GET /api/v1/channels/k6` endpoint.

- Displays a new component above the script box that lets users choose the desired k6 version.

image

- The component consumes the GET /api/v1/channels/k6 endpoint

image

- This selection is saved to the scripted/browser check's model under `settings.scripted|browser.channel`

- Some channels can be deprecated. When this happens, it will be filtered out of the list of available channels to choose. When an existing check has a channel assigned that turns deprecated, it will be shown along with a warning message.

image

Note: we were using a `disabeld` prop before that was removed from backend support, so it was not included here.


VikaCep:
> Can we work on a separate feature to solve this problem (@w1kman has ideas) then get this in?

Created a separate [PR](https://github.com/grafana/synthetic-monitoring-app/pull/1466) to address this

PR #1470, opened 2025-11-20
fix: use configured datasource instead of default one

VikaCep:
Ref https://github.com/grafana/support-escalations/issues/19505
Related to https://github.com/grafana/synthetic-monitoring-app/pull/911 

### Problem

The Synthetic Monitoring app hardcoded fallbacks to `uid: 'grafanacloud-logs'` for log queries, preventing users from configuring alternative Loki datasources. This blocked use cases where customers need to use different Loki datasources.

**Affected areas:**
- Runtime log queries always tried `grafanacloud-logs` UID first
- Initialization logic prioritized hardcoded UIDs over provisioned configuration

### Solution

Updated the datasource resolution logic to respect the configured datasource in `jsonData.logs` before falling back to hardcoded defaults.

**Changes:**

1. **`DataSource.ts` - `getLogsDS()` and `getMetricsDS()`**
   - Now attempt to use the configured datasource from `jsonData` first
   - Only fall back to hardcoded Grafana Cloud UIDs if the configured datasource doesn't exist

2. **`useAppInitializer.ts` - `findDatasourceByNameAndUid()`**
   - Updated to try finding datasources by configured UID before falling back to default Cloud UIDs
   - Respects datasource configuration from provisioning metadata

### Testing

To test:
1. Provision SM plugin with custom Loki datasource in `jsonData.logs`
2. Verify runtime queries use the configured datasource
3. Verify fallback behavior when configured datasource doesn't exist

### Notes

**GCOM Provisioning Required:**
If the SM datasource is deleted and re-initialized, it will reset to whatever GCOM provisions (the default datasource).

**Future Enhancement:**
We may evaluate adding a UI dropdown in the plugin configuration to allow users to select from available Loki datasources, removing the dependency on GCOM provisioning for this use case.

image

### DIFF
--- a/src/datasource/DataSource.test.ts
+++ b/src/datasource/DataSource.test.ts
@@ -1,6 +1,7 @@
+import { config } from '@grafana/runtime';
 import { BASIC_CHECK_ALERTS } from 'test/fixtures/checkAlerts';
 import { BASIC_HTTP_CHECK } from 'test/fixtures/checks';
-import { SM_DATASOURCE } from 'test/fixtures/datasources';
+import { LOGS_DATASOURCE, METRICS_DATASOURCE, SM_DATASOURCE } from 'test/fixtures/datasources';
 import { PRIVATE_PROBE } from 'test/fixtures/probes';
 import { apiRoute, ApiRoutes, getServerRequests } from 'test/handlers';
 import { server } from 'test/server';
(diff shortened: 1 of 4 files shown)


VikaCep:
> If we want to support this, we probably need to allow the user to specify the data source configuration, that will have to be stored in backend, and the plugin should consume whatever is set there. Additionally this should be propagated to HG somehow, so resetting the plugin does not revert the configuration.
This probably deserves a better discussion. I would consider creating an issue for it so we can triage it.

I agree, this PR was aimed specifically to address the use case of https://github.com/grafana/support-escalations/issues/19505. They were previously taking advantage of a workaround in order to consume the LBAC datasource and it stopped working, so this solution was a temporary fix. (see https://raintank-corp.slack.com/archives/C0162C1K9E1/p1763384093931469)

I do agree the right and long-term solution is to explicitly allow users to select the desired datasources in the UI and store it, so it doesn't reset when the app is reinstalled or the SM datasource restored, but as you said, that would have to be triaged and treated as a new feature. We can close this PR if we want to take that road.

ckbedwell (approved):
Let's just do it. What's the worst that can happen? 😬

It'll be up to the user to manage this properly and we can look at supporting this more officially in the future.

PR #1487, opened 2025-11-25
feat: pass eventInfo to useCUDChecks in useHandleSubmitCheckster

ckbedwell:
## Problem

The `useHandleSubmitCheckster` hook accepts an `eventInfo` parameter for tracking purposes, but it was not being passed through to the underlying `useCUDChecks` hook. This meant that check creation and update events were not being properly tracked with the check type information, making it difficult to analyze user behavior and check type usage patterns.

## Solution

Updated `useHandleSubmitCheckster` to pass the `eventInfo` parameter through to `useCUDChecks`, ensuring that when `eventInfo` is provided, it is properly forwarded to both `useCreateCheck` and `useUpdateCheck` hooks. This enables proper tracking of check creation and update events with check type information.

The `eventInfo` parameter is converted to the format expected by `FaroEventMeta['info']` (Record<string, string>) before being passed to `useCUDChecks`.

### DIFF
--- a/src/test/db/index.ts
+++ b/src/test/db/index.ts
@@ -18,8 +18,11 @@ import {
 } from 'types';
 
 const baseCheckModel = ({ sequence }: { sequence: number }) => {
-  const timeout = faker.number.int({ min: 30, max: 60 * 1000 });
-  const frequency = faker.number.int({ min: timeout, max: 60 * 1000 });
+  const timeout = faker.number.int({ min: 1000, max: 60 * 1000 });
+  const frequency = faker.number.int({ 
+    min: Math.max(timeout, 10 * 1000), 
+    max: 60 * 60 * 1000
+  });
 
   return {
     id: sequence,


src/hooks/useHandleSubmitCheckster.ts
```diff
@@ -27,7 +27,9 @@ export function useHandleSubmitCheckster(initialCheck?: Check) {
   );
 
   const alertsEnabled = useFeatureFlag(FeatureName.AlertsPerCheck).isEnabled;
-  const { updateCheck, createCheck } = useCUDChecks(); // Omitting eventInfo - TODO: Follow-up on that, use error from `useCUDChecks`
+  const { updateCheck, createCheck } = useCUDChecks({
+    eventInfo: eventInfo ? { checkType: eventInfo.checkType } : undefined,
```
VikaCep: why not just the following?

```suggestion
const { updateCheck, createCheck } = useCUDChecks({ eventInfo });
```

Also, is the checkType the only info that is tracked when submitting?

ckbedwell: Yeah -- that's the only thing we track at the moment. It also triggers our tracking event for submissions --I noticed this when I was updating our Analytics dashboard in ops I realised we stopped tracking how many checks were created in the app when checkster was enabled 😢

package.json
```diff
@@ -21,7 +21,7 @@
     "server": "docker compose up --build",
     "sign": "npx --yes @grafana/sign-plugin@latest",
     "test:changed": "jest --watch --onlyChanged",
-    "test:ci": "jest --passWithNoTests --maxWorkers 4 --bail=20",
```
ckbedwell, line 24: This worked! It finally runs all the test suites so you can see all the failing tests across all 4 workers.

src/data/useChecks.ts
```diff
@@ -69,11 +70,7 @@ export function useCreateCheck({ eventInfo, onError, onSuccess }: MutationProps<
     },
     onSuccess: (data) => {
       onSuccess?.(data);
-
-      if (eventInfo?.checkType) {
-        const checkType = eventInfo.checkType as CheckType;
-        trackCheckCreated({ checkType });
-      }
+      trackCheckCreated({ checkType: getCheckType(data.settings) });
```
ckbedwell, line 73: This is easier to handle as it just checks what was successfully submitted for the check type.

src/test/db/index.ts
```diff
@@ -17,21 +17,26 @@ import {
...
-  alerts: [],
-  basicMetricsOnly: faker.datatype.boolean(),
-  labels: [{ name: faker.animal.petName(), value: faker.color.human() }],
-  probes: [],
-  modified: Math.floor(faker.date.recent().getTime() / 1000),
-  created: Math.floor(faker.date.past().getTime() / 1000),
-});
+const baseCheckModel = ({ sequence }: { sequence: number }) => {
+  const timeout = faker.number.int({ min: 30, max: 60 * 1000 });
+  const frequency = faker.number.int({ min: timeout, max: 60 * 1000 });
```
ckbedwell: Noticed this would often create checks that wouldn't pass validation.

VikaCep: this should improve it

```suggestion
  const timeout = faker.number.int({ min: 1000, max: 60 * 1000 });
  const frequency = faker.number.int({ 
    min: Math.max(timeout, 10 * 1000), 
    max: 60 * 60 * 1000
  });
```

PR #1489, opened 2025-11-26
feat: add error logs only toggle to logs panel in timepoint explorer

markjmeier:

Introduces a toggle to only show error logs.  For checks with lots of logs, these can be hard to find/surface.

It's worth considering if we should default this to on, if there are any error logs present. We should also consider if theres potentially a better way to filter logs. this is a quick and dirty fix for users with a large number of logs lines in a check

n/a

Fixes #

image

### DIFF
--- a/src/scenes/components/LogsRenderer/LogsEvent.tsx
+++ b/src/scenes/components/LogsRenderer/LogsEvent.tsx
@@ -1,6 +1,6 @@
-import React, { useMemo, useState } from 'react';
+import React, { useMemo } from 'react';
 import { dateTimeFormat, GrafanaTheme2 } from '@grafana/data';
-import { Box, InlineSwitch, useStyles2 } from '@grafana/ui';
+import { Box, Text, useStyles2 } from '@grafana/ui';
 import { css, cx } from '@emotion/css';
 import { MSG_STRINGS_HTTP } from 'features/parseCheckLogs/checkLogs.constants.msgs';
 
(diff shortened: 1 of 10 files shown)


src/scenes/components/LogsRenderer/LogsRaw.tsx
```diff
@@ -29,9 +30,29 @@ const LOGS_HEIGHT = 400;
 
 export const LogsRaw = <T extends UnknownParsedLokiRecord>({ logs }: { logs: T[] }) => {
   const [width, setWidth] = useState(0);
+  const [errorOnly, setErrorOnly] = useState(false);
```
w1kman: Nit-ish: `errorLogsOnly` seems like a better fit, based on what it controls.

src/scenes/components/LogsRenderer/LogsRaw.tsx
```diff
@@ -29,9 +30,29 @@ const LOGS_HEIGHT = 400;
...
+      const level = log.labels?.level || log.labels?.detected_level;
+      return level?.toLowerCase() === 'error';
+    });
+  }, [logs, errorOnly]);
 
   return (
     <div>
+      <Box paddingBottom={1} display="flex" justifyContent="flex-end">
+        <InlineSwitch
+          label="Error logs only"
```
w1kman: Enabling this will throw an error

src/scenes/components/LogsRenderer/LogsEvent.tsx
```diff
@@ -17,33 +17,55 @@ export const LogsEvent = <T extends ParsedLokiRecord<Record<string, string>, Rec
   mainKey: string;
 }) => {
   const styles = useStyles2(getStyles);
+  const [errorOnly, setErrorOnly] = useState(false);
```
w1kman: Nit-ish: `errorLogsOnly` seems like a better fit, based on what it controls.

w1kman (changes requested):
The component explodes when "Error logs only" is enabled at the same time as "Raw logs".

https://github.com/user-attachments/assets/d8600fe0-5ed7-4f05-bd14-a06d9c851ab4

IF we intend to fix the raw logs switch, I'd suggest moving the state outside of the two components, so that they share the state of the Switch. It somewhat frustrating to toggle the switch again when switching to raw and vice versa.

**The switch**
- The label for the switch has a `pointer`, but clicking it does nothing. 
- The label text/color does not follow the styles of a form field

https://github.com/user-attachments/assets/c4e80c31-6d76-47fc-8c9e-2c8768199e07

Suggestion: Use an `InlineField` (with appropriate flex-box styling)  with a `Switch` as child. It will allow for the label to be clickable as well as styling the label.

Image

**No empty state**
When there are no error logs, there is "nothing".
Consider adding an empty state that clearly indicates that stuff is hidden and can be displayed

Image

Image

**Thoughts**
Ideally, it should be possible to filter on any log level.

ckbedwell:
Going to close this as we are going to address this in upcoming work (but keep them coming!)

PR #1496, opened 2025-11-28
feat: suggestion for duplicate functionality

ckbedwell:
Suggestion for simplifying the duplication logic.

I wanted to explore simplifying the duplicate logic as I felt like 90% of the functionality was already in place to handle this.

Targeting the branch that is removing the merge logic as that's where I started to reduce the noise in the PR.

### DIFF
--- a/src/components/Checkster/contexts/ChecksterContext.tsx
+++ b/src/components/Checkster/contexts/ChecksterContext.tsx
@@ -211,9 +211,7 @@ export function ChecksterProvider({
     [formMethods, isNew, onCheckTypeChange, stashCurrentValues]
   );
 
-  const canChangeCheckType = useMemo(() => {
-    return isNew || isDuplicate;
-  }, [isNew, isDuplicate]);
+  const canChangeCheckType = isNew || isDuplicate;
 
   useEffect(() => {
     if (isDuplicate) {


src/components/Checkster/components/form/ChooseCheckType.tsx
```diff
@@ -11,7 +11,7 @@ import { getCheckTypeOption } from '../../utils/check';
 import { StyledField } from '../ui/StyledField';
 
 export function ChooseCheckType() {
-  const { changeCheckType, checkType, isNew } = useChecksterContext();
+  const { changeCheckType, checkType, canChangeCheckType } = useChecksterContext();
```
ckbedwell, line 14: Abstract this so it is simple to understand. Passes the responsibility for this logic upstream.

src/components/Checkster/contexts/ChecksterContext.tsx
```diff
@@ -225,7 +212,18 @@ export function ChecksterProvider({
...
+  const canChangeCheckType = useMemo(() => {
+    return isNew || isDuplicate;
+  }, [isNew, isDuplicate]);
+
+  useEffect(() => {
+    if (isDuplicate) {
+      formMethods.setValue('job', `${check?.job} (Copy)`, { shouldDirty: true });
+      formNavigation.completeAllSteps();
+    }
+  }, [formMethods, check, isDuplicate, formNavigation]);
```
ckbedwell, line 221: This effect handles all of the special duplication logic for the form.

src/page/NewCheck/NewCheckV2.tsx
```diff
@@ -70,12 +81,15 @@ export function NewCheckV2() {
...
-          disabled={isOverlimit || !canWriteChecks}
+        <ChecksterProvider
           checkType={checkType || CHECK_TYPE_GROUP_DEFAULT_CHECK[group.value]}
-          onSave={handleSubmit}
+          disabled={isOverlimit || !canWriteChecks}
           onCheckTypeChange={handleCheckTypeChange}
-        />
+          check={duplicateCheck}
+          isDuplicate={!!duplicateCheck}
+        >
```
ckbedwell, line 90: I couldn't see a good reason why not to add the Provider here so it is the same pattern as `<EditCheckV2 />`.

src/components/Checkster/contexts/ChecksterContext.tsx
```diff
@@ -225,7 +211,18 @@ export function ChecksterProvider({
     [formMethods, isNew, onCheckTypeChange, stashCurrentValues]
   );
 
-  const value = useMemo(() => {
+  const canChangeCheckType = useMemo(() => {
+    return isNew || isDuplicate;
+  }, [isNew, isDuplicate]);
```
w1kman: No need to use `useMemo` as it is a primitive type as well as more or less zero computation to set value (compared to checking dependency array, it's way more performant to skip the memo). 

```suggestion
  const canChangeCheckType = isNew || isDuplicate;
```

VikaCep (dismissed):
Much cleaner after removing the unneeded previous logic to handle backwards compatibliltiy 💪 💯

src/components/Checkster/contexts/ChecksterContext.tsx
VikaCep: I like you kept the responsibility to add the `(Copy)` portion and make the form dirty at this point as in this [commit](https://github.com/grafana/synthetic-monitoring-app/pull/1305/commits/6188334bd442893009d8fd7616c333de4b38169b#diff-36e864582eca1a66cd88faeba19140a3f89235e239f878aa5df70c882f9c0828R256)

PR #1523, opened 2025-12-18
feat: add option to show terraform imports as blocks

VikaCep:
## Description

Adds Terraform import blocks as an alternative format for importing resources, with improved UX through syntax highlighting and a tabbed interface.

### What changed

- Added support for Terraform `import {}` blocks alongside existing CLI commands
- Implemented tabbed interface to switch between CLI commands and import blocks formats

### Why these changes

Terraform import blocks offer significant performance improvements over CLI commands, especially when working with remote state backends (e.g., GCS buckets). As suggested by the team, import blocks are processed as a single operation by Terraform and can be removed once resources are imported, making them cleaner for migration workflows.

However, CLI commands remain valuable for users who:
- Are already familiar with this workflow
- Have simpler, local state setups
- Prefer copy-paste commands for quick imports

image
image

### DIFF
--- a/src/hooks/useTerraformConfig.ts
+++ b/src/hooks/useTerraformConfig.ts
@@ -27,9 +27,9 @@ function generateImportBlocks(
   probesConfig: TFProbeConfig,
   probes: Probe[]
 ): {
-  checkImportBlocks: string;
-  checkAlertsImportBlocks: string;
-  probeImportBlocks: string;
+  checkImportBlocks: string[];
+  checkAlertsImportBlocks: string[];
+  probeImportBlocks: string[];
 } {
   const checkBlocks = checks.map((check) => {
     const resourceName = generateCheckResourceName(check);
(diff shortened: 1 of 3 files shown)


src/hooks/useTerraformConfig.ts
```diff
@@ -22,6 +22,48 @@ import {
...
+    return `import {
+  to = grafana_synthetic_monitoring_probe.${probeName}
+  id = "${probeId}:<PROBE_ACCESS_TOKEN>"
+}`;
+  });
+
+  return {
+    checkImportBlocks: checkBlocks.join('\n\n'),
+    checkAlertsImportBlocks: checkAlertsBlocks.join('\n\n'),
+    probeImportBlocks: probeBlocks.join('\n\n'),
```
w1kman: Nit: I'd probably "just" return the data without converting each property to string, and leave the presentation logic to the consumer (the component).

src/page/ConfigPageLayout/tabs/CodeBlockDisplay.tsx
```diff
@@ -0,0 +1,90 @@
...
+      borderRadius: theme.shape.radius.default,
+      border: `1px solid ${theme.colors.border.medium}`,
+      overflowY: 'auto',
+      maxHeight: '600px',
+    }),
+    code: getPrismCodeStyle(theme),
+    copyButton: css({
+      marginLeft: 'auto',
+    }),
+  };
```
w1kman, line 88: Nit: template styles are easier to read and are more type safe (assuming you have code quality tools for CSS).

w1kman (dismissed):
LGTM 👌🏻

PR #1530, opened 2026-01-07
fix: k6 channels endpoint URL and UI display

VikaCep:
## Problem

The k6 channels API endpoint was missing the proper URL prefix, and the UI was concatenating `.x` to channel names even though the backend already includes this suffix in the channel names.

### Issues addressed:

1. **Missing URL prefix**: The API call to `/sm/channels/k6` was not using the full instance URL, causing routing issues
2. **Redundant `.x` suffix**: The UI was adding `.x` to channel names (e.g., `v1` → `v1.x`) when rendering the channel selector, but the backend already returns channel names with the `.x` suffix included

### 1. Added `instanceSettings.url` prefix to k6 channels endpoint

Updated the `listK6Channels()` method in `DataSource.ts` to use the full instance URL.

### 2. Removed `.x` suffix concatenation from UI

The backend API already returns channel names with the `.x` suffix (e.g., `"v1.x"`, `"v2.x"`), so the UI no longer needs to add it. Updated `K6ChannelSelect.tsx` to display the channel names as-is:

image

### DIFF
--- a/src/components/CheckEditor/FormComponents/ChannelDetails.test.tsx
+++ b/src/components/CheckEditor/FormComponents/ChannelDetails.test.tsx
@@ -11,21 +11,21 @@ describe('ChannelDetails', () => {
   const mockChannels: K6Channel[] = [
     {
       id: 'v1',
-      name: 'v1',
+      name: 'v1.x',
       default: false,
       deprecatedAfter: '2125-12-31T00:00:00Z', // Far into the future, so it's not deprecated yet'
       manifest: 'k6>=1',
     },
     {
       id: 'v2',
-      name: 'v2',
+      name: 'v2.x',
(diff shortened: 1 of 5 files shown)


VikaCep:
Looks like this decision is being reverted in the API and the original implementation is going to be the chosen one, so this PR shouldn't be merged (https://github.com/grafana/synthetic-monitoring-api/pull/1874)

ckbedwell:
~~I spoke @ka3de earlier this week and I think they are going to keep the original implementation.~~

~~@ka3de is that correct? Shall we wait until you have confirmed the direction you will take before we close this PR.~~

I need to read comments before I leave my own.. 🤦  @VikaCep let's close this.

PR #1531, opened 2026-01-07
Fix routing preview visualization

VikaCep:
## Problem

The alert routing preview feature was failing with the following error:

```
Plugin failed to load
Error: Warning: Middleware for RTK-Query API at reducerPath "notifications.alerting.grafana.app/v0alpha1" has not been added to the store.
You must add the middleware for RTK-Query to function correctly!
```

or

```
Plugin failed to load
Details
TypeError: g.current.isRequestSubscribed is not a function

    at ZE (https://plugins-cdn.grafana.net/grafana-synthetic-monitoring-app/1.43.0/public/plugins/grafana-synthetic-monitoring-app/962.93e179f421b17fdeb963.js:1044:17700)
    at div
    at div
    at Cw (https://plugins-cdn.grafana.net/grafana-synthetic-monitoring-app/1.43.0/public/plugins/grafana-synthetic-monitoring-app/962.93e179f421b17fdeb963.js:1044:37516)
```

depending whether it's run with Graft or without it.

This error started appearing after the `12.3.0` Grafana release. The root cause is that the `useMatchInstancesToRouteTrees` hook from `@grafana/alerting` uses RTK-Query internally, which requires middleware to be registered in the Redux store. Plugins don't have access to configure Grafana's main Redux store, and it breaks when trying to run it.

## Solution

Created a custom implementation of `useMatchInstancesToRouteTrees` that bypasses the RTK-Query middleware requirement while maintaining full compatibility with the original hook.

The implementation:
- Manually fetches routing tree data using `getBackendSrv()`
- Uses the standalone `matchInstancesToRouteTrees` function from `@grafana/alerting` to process the data
- Returns the same interface as the original hook, making it a drop-in replacement

**Note:** This should be a temporary solution as the `grafana/alerting` hook should not break because of this. The alerting team has been notified of the problem.

Before:
image

After: 

image

### DIFF
--- a/src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
+++ b/src/components/CheckForm/AlertsPerCheck/AlertRoutingPreview.tsx
@@ -1,6 +1,5 @@
 import React, { useMemo } from 'react';
 import { useFormContext } from 'react-hook-form';
-import { useMatchInstancesToRouteTrees } from '@grafana/alerting';
 import { GrafanaTheme2 } from '@grafana/data';
 import { Alert, Icon, LoadingPlaceholder, Text, TextLink, useStyles2 } from '@grafana/ui';
 import { css } from '@emotion/css';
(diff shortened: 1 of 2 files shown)


w1kman (approved):
LGTM 🥳

VikaCep:
The hook's fix is in `grafana/alerting@12.4.0-20823894528` (as per https://github.com/grafana/grafana/pull/116016) but it [fails](https://github.com/grafana/grafana/actions/runs/20823907227) to publish to npm so we can't consume it. Merging this fix until we can upgrade the package's version.

PR #1532, opened 2026-01-08
fix: upgrade grafana/alerting package

VikaCep:
## Problem
The alert routing preview feature was failing with the following error:

```
Error: Warning: Middleware for RTK-Query API at reducerPath "notifications.alerting.grafana.app/v0alpha1" has not been added to the store.
You must add the middleware for RTK-Query to function correctly!
```

This error started appearing after the `12.3.0` Grafana release. The root cause is that the `useMatchInstancesToRouteTrees` hook from `@grafana/alerting` uses RTK-Query internally, which requires middleware to be registered in the Redux store. Plugins don't have access to configure Grafana's main Redux store, and it breaks when trying to run it. 
In Grafana 12.3, the middleware was registered under the old path (`notifications.alerting/v0alpha0`), but the newer alerting package expects it under `notificationsAlertingAPIv0alpha1`. Grafana 12.4+ registers it under the new path.

In order to fix it while the alerting team worked on a fix, https://github.com/grafana/synthetic-monitoring-app/pull/1531 was implemented. Now that we can again consume `useMatchInstancesToRouteTrees` from the alerting package, the workaround code added in the linked PR can now be removed.

## Solution
- Upgraded `@grafana/alerting` to canary, which now depends on the publicly published `@grafana/api-clients` package
- Added `moduleResolution: "bundler"` to `tsconfig.json` because `@grafana/api-clients` uses subpath exports which require bundler resolution for TypeScript to resolve the types correctly
- Replaced deep imports into package internals (`@grafana/alerting/dist/...`, `@grafana/data/dist/...`, `@testing-library/user-event/index`) that bundler resolution no longer allows
- The alert routing preview feature requires **Grafana 12.4+** at runtime. On older versions it degrades gracefully via the existing `ErrorBoundary`

See the full conversation for extra context  https://raintank-corp.slack.com/archives/C041QUZ4SJ3/p1772042085833859?thread_ts=1759838034.080519&cid=C041QUZ4SJ3

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,7 +6,7 @@
   "ts-node": {
     "compilerOptions": {
       "module": "commonjs",
-      "target": "es5",
+      "target": "es6",
       "esModuleInterop": true,
       "moduleResolution": "node"
     },


tsconfig.json
```diff
@@ -1,3 +1,15 @@
 {
-  "extends": "./.config/tsconfig.json"
+  "extends": "./.config/tsconfig.json",
+  "compilerOptions": {
+    "moduleResolution": "bundler"
+  },
+  "ts-node": {
+    "compilerOptions": {
+      "module": "commonjs",
+      "target": "es5",
```
g3john: I wonder if we could target es6?

VikaCep: Changed it and there are no issues, so let's go with it!

PR #1540, opened 2026-01-19
[fix] Normalize Loki data frames regardless of schema format

ckbedwell:
## Problem

The Timepoint Explorer was failing to correctly parse and display logs when Loki data frames arrived in the old schema format (using "Time" and "Line" field names) instead of the new schema format (using "timestamp" and "body" field names). This inconsistency caused the Timepoint Explorer to break when encountering data frames with the older schema, preventing users from viewing execution logs correctly.

## Solution

This PR adds normalization logic that detects the schema format of incoming Loki data frames and converts them to a consistent new schema format before parsing. The normalization function handles both old and new schema formats, ensuring backward compatibility while standardizing on the new schema. This allows the Timepoint Explorer to work reliably regardless of which schema format the Loki data frames use.

The fix includes comprehensive test coverage for both schema formats to ensure the normalization works correctly in all scenarios.

### DIFF
--- a/src/features/parseLokiLogs/parseLokiLogs.ts
+++ b/src/features/parseLokiLogs/parseLokiLogs.ts
@@ -22,7 +22,9 @@ export function parseLokiLogs<T, R>(dataFrame: LokiDataFrame<T, R>, parser?: Fie
   const normalizedDataFrame = normalizeLokiDataFrame(dataFrame);
   // After normalization, fields are guaranteed to be LokiFields<T, R>
   if (!isLokiFields(normalizedDataFrame.fields)) {
-    throw new Error('Failed to normalize LokiDataFrame fields');
+    // Log error but return empty array to prevent UI crashes
+    console.error('Failed to normalize LokiDataFrame fields', normalizedDataFrame);
+    return [];
   }
   const flattenedLogs = flattenLogs(normalizedDataFrame.fields, parser);
   const sortedLogs = sortLogs(flattenedLogs);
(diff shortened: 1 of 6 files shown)


src/features/parseLokiLogs/parseLokiLogs.ts
```diff
@@ -1,111 +1,141 @@
...
+type FieldParser = Partial<Record<LokiFieldNamesOld | LokiFieldNames, (value: any) => any>>;
 
 // Expects structured log data from Loki. LogQL queries must include '| logfmt'
 // parser to extract individual fields before calling this function.
 export function parseLokiLogs<T, R>(dataFrame: LokiDataFrame<T, R>, parser?: FieldParser) {
   const normalizedDataFrame = normalizeLokiDataFrame(dataFrame);
   // After normalization, fields are guaranteed to be LokiFields<T, R>
-  const flattenedLogs = flattenLogs(normalizedDataFrame.fields as LokiFields<T, R>, parser);
+  if (!isLokiFields(normalizedDataFrame.fields)) {
+    throw new Error('Failed to normalize LokiDataFrame fields');
```
VikaCep: is this error visible in the ui?

src/scenes/components/TimepointExplorer/TimepointListEntryReachability.tsx
```diff
@@ -47,7 +47,7 @@ export const TimepointListEntryReachability = ({ timepoint }: TimepointListEntry
   const executionsToRender = Object.values(statefulTimepoint.probeResults)
     .flat()
     .filter((execution) => {
-      const probeSuccess = execution[LokiFieldNames.Labels].probe_success;
+      const probeSuccess = execution[LokiFieldNamesOld.Labels].probe_success;
```
VikaCep: why are you referencing `LokiFieldNamesOld`? from the component's side shouldn't it be agnostic of whether it's new or old and just use something like `LokiFieldNames`? Same in all other references in this file.

PR #1543, opened 2026-01-19
fix: align channel data structure with backend API specification

VikaCep:
## Description

Aligns the frontend's k6 channel data structure with the new backend API specification. The channel information is now stored at the root level (`channels.k6`) instead of nested within check type settings. See https://github.com/grafana/synthetic-monitoring/issues/408#issuecomment-3595507029 for its definition and discussion.

## Backend Context

The backend API now expects/returns channel data in this format:

**Request (create/update check):**
```
{
  "channels": {
    "k6": {
      "id": "v1"
    }
  }
}
```

**Response (fetching checks):**

```
{
  "channels": {
    "k6": {
      "id": "v1",
      "name": "v1.x",
      "default": true,
      "manifest": "^1.0.0",
      "deprecatedAfter": "2027-06-01T00:00:00Z"
    }
  }
}
```

Related backend PR: https://github.com/grafana/synthetic-monitoring-api/pull/1876


ckbedwell (approved):
LGTM! I much prefer this to the previous schema.

Issue #1227, opened 2025-08-13
Migrate Select and MultiSelect components to Combobox


VikaCep:
After https://github.com/grafana/synthetic-monitoring-app/pull/1226, we started using `grafana/ui` v12 which deprecates `Select` components in favor of `Combobox`. We should replace its usage as currently the lint warning is being silenced.

`MultiSelect` should also be moved to `Combobox` as they were recently deprecated in 12.3.0.

See https://github.com/grafana/plugin-tools/blob/674424824e159fec79c7917b916b66e690a9062a/docusaurus/docs/migration-guides/update-from-grafana-versions/v11.6.x-v12.0.x.md?plain=1#L160

Shisui-Genjutsu:
Can I take this?

PR #1553, opened 2026-01-23
chore: migrate from Select to Combobox

g3john:
## Summary
`Select` has been deprecated in favour of `Combobox`

Addresses https://github.com/grafana/synthetic-monitoring-app/issues/1227

## Changes
- Replaced all uses of `Select` with `Combobox`
- Added `Fields` for labels
- Added combobox test util functions
- UI changes:
Screenshot 2026-01-23 at 3 20 27 PM

## Additional context
Some features that existed in `Select` is not available in `Combobox`. [Slack thread](https://raintank-corp.slack.com/archives/C025WTVRBSN/p1767819595524819?thread_ts=1767805209.446389&cid=C025WTVRBSN)

### DIFF
--- a/src/page/CheckList/components/CheckListHeader.tsx
+++ b/src/page/CheckList/components/CheckListHeader.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { GrafanaTheme2, SelectableValue } from '@grafana/data';
 import { Button, Checkbox, Combobox, Field, Icon, Stack, Tooltip, useStyles2 } from '@grafana/ui';
-import { css, cx } from '@emotion/css';
+import { css } from '@emotion/css';
 import { DataTestIds } from 'test/dataTestIds';
 
 import { CheckFiltersType, CheckListViewType, FilterType } from 'page/CheckList/CheckList.types';


src/components/CheckForm/AlertsPerCheck/FailedExecutionsAlert.tsx
```diff
@@ -122,14 +122,14 @@ export const FailedExecutionsAlert = ({
...
-                <Select // eslint-disable-line @typescript-eslint/no-deprecated
+                <Combobox
                   {...fieldProps}
                   disabled={!selected || formState.disabled}
                   data-testid={CHECKSTER_TEST_ID.feature.perCheckAlerts[alert.type].periodCombobox}
                   id={`alert-period-${alert.type}`}
                   options={validPeriods}
                   value={field.value}
-                  onChange={({ value = null } = {}) => {
+                  onChange={({ value = null }: { value?: string | null }) => {
```
VikaCep, line 135: Can we also add 

```
 width="auto"
minWidth={8}
maxWidth={10}
```

So it looks like the latency alerts?

Image

src/components/CheckForm/AlertsPerCheck/FailedExecutionsAlert.tsx
```diff
@@ -122,14 +122,14 @@ export const FailedExecutionsAlert = ({
               const { ref, ...fieldProps } = field; // ref is unused, this is to silence warnings
 
               return (
-                <Select // eslint-disable-line @typescript-eslint/no-deprecated
+                <Combobox
```
VikaCep, line 125: When periods display invalid options those should be disabled so users can't select them:

Image

This happens when you select a check frequency that's not so small, for example 30 min. 

Image

Can we disable these options in that case with the new combobox? (same for latency alerts)

src/components/AlertRuleForm.tsx
```diff
@@ -269,8 +268,13 @@ export const AlertRuleForm = ({ canEdit, rule, onSubmit }: Props) => {
...
                       const { ref, ...rest } = field;
                       return (
-                        // eslint-disable-next-line @typescript-eslint/no-deprecated
-                        <Select {...rest} options={TIME_UNIT_OPTIONS} aria-label="Time unit" disabled={!canEdit} />
+                        <Combobox
+                          {...rest}
+                          options={TIME_UNIT_OPTIONS}
+                          disabled={!canEdit}
+                          value={field.value.value}
+                          data-testid="alert-rule-form-time-unit-combobox"
```
VikaCep: Maybe you can move that testId to `DataTestIds` in `src/test/dataTestIds.ts`

src/components/CheckFormAlert.tsx
```diff
@@ -37,15 +36,16 @@ export const CheckFormAlert = () => {
         </p>
         <p>Tip: adding multiple probes can help to prevent alert flapping for less frequent checks</p>
       </div>
-      <Field label="Select alert sensitivity" data-fs-element="Alert sensitivity select">
+      <Field label="Select alert sensitivity" htmlFor="alert-sensitivity-select" data-fs-element="Alert sensitivity select">
```
VikaCep, line 40: <3 <3

src/components/constants.ts
```diff
@@ -502,36 +502,28 @@ export const METHOD_OPTIONS = [
   },
 ];
 
-export const headerNameOptions = [
```
VikaCep, line 504: This was unused right? Nice catch!

src/page/CheckList/components/CheckFilters.tsx
```diff
@@ -86,63 +86,69 @@ export function CheckFilters({ onReset, onChange, checks, checkFilters, includeS
...
-              aria-label="Filter by status"
-              data-testid="check-status-filter"
-              options={CHECK_LIST_STATUS_OPTIONS}
+            <Field label="Status" htmlFor="check-status-filter" data-fs-element="Status select" className={css({
+              marginBottom: 0,
+            })}>
+              <Combobox
+                id="check-status-filter"
+                aria-label="Filter by status"
+                data-testid="check-status-filter"
```
VikaCep: Maybe you can move that testId to `DataTestIds` in `src/test/dataTestIds.ts`

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -119,20 +119,17 @@ export const CheckListHeader = ({
...
-          options={CHECK_LIST_SORT_OPTIONS}
-          defaultValue={CHECK_LIST_SORT_OPTIONS[0]}
-          width={25}
-          onChange={onSort}
-          value={sortType}
-        />
+        <Field label="Sort" htmlFor="sort-by-select" data-fs-element="Sort by select">
+          <Combobox
+            id="sort-by-select"
+            data-testid="sort-checks-by-combobox"
```
VikaCep: Maybe you can move that testId to `DataTestIds` in `src/test/dataTestIds.ts`

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -119,20 +119,17 @@ export const CheckListHeader = ({
...
+            prefixIcon='sort-amount-down'
```
VikaCep: Seems like this icon is not working? 

Image

Also, maybe it would be better to display the label horizontally by setting the `horizontal={true}` prop in the `Field` component, as to prevent adding extra height on this field.

src/page/CheckList/CheckList.bulkActions.test.tsx
```diff
@@ -174,18 +175,16 @@ describe('CheckList - Bulk Actions', () => {
     expect(checksA[0]).toHaveTextContent(`89280 executions / month`);
     expect(checksA[1]).toHaveTextContent(`44640 executions / month`);
 
-    const sortPicker = await screen.getByLabelText('Sort checks by');
-    await user.click(sortPicker);
-    await user.click(screen.getByText(`Asc. Executions`, { selector: 'span' }));
+    await selectOption(user, { dataTestId: 'sort-checks-by-combobox', option: 'Asc. Executions' });
```
VikaCep: If we move the `testIds` to the constants file, then we can reference them here instead 👍

src/components/AlertRuleForm.tsx
```diff
@@ -227,8 +227,7 @@ export const AlertRuleForm = ({ canEdit, rule, onSubmit }: Props) => {
                   <Controller
```
ckbedwell, line 228: Total scope creep here as this wasn't part of the original ticket but `react-hook-form` has a hook called `useController` which is a much nicer pattern then having this container and the render props.

Feel free to make a ticket and boot that to tech debt for another PR (and fix them all in one fell swoop) or can incrementally upgrade them as we see them.

src/page/CheckList/CheckList.filters.test.tsx
```diff
@@ -152,55 +151,55 @@ describe('CheckList - Filtering', () => {
...
 
-    const select = await getSelect({ label: `Filter by type` });
-    await user.click(select[0]);
-    const listBox = screen.getByLabelText(`Select options menu`);
+    const dialog = getModalContainer();
+    const combobox = await getCombobox({ label: `Type` }, dialog);
+    await user.click(combobox);
 
-    expect(within(listBox).getByText(`Scripted`)).toBeInTheDocument();
+    expect(await screen.findByText('Scripted', {}, { timeout: 3000 })).toBeInTheDocument();
```
VikaCep: Not a big deal, but it calls my attention that we need a timeout here since the values are loaded from a constants file. Is there any other UI element we can wait for to be loaded instead of relying on the timeout? I see in the next test we do the following instead:

`await screen.findAllByText('HTTP'); // Wait for the options to render`

src/components/constants.ts
```diff
@@ -37,7 +37,7 @@ export const DNS_RESPONSE_CODES = enumToStringArray(DnsResponseCodes).map((respo
   value: responseCode,
 }));
 
-export const DNS_RESPONSE_MATCH_OPTIONS = [
+export const DNS_RESPONSE_MATCH_OPTIONS: Array<ComboboxOption<ResponseMatchType>> = [
```
ckbedwell, line 39: Do we need to be explicit about these types? The combobox shouldn't rely on them being explicit here?

The reason I care about this is because I am still haunted by this [file](https://github.com/grafana/synthetic-monitoring-app/blob/main/src/test/fixtures/CONSTANTS_TEMP.ts)  

__TEMP <--> 22 months ago_ 😅😅😅

src/page/CheckList/CheckList.constants.ts
```diff
@@ -1,8 +1,8 @@
-import { SelectableValue } from '@grafana/data';
+import { ComboboxOption } from '@grafana/ui';
 
 import { CheckEnabledStatus } from 'types';
 
-export const CHECK_LIST_STATUS_OPTIONS: Array<SelectableValue<CheckEnabledStatus>> = [
+export const CHECK_LIST_STATUS_OPTIONS: Array<ComboboxOption<CheckEnabledStatus>> = [
```
ckbedwell, line 5: Same question here as above.

src/components/CheckForm/AlertsPerCheck/FailedExecutionsAlert.tsx
g3john: Just checked, `Combobox` doesn't have the functionality to disable options. Fortunately it shows an error tooltip and doesn't allow you to save so we'll leave it as is

src/components/constants.ts
g3john: We don’t _need_ to type them, but since they’re exported and reused, it helps prevent accidental widening and breaking changes

Do you foresee the explicit typing to cause issues?

src/components/AlertRuleForm.tsx
g3john: Created a ticket for us to track https://github.com/grafana/synthetic-monitoring-app/issues/1558

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -85,7 +86,7 @@ export const CheckListHeader = ({
         <div>
           Currently showing {currentPageChecks.length} of {checks.length} total checks
         </div>
-        <div className={styles.stack}>
+        <div className={cx(styles.stack, styles.defaultGap)}>
```
VikaCep: This should be replaced by 

`<Stack gap={2} alignItems="center">`

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -102,15 +103,15 @@ export const CheckListHeader = ({
         </div>
       </div>
       <div className={styles.row}>
-        <div className={styles.stack}>
+        <div className={cx(styles.stack, styles.defaultGap)}>
```
VikaCep:
```suggestion
        <Stack gap={2} alignItems="center">
```

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -119,20 +120,19 @@ export const CheckListHeader = ({
...
-              <Icon name="sort-amount-down" /> Sort
-            </div>
-          }
-          options={CHECK_LIST_SORT_OPTIONS}
-          defaultValue={CHECK_LIST_SORT_OPTIONS[0]}
-          width={25}
-          onChange={onSort}
-          value={sortType}
-        />
+        <div className={cx(styles.stack, styles.sortChecksGap)}>
```
VikaCep:
```suggestion
    <Stack gap={0.5} alignItems="center">
```

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -149,6 +149,18 @@ const getStyles = (theme: GrafanaTheme2) => ({
   stack: css({
     alignItems: `center`,
     display: `flex`,
+  }),
+  defaultGap: css({
     gap: theme.spacing(2),
   }),
+  sortChecksGap: css({
+    gap: theme.spacing(0.5),
+  }),
```
VikaCep:
```suggestion

```

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -149,6 +149,18 @@ const getStyles = (theme: GrafanaTheme2) => ({
   stack: css({
     alignItems: `center`,
     display: `flex`,
+  }),
+  defaultGap: css({
     gap: theme.spacing(2),
   }),
```
VikaCep:
```suggestion

```

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -149,6 +149,18 @@ const getStyles = (theme: GrafanaTheme2) => ({
   stack: css({
     alignItems: `center`,
     display: `flex`,
+  }),
```
VikaCep:
```suggestion

```

PR #1555, opened 2026-01-23
feat: migrate home dashboard to react scenes

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1560

## Problem

While all other dashboards use react-scenes, the dashboard displayed in the home page (summary) was still using the old scenes version (no react). 

## Solution

Migrated the Summary/Home dashboard to `@grafana/scenes-react` to match the pattern used by other dashboards.

### Key Changes

**Removed:**
- All old scene-based implementations (summaryScene.ts, errorRateMap.ts, latencyTimeseries.ts, etc.)

image


ckbedwell (approved):
It adds the check IDs to the summary table but I quite like that! Great job 👏

PR #1559, opened 2026-01-27
chore: address deprecated config.apps from GrafanaBootConfig

VikaCep:
# Addressing deprecated config.apps from GrafanaBootConfig

Closes https://github.com/grafana/synthetic-monitoring/issues/455

## IMPORTANT:

This should be merged **after** grafana 12.4 is released, which is scheduled on Feb 24, 2026 (See [thread](https://raintank-corp.slack.com/archives/C01C4K8DETW/p1768412397235989?thread_ts=1768370687.760569&cid=C01C4K8DETW))

## Problem
Grafana is deprecating `config.apps` from `GrafanaBootConfig` as part of their multitenancy work to enable asynchronous loading of plugin-related metadata. The deprecation was introduced in `Grafana 12.4` and `config.apps` will be removed entirely in `Grafana 13.2`.
Synthetic Monitoring was using the deprecated API in one location:

```
// src/components/App.tsx
version: config.apps['grafana-synthetic-monitoring-app'].version
```

This was used to initialize Faro with the plugin version at module scope.

## Solution
- Replaced `config.apps['grafana-synthetic-monitoring-app'].version` with `getAppPluginVersion('grafana-synthetic-monitoring-app')` from` @grafana/runtime`
- Updated `@grafana/data`, `@grafana/runtime`, `@grafana/schema`, and `@grafana/ui` from `^12.1.0` to `^12.4.0`
- Updated `grafanaDependency` in `plugin.json` from `>=12.0.0` to `>=12.4.0`

The new `getAppPluginVersion()` function provides the same functionality and is the recommended migration path for plugin authors as per Grafana's migration guide.

### DIFF
--- a/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx
+++ b/src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx
@@ -102,6 +102,7 @@ describe('SecretEditModal', () => {
       plaintext: 'secret-value',
     };
 
+    // fireEvent.change is used here because user.type is unreliable with these react-hook-form controlled inputs after the @grafana/ui 12.4 upgrade
     fireEvent.change(screen.getByLabelText(/Name/), { target: { value: inputValues.name } });
     fireEvent.change(screen.getByLabelText(/Description/), { target: { value: inputValues.description } });
     fireEvent.change(screen.getByLabelText(/Value/), { target: { value: inputValues.plaintext } });


src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx
```diff
@@ -1,6 +1,5 @@
 import React from 'react';
-import { screen, waitFor } from '@testing-library/react';
-import userEvent from '@testing-library/user-event';
+import { fireEvent, screen, waitFor } from '@testing-library/react';
```
g3john, line 2: Is `fireEvent` and getting `user` from the `render` preferred over `userEvent`?

src/components/ConfirmLeavingPage/ConfirmUnsavedModal.test.tsx
```diff
@@ -47,7 +47,9 @@ describe('<ConfirmUnsavedModal>', () => {
     const { user } = render(<ConfirmUnsavedModal onLeavePage={jest.fn()} onStayOnPage={onStayOnPage} />);
 
     await screen.findByTestId(DataTestIds.ConfirmUnsavedModalHeading);
-    await user.keyboard('{Esc}');
+    const dialog = screen.getByRole('dialog');
+    await user.click(dialog);
+    await user.keyboard('{Escape}');
```
g3john, line 52: Did the upgrade start causing this to fail?

src/test/jest-setup.tsx
```diff
@@ -49,6 +49,12 @@ global.IntersectionObserver = jest.fn(() => ({
   thresholds: [],
 }));
 
+global.ResizeObserver = jest.fn(() => ({
+  observe: jest.fn(),
+  unobserve: jest.fn(),
+  disconnect: jest.fn(),
+}));
```
g3john, line 56: Haven't seen this before! What does it do? Also curious about the `IntersectionObserver` above

src/components/ConfirmLeavingPage/ConfirmUnsavedModal.test.tsx
VikaCep: Yes, the `@grafana/ui` 12.4 upgrade changed how the Modal component handles keyboard events. The dialog element now needs focus before it responds to Escape. Previously `user.keyboard('{Esc}')` worked without focusing anything, but after the upgrade we need to click the dialog first to give it focus, then dispatch the keyboard event

src/test/jest-setup.tsx
VikaCep: `ResizeObserver` is a browser API that doesn't exist in jsdom (Jest's test environment). The `@grafana/ui` 12.4 upgrade introduced internal usage of `ResizeObserver` in some components, so without this mock tests crash with `ReferenceError: ResizeObserver is not defined`. The `IntersectionObserver` mock above serves the same purpose. That one was added during the React Router v6 migration.

src/page/ConfigPageLayout/tabs/SecretsManagementTab/SecretEditModal.test.tsx
VikaCep: `userEvent` is generally preferred, but `fireEvent.change` works best for populating the form inputs in this component. After the upgrade, `user.type` doesn't reliably set values on these inputs and tests fail in CI as the form fields end up empty or partially filled, which causes react-hook-form validation to fail and prevents form submission. `fireEvent.change` avoids this by setting the full value in a single event. `user.click` and `user.type` are still used for all other interactions (buttons, label fields, etc.). I've added a comment explaining the workaround.

PR #1566, opened 2026-01-29
fix: normalize Loki data frames in check test button

VikaCep:
### Problem
The check test button was broken when Loki returned data frames using the new schema format (with field names `timestamp` and `body`) instead of the old schema format (with field names `Time` and `Line`). This caused the ad-hoc check logs to fail to parse, preventing users from seeing the results of their check tests.

This is the same issue that was fixed for the Timepoint Explorer in PR #1540.

### Solution
This PR applies the same normalization pattern from PR #1540 to the check test button's log parsing logic.
A new `normalizeDataFrameJSON()` function has been added that converts old schema field names to new schema field names before processing. This function reuses the existing `normalizeLokiDataFrame()` logic by converting `DataFrameJSON` to `DataFrame`, normalizing, and converting back. This ensures both functions use identical normalization logic and stay in sync.
The `useAdHocLogs` hook now normalizes incoming frames before extracting log data, allowing it to handle both old and new Loki schema formats seamlessly.

### DIFF
--- a/src/components/Checkster/feature/adhoc-check/useAdHocLogs.ts
+++ b/src/components/Checkster/feature/adhoc-check/useAdHocLogs.ts
@@ -1,10 +1,9 @@
 import { useQuery } from '@tanstack/react-query';
 import { DataFrameJSON } from '@grafana/data';
-
 import { normalizeDataFrameJSON } from 'features/parseLokiLogs/parseLokiLogs';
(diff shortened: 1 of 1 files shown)


ckbedwell (approved):
LGTM Thank you for turning this around so fast ❤️

PR #1573, opened 2026-01-30
chore: migrate to React Router v6

### LINKED ISSUE #713: App plugin: Update to react-router v6
Starting from Grafana 10, plugins can start using the v6 of react-router. Overall, react-router v6 aims to simplify route configuration and provide a more flexible and intuitive API for developers.

### Why?
We are going to deprecate using react-router v5 in plugins from Grafana 11, and we would like to make sure our internal app plugins are using the v6 version by the time we finish migrating Grafana core itself.

### How?
We have created a migration guide for plugins (please see the links below), and also don't hesitate to reach out in `#grafana-plugins-platform` in Slack in case you have any questions.

- [x] Install `react-router-dom-v5-compat`
- [x] Use `v6` `<Routes />` and `<Route />`
- [x] Use `v6` hooks instead of `v5` hooks 
- [x] Wait for [scenes](https://github.com/grafana/scenes/issues/608)
- [x] Change dependency from `react-router-dom-v5-compat` to `react-router-dom@6`
- [x] Update `.cprc.json` (`"useReactRouterV6": true`)
- [ ] Update test setup for `render` (perhaps we still need to use the Compat setup?)

- [How to migrate a Grafana app plugin to react-router v6](https://grafana.com/developers/plugin-tools/migration-guides/update-from-grafana-versions/migrate-10_0_x-to-10_1_x#update-to-react-router-v6)
- [Official React Router v5 to v6 migration guide](https://reactrouter.com/en/main/upgrading/v5)
- [Community topic](https://community.grafana.com/t/migrating-app-plugins-to-use-react-router-v6/115410)

VikaCep:
# Migrate to React Router v6

Closes #713

## Problem
The application was using `react-router-dom-v5-compat` as a bridge between React Router v5 and v6. This compatibility layer:

- Added unnecessary bundle size overhead
- Required maintaining two different routing patterns in the codebase
- Is deprecated and will be removed in future Grafana versions

Additionally, the test infrastructure relied on React Router v5 APIs through `@grafana/ui` components (`Link`, `TextLink`) and `@grafana/runtime`'s `locationService`, which caused test failures after migrating to v6.

## Solution
This PR completes the migration to React Router v6 by:

**Runtime changes:**
- Upgraded `react-router-dom` from v5-compat to v6.27.0
- Removed `react-router-dom-v5-compat` dependency entirely
- Updated all navigation to use v6 APIs (`useNavigate`, `useParams`, `useLocation`)
- Updated programmatic navigation to use `locationService` from `@grafana/runtime` 
    ==> `locationService.push()/replace()` is preferred as it's a Grafana's abstraction, integrates with Grafana's location tracking/observability. `useNavigate()`, React Router's native hook, works but bypasses Grafana's layer

**Test infrastructure updates:**
- Refactored `test/render.tsx` to use `Router` with `locationService.getHistory()` so both React Router and locationService share the same history instance
- Updated `@grafana/runtime` mock for `locationService` implementation including `history.block()` support required by `ConfirmLeavingPage`
- Updated `@grafana/ui` mock for `Link` and `TextLink` components to work without React Router v5 context (these components internally depend on v5)

**Dependencies:**
- React upgraded to 18.3.0
- Grafana packages upgraded to 12.2.0
- Updated `@grafana/create-plugin` to v6.7.8
- Enabled `useReactRouterV6` feature flag

## Notes

The test mocks for `@grafana/ui` components (`Link`, `TextLink`) are necessary because these components internally use React Router v5 APIs. Once Grafana updates `@grafana/ui` to support v6 natively these mocks should be simplified or removed.

### DIFF
--- a/src/components/Checkster/components/form/FormHttpAuthenticationField.tsx
+++ b/src/components/Checkster/components/form/FormHttpAuthenticationField.tsx
@@ -32,7 +32,7 @@ export function FormHttpAuthenticationField({ basicAuthField, bearerTokenField }
   const bearerToken = watch(bearerTokenField) as string | undefined; // TODO: Fix casting
 
   const authType = getHttpAuthType(basicAuth, bearerToken);
-  const dismountAuthType = useRef<HTTPAuthType | undefined>(authType);
+  const dismountAuthType = useRef<HTTPAuthType>(authType);
 
   // Ensure form values are in sync with authType
   const handleChangeAuthType = (value: HTTPAuthType) => {
--- a/src/components/Checkster/utils/form.ts
+++ b/src/components/Checkster/utils/form.ts
@@ -65,7 +65,7 @@ export function getFieldErrorProps<T extends CheckFormFieldPath = CheckFormField
 export function getHttpAuthType(
   basicAuth?: { username?: string; password?: string },
   bearerToken?: string
-): HTTPAuthType | undefined {
+): HTTPAuthType {
   if (basicAuth && (basicAuth.username !== undefined || basicAuth.password !== undefined)) {
     return HTTPAuthType.BasicAuth;
   }


src/components/ConfirmLeavingPage/ConfirmLeavingPage.test.tsx
```diff
@@ -1,6 +1,5 @@
 import React, { PropsWithChildren } from 'react';
-import { Router } from 'react-router-dom';
-import { CompatRouter, Route, Routes } from 'react-router-dom-v5-compat';
+import { Route, Routes, unstable_HistoryRouter as HistoryRouter } from 'react-router-dom';
```
g3john: Is there an alternative to using `unstable_HistoryRouter` that we can use instead?

VikaCep: Good that you pointed this out! I was using `unstable_HistoryRouter` so the test router and the component could share the same history instance from `locationService.getHistory()` because `ConfirmLeavingPage` calls `history.block()` directly. But I could swap it out for just using the `<Router>` component with manual location state syncing, which is the same pattern used in the render helper.

g3john (approved):
LGTM!

src/components/Checkster/components/form/FormHttpAuthenticationField.tsx
```diff
@@ -32,7 +32,7 @@ export function FormHttpAuthenticationField({ basicAuthField, bearerTokenField }
   const bearerToken = watch(bearerTokenField) as string | undefined; // TODO: Fix casting
 
   const authType = getHttpAuthType(basicAuth, bearerToken);
-  const dismountAuthType = useRef<HTTPAuthType>(authType);
+  const dismountAuthType = useRef<HTTPAuthType | undefined>(authType);
```
ckbedwell: Nit: this never resolves to undefined when I look at `getHttpAuthType`

PR #1574, opened 2026-01-30
feat: improve Timepoint Explorer selection UX

g3john:
# What this PR does / why we need it :
This PR makes the following changes
- Dim non-selected bars slightly
- Scroll to logs when a bar/entry is clicked
- Updated empty logs placeholder text
# Which issue(s) this PR fixes:
This PR addresses the first `Interaction patterns to improve responsiveness & discoverability` item in [this doc](https://docs.google.com/document/d/1V4GoA-rZvz7Ehu28378mprwC1AHxxwYEceKTgC9_Hp8/edit?tab=t.0#heading=h.ew56e4277plk)

https://github.com/user-attachments/assets/c0c04b42-5a24-409c-91e6-5a9bba5806d2

### DIFF
--- a/src/scenes/components/TimepointExplorer/__tests__/TimepointExplorer.test.tsx
+++ b/src/scenes/components/TimepointExplorer/__tests__/TimepointExplorer.test.tsx
@@ -18,7 +18,7 @@ import {
 import { TimepointExplorer } from 'scenes/components/TimepointExplorer/TimepointExplorer';
 import { DataTestIds } from 'test/dataTestIds';
 import { mockFeatureToggles } from 'test/utils';
-import { FeatureName } from 'types';
+import { FeatureName, HTTPCheck } from 'types';
 import {
   REF_ID_EXECUTION_LIST_LOGS,
   REF_ID_EXECUTION_VIEWER_LOGS,
@@ -27,6 +27,14 @@ import {
 } from 'scenes/components/TimepointExplorer/TimepointExplorer.constants';
 
 const baseTime = Date.now();
+
+const TIME_MODIFIED_HTTP_CHECK: HTTPCheck = {
+  ...BASIC_HTTP_CHECK,
+  frequency: 60000,
+  timeout: 10000,
+  created: Math.floor((baseTime - 60 * 60 * 1000) / 1000),
+  modified: Math.floor((baseTime - 10 * 60 * 1000) / 1000),
+};
 
 function setupMSWHandlers() {
   server.use(
@@ -80,7 +88,7 @@ function setupMSWHandlers() {
 
 
 function renderTimepointExplorer() {
-  return <TimepointExplorer check={BASIC_HTTP_CHECK} />;
+  return <TimepointExplorer check={TIME_MODIFIED_HTTP_CHECK} />;
 }
 
 const mockScrollIntoView = jest.fn();


src/scenes/components/TimepointExplorer/TimepointViewer.tsx
```diff
@@ -154,6 +165,7 @@ const getStyles = (theme: GrafanaTheme2) => ({
     display: flex;
     flex-direction: column;
     gap: ${theme.spacing(2)};
+    scroll-margin-top: 25vh;
```
g3john, line 168: This scrolls to 25% of the view height above the logs, so that the user will be able to see a little bit of the graph

src/scenes/components/TimepointExplorer/__tests__/TimepointExplorer.mocks.ts
```diff
@@ -1,9 +1,11 @@
+import { dateTime } from '@grafana/data';
+
 jest.mock('@grafana/scenes-react', () => {
   const actual = jest.requireActual('@grafana/scenes-react');
 
   let mockTimeRange = {
-    from: new Date('2024-01-01T00:00:00Z'),
-    to: new Date('2024-01-02T00:00:00Z'),
+    from: dateTime('2024-01-01T00:00:00Z'),
+    to: dateTime('2024-01-02T00:00:00Z'),
```
g3john: The [XAxis component](https://github.com/grafana/synthetic-monitoring-app/blob/56f169cde28bc9debfe9edd0bd72f4c6f94addf7/src/scenes/components/TimepointExplorer/XAxis.tsx#L41) is expecting a `DateTime` object. It didn't throw an error before because we are not mocking `timepoints` which now renders the `XAxisContent`

src/scenes/components/TimepointExplorer/__tests__/TimepointExplorer.test.tsx
```diff
@@ -10,6 +10,61 @@ import { TimepointExplorer } from 'scenes/components/TimepointExplorer/Timepoint
...
+}));
+
+jest.mock('scenes/components/TimepointExplorer/TimepointViewer.hooks', () => ({
+  useTimepointLogs: jest.fn(() => ({
+    data: [],
+    isFetching: false,
+    isLoading: false,
+    refetch: jest.fn(),
+  })),
+}));
```
ckbedwell: I'm not necessarily against mocking the hook here but my preference is we use MSW to mock the base dependency. The hook is an implementation detail so if we were to change it at some point in the future we would break the tests, which undermines one of their main benefits. If we aim to mock as high as possible (our out-of-repo dependencies) and test from the perspective of a user we create the most durable tests.

So, try and mock the server response instead here and see how that looks/feels. The Time Point Explorer does have a lot of inter-connected dependencies which is why it doesn't have many tests so we might find this approach is a pain in the ass as it creates a horrible web and what you've done here is the cleanest solution but give it a go and see how you find it!

src/scenes/components/TimepointExplorer/__tests__/TimepointExplorer.test.tsx
```diff
@@ -10,6 +10,61 @@ import { TimepointExplorer } from 'scenes/components/TimepointExplorer/Timepoint
...
+
+const mockScrollIntoView = jest.fn();
+Element.prototype.scrollIntoView = mockScrollIntoView;
```
ckbedwell: Can this get reset and/or moved out of the global scope if possible?

src/scenes/components/TimepointExplorer/TimepointExplorer.hooks.ts
```diff
@@ -475,48 +473,6 @@ export function useCurrentAdjustedTime(check: Check) {
...
 
-interface UseIsInitialisedProps {
-  check: Check;
-  currentAdjustedTime: UnixTimestamp;
-  handleViewerStateChange: (viewerState: ViewerState) => void;
-  isLoading: boolean;
-  timepoints: StatelessTimepoint[];
-}
-
-export function useIsInitialised({
```
ckbedwell, line 486: How confident are we we don't need this?

ckbedwell (changes requested):
Great changes. 👏 

Main CR is just exploring if MSW would be better rather than mocking the hook.

src/scenes/components/TimepointExplorer/TimepointExplorer.hooks.ts
g3john: I'll put it back. I initially had to remove it because the initialization selects a timepoint, so it would auto-scroll on page-load. I then added the `shouldScrollToViewer` flag that only enables on timepoint click, so that issue won't be the case anymore

g3john:
> Was just about to approve this then noticed there is one place you've missed that doesn't scroll to the logs viewer (it is fairly hidden 😅)
> 
>  Screen.Recording.2026-02-03.at.17.16.39.mov 
> Once that is done, this is going in!

Ah thanks for calling that one out. Added!

PR #1575, opened 2026-02-02
fix: add validation to reject leading/trailing whitespace in target field

VikaCep:
## Summary

- Adds validation to reject leading/trailing whitespace in the target/instance field
- Prevents a bug where trailing whitespace causes dashboard metrics to fail to load (see https://github.com/grafana/support-escalations/issues/20484)

## Problem

A customer reported that after duplicating a browser check, the dashboard showed no metrics even though the check was running successfully. Investigation revealed:

1. The check was saved with a trailing space in the Instance field: `"FRANCE "` instead of `"FRANCE"`
2. Metrics were generated with `instance="FRANCE "` (with trailing space)
3. When viewing the dashboard, URL parameter handling strips trailing whitespace: `var-instance=FRANCE`
4. The probe variable query `sm_check_info{instance="$instance"}` uses `"FRANCE"` (no space)
5. No match -> empty probe dropdown -> dashboard shows "N/A" for all metrics

The main Checks list page displayed data correctly because it queries metrics directly without the URL variable interpolation.

## Solution

Add validation to show an error when the target field contains leading or trailing whitespace:

> "Target cannot have leading or trailing whitespace"

This approach:
- Forces users to manually fix the whitespace, making them aware of the change
- Avoids silently transforming values which could break Terraform configs or metric continuity
- Matches the pattern used by `jobSchema` for rejecting invalid characters

image

### DIFF
--- a/src/schemas/forms/BrowserCheckSchema.ts
+++ b/src/schemas/forms/BrowserCheckSchema.ts
@@ -1,6 +1,5 @@
 import { createFrequencySchema } from 'schemas/general/Frequency';
 import { createTimeoutSchema } from 'schemas/general/Timeout';
-import { noLeadingTrailingWhitespace } from './utils/validation';
 import { z, ZodType } from 'zod';
 
 import { BrowserSettings, CheckFormValuesBrowser, CheckType, ProbeWithMetadata } from 'types';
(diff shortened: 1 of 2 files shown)


src/schemas/general/Target.ts
```diff
@@ -4,4 +4,12 @@ const MAX_CHARACTERS = 2040;
...
-  .max(MAX_CHARACTERS, { message: `Target must be ${MAX_CHARACTERS} characters or less` });
+  .max(MAX_CHARACTERS, { message: `Target must be ${MAX_CHARACTERS} characters or less` })
+  .superRefine((value, ctx) => {
+    if (value !== value.trim()) {
+      ctx.addIssue({
+        code: 'custom',
+        message: `Target cannot have leading or trailing whitespace`,
+      });
+    }
+  });
```
g3john: Is it worth adding a reusable `leadingTrailingWhitespaceValidation` function in `src/schemas/forms/script/validation.ts` that can take in a subject name like 'Target' or 'Instance'?

VikaCep: Good idea, done!

Issue #1556, opened 2026-01-26
Add plugin internationalization (i18n) support


VikaCep:
# Add plugin internationalization (i18n) support

## Problem

The Synthetic Monitoring plugin currently ships English-only UI strings. Grafana’s how-to guide for plugin internationalization ([https://grafana.com/developers/plugin-tools/how-to-guides/plugin-internationalization](https://grafana.com/developers/plugin-tools/how-to-guides/plugin-internationalization)) provides a standard workflow. We should adopt that workflow so the plugin can be translated, load locale files dynamically, and give contributors a clear, supported process for adding translations. This ticket is scoped to implementing the i18n mechanism and proving the flow with a handful of translations, not translating every string.

## Completion criteria

**Phase 1: Implement the i18n mechanism and a small proof of concept**

* **Implement the i18n mechanism**: declare supported languages, initialise plugin translations, add extraction tooling and ESLint i18n rules.
* **Proof of concept translations**: add `src/locales/en-US/<plugin-id>.json` and `src/locales/es-ES/<plugin-id>.json` (or another second language) with a small set of keys.
* **Verification**: confirm translations load at runtime.
* **Docs**: add a `README/CONTRIBUTING` blurb describing how to run the extraction, where translation files live, and how to test locally.

**Phase 2: Crowdin support**

* Request Crowdin access (via Okta) and run the Crowdin “create project” workflow to scaffold a project and obtain the `crowdin_project_id`. Project name should match the repo name.
* Add `crowdin.yml` pointing to the `en-US` source files and add the Crowdin GitHub workflows to the repo: `i18n-verify`, `i18n-crowdin-upload`, `i18n-crowdin-download`, and `i18n-crowdin-create-tasks`.
* Verify the upload/download PRs and the `i18n-verify` job behave as expected; document the Crowdin project id and any repo-specific workflow changes in README/CONTRIBUTING.

The goal is to enable the translation workflow and prove it works; a follow-up ticket can expand coverage across the whole plugin.

## Resources

* Phase 1 docs: [https://grafana.com/developers/plugin-tools/how-to-guides/plugin-internationalization](https://grafana.com/developers/plugin-tools/how-to-guides/plugin-internationalization)
[...]

PR #1589, opened 2026-02-11
feat: add i18n support (Phase 1)

g3john:
# What this PR does / why we need it
We want to add i18n support to the SM plugin. This PR sets up i18n for contribution, and adds some translations as a proof of concept

Screenshot 2026-02-11 at 4 00 52 PM

# Which issue(s) this PR fixes

Implements Phase 1 of https://github.com/grafana/synthetic-monitoring-app/issues/1556

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2598,11 +2598,6 @@
     redux-thunk "^3.1.0"
     reselect "^5.1.0"
 
-"@remix-run/router@1.20.0":
-  version "1.20.0"
-  resolved "https://registry.yarnpkg.com/@remix-run/router/-/router-1.20.0.tgz#03554155b45d8b529adf635b2f6ad1165d70d8b4"
-  integrity sha512-mUnk8rPJBI9loFDZ+YzPGdeniYK+FTmRD1TMCz7ev2SNIozyKKpnGgsxO34u6Z4z/t0ITuu7voi/AshfsGsgFg==
-
 "@remix-run/router@1.23.2":
   version "1.23.2"
   resolved "https://registry.yarnpkg.com/@remix-run/router/-/router-1.23.2.tgz#156c4b481c0bee22a19f7924728a67120de06971"
@@ -9270,7 +9265,7 @@ react-custom-scrollbars-2@4.5.0:
     prop-types "^15.5.10"
     raf "^3.1.0"
 
-"react-data-grid@github:grafana/react-data-grid#a922856b5ede21d55db3fdffb6d38dc76bdc7c58":
+react-data-grid@grafana/react-data-grid#a922856b5ede21d55db3fdffb6d38dc76bdc7c58:
   version "7.0.0-beta.56"
   resolved "https://codeload.github.com/grafana/react-data-grid/tar.gz/a922856b5ede21d55db3fdffb6d38dc76bdc7c58"
   dependencies:
@@ -9465,13 +9460,6 @@ react-router@5.3.4:
     react-is "^16.6.0"
     tiny-invariant "^1.0.2"
     tiny-warning "^1.0.0"
-
-react-router@6.27.0:
-  version "6.27.0"
-  resolved "https://registry.yarnpkg.com/react-router/-/react-router-6.27.0.tgz#db292474926c814c996c0ff3ef0162d1f9f60ed4"
-  integrity sha512-YA+HGZXz4jaAkVoYBE98VQl+nVzI+cVI2Oj/06F5ZM+0u3TgedN9Y9kmMRo2mnkSK2nCpNQn0DVob4HCsY/WLw==
-  dependencies:
-    "@remix-run/router" "1.20.0"
 
 react-router@6.30.3:
   version "6.30.3"


.config/.cprc.json
```diff
@@ -1,4 +1,4 @@
 {
-  "version": "6.7.8",
+  "version": "6.8.4",
```
g3john, line 2: auto updated when running `npx @grafana/create-plugin@latest update`

CONTRIBUTING.md
```diff
@@ -75,4 +75,152 @@ We use the file nesting feature to help manage the growing number of files in th
 }
 ```
```
g3john, line 77: Is this the best place to put this documentation?

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
+{
+  "addNewCheckButton": {
+    "createNewCheck": "Crear nuevo cheque"
```
VikaCep:
```suggestion
    "createNewCheck": "Crear nueva prueba"
```

I think "prueba" sounds more natural than "chequeo". 
Cheque on the other hand refers to the one from the bank 😄

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "checkFilterGroup": {
+    "activeFiltersText": "({{activeFilters}} activo)",
+    "filterTitle": "Filtros adicionales {{activeFiltersText}}"
+  },
+  "checkFilters": {
+    "alerts": "Alertas",
+    "all": "Todos",
+    "allProbes": "Todas las sondas",
```
VikaCep:
```suggestion
    "allProbes": "Todos los agentes",
```
I'd use "agentes" directly. Otherwise "ubicaciones" could work as well.

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+    "filterByAlertsAriaLabel": "Filtrar por alertas",
+    "filterByStatusAriaLabel": "Filtrar por estado",
+    "filterByTypeAriaLabel": "Filtrar por tipo",
+    "probes": "Sondas",
```
VikaCep:
```suggestion
    "probes": "Agentes",
```

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+    "searchChecksAriaLabel": "Buscar chequeos",
```
VikaCep:
```suggestion
    "searchChecksAriaLabel": "Buscar pruebas",
```

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+    "searchPlaceholder": "Buscar por nombre de trabajo, punto final o etiqueta",
```
VikaCep:
```suggestion
    "searchPlaceholder": "Buscar por nombre del job, endpoint o etiqueta",
```
I'd keep the technical terms as they are, since they don't translate well.

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+    "type": "Tipo",
+    "withAlerts": "Con alertas",
+    "withoutAlerts": "Sin alertas"
+  },
+  "checkList": {
+    "header": {
+      "currentlyShowing": "Actualmente se muestran {{currentPageChecksLength}} de {{checksLength}} chequeos totales",
```
VikaCep:
```suggestion
      "currentlyShowing": "Mostrando {{currentPageChecksLength}} de {{checksLength}} pruebas",
```

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+      "deselectAll": "Deseleccionar todos",
+      "selectAll": "Seleccionar todos",
+      "selectAllAriaLabel": "Seleccionar todos",
+      "setThresholds": "Establecer umbrales",
```
VikaCep:
```suggestion
      "setThresholds": "Definir umbrales",
```

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+      "sortOptions": {
+        "ascExecutions": "Ejecuciones asc. ",
```
VikaCep:
```suggestion
        "ascExecutions": "Ejecuciones asc.",
```

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+        "ascReachability": "Accesibilidad asc. ",
```
VikaCep:
```suggestion
        "ascReachability": "Accesibilidad asc.",
```

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+        "descExecutions": "Ejecuciones desc. ",
```
VikaCep:
```suggestion
        "descExecutions": "Ejecuciones desc.",
```

src/locales/es-ES/grafana-synthetic-monitoring-app.json
```diff
@@ -0,0 +1,41 @@
...
+        "descReachability": "Accesibilidad desc. "
```
VikaCep:
```suggestion
        "descReachability": "Accesibilidad desc."
```

src/module.tsx
```diff
@@ -1,10 +1,14 @@
 import React, { lazy, Suspense } from 'react';
 import { AppPlugin, AppRootProps } from '@grafana/data';
+import { initPluginTranslations } from '@grafana/i18n';
 import { Spinner } from '@grafana/ui';
 import { type SMPluginConfigPageProps } from 'configPage/PluginConfigPage';
+import pluginJson from 'plugin.json';
 
 import { ProvisioningJsonData } from './types';
 
+await initPluginTranslations(pluginJson.id);
```
VikaCep: If we add this here, then I believe we can remove it from https://github.com/grafana/synthetic-monitoring-app/blob/feat/add-i18n-support/src/components/ScenesProvider.tsx#L7

src/locales/es-ES/grafana-synthetic-monitoring-app.json
VikaCep: It's also how it's set in the breadcrumbs 

Image

CONTRIBUTING.md
VikaCep: I think that's a good location for it.
Just a heads up that the developer-facing stuff here is solid and will stick around (how to use` t()/Trans`, key naming, best practices). But the workflow sections (extraction, adding languages, file structure) will probably need a rewrite once we get Crowdin set up in Phase 2.

src/module.tsx
VikaCep: Having it in `module.tsx`  produces a a `+281.38%` increase according to Danger.js. 
The bundle size increase is because it imports whole package to the main bundle that Grafana loads upfront.
We could avoid this by initialising translations inside the `lazy()` loaders instead, so it loads alongside the app code on demand rather than upfront.

g3john:
> Looking good! What's the plan for when we say this PR is ready to go in?
> 
> I just looked at the parent issue and I think a missing piece is updating our CI/CD workflow to integrate this. At the moment it seems easy to add a new translation key but forget to run the i18n extract script. Do we have plans to address that?

@ckbedwell 
Good call! At first thought I figure it was fine since it won't break anything if they did forget to run the script, but contributors working on adding translations wouldn't even know the string exists if they weren't extracted.

Added a workflow to the CI to validate, and fails if they forgot to extract

VikaCep (approved):
👏 👏 👏  t('pr.approved', 'Great work!')

src/page/CheckList/components/CheckListHeader.tsx
```diff
@@ -110,7 +115,7 @@ export const CheckListHeader = ({
               indeterminate={isSomeSelected}
               value={isAllSelected}
               disabled={checks.length === 0}
-              aria-label="Select all"
+              aria-label={t('checkList.header.selectAllAriaLabel', 'Select all')}
```
ckbedwell, line 118: As this matches what is on line 82 we should use the same key. We should document that we will default to brevity for shared strings and then if they need to be split up at a later date we will add it.

I know the Grafana team had a conversation about how to handle shared strings which are repeatedly used throughout their codebase but I don't know what conclusions they arrived at and what their key naming conventions for those are. Could be worth looking up?

Not for this PR (_if ever... juice / squeeze matrix_ ) but I wonder if there is a solution where for common strings we can use Grafana's translation files rather than us duplicating them, too?

g3john: Good question! I couldn't find anything on slack or google docs relating to this, so I started a [thread](https://raintank-corp.slack.com/archives/C025WTVRBSN/p1771605145842059)

g3john: Seems like the sentiment is to use separate keys, since the two “Select all” strings have different use cases — one is tooltip content and the other is an aria-label

Another argument we could make is that the same word in English can have different meanings, and may need different translations depending on context and location

With that in mind, if we were to put common strings in a centralized translation file though, we’d need to be explicit about the context and intended use for each string

Issue #1558, opened 2026-01-27
Use react-hook-form's useController hook over Controller component


g3john:
`react-hook-form` has a hook called `useController` which is a much nicer pattern then having the Controller container and the render props.

Controller component pattern [example](https://github.com/grafana/synthetic-monitoring-app/blob/55a6d4856036dc97ba9a2e754cbe49955f1152cd/src/components/AlertRuleForm.tsx#L227-L235)

useController hook pattern [example](https://github.com/grafana/synthetic-monitoring-app/blob/55a6d4856036dc97ba9a2e754cbe49955f1152cd/src/components/CheckEditor/FormComponents/Frequency.tsx#L54-L60)

[Reference comment by @ckbedwell](https://github.com/grafana/synthetic-monitoring-app/pull/1553#discussion_r2728476293)

PR #1590, opened 2026-02-12
chore: useController hook over Controller component

g3john:
## Summary

Replaces all usages of `react-hook-form`'s `Controller` render-prop component with the `useController` hook across the codebase, resolving #1558.

### DIFF
--- a/src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.tsx
+++ b/src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo } from 'react';
-import { useController, useFormContext } from 'react-hook-form';
+import { useFormContext } from 'react-hook-form';
 import { GrafanaTheme2 } from '@grafana/data';
 import { Field, Stack, TextLink, useStyles2 } from '@grafana/ui';
 import { css } from '@emotion/css';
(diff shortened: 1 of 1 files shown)


src/components/CheckForm/AlertsPerCheck/FailedExecutionsAlert.tsx
```diff
@@ -114,32 +118,22 @@ export const FailedExecutionsAlert = ({
           invalid={!!periodError}
           error={periodError}
           validationMessageHorizontalOverflow={true}
+          disabled={!selected || formState.disabled}
```
g3john, line 121: Now that we're not using `Controller`, `InlineField` needs the `disabled` prop which propagates onto the child

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.tsx
```diff
@@ -53,7 +53,9 @@ export const AlertsPerCheck = () => {
     revalidateForm<CheckFormValues>(`alerts.${type}`);
   };
 
-  const selectedAlerts = getValues('alerts');
+  const selectedAlerts = watch('alerts');
+
+  useController({ control, name: 'alerts' });
```
ckbedwell: Do we need the controller here or is this all managed by the AlertsList already?

ckbedwell (approved):
One small question where I think there is some redundancy but otherwise LGTM!

We usually have to deconstruct all the refs and omit them from being passed down as we aren't using forwardRef wrappers for our components so I was about to suggest while we're at it in this PR let's reduce some of the console noise in the tests but I think when we move to React 19 it has killed the need to use `forwardRef` wrappers so I'm happy to leave that work and see if it gets killed when we transition?

src/components/CheckForm/AlertsPerCheck/AlertsPerCheck.tsx
g3john: Ah good catch. I switched the `getValues` to `watch` to see if we could remove `useController` but forgot to actually remove it

ckbedwell (approved):
LGTM!

PR #1594, opened 2026-02-17
fix: check test error state

g3john:
## What this PR does / why we need it
[Context](https://raintank-corp.slack.com/archives/C0175SS6SA3/p1771235434192789)
- [Switches from Script to Execution when it shouldn't]: Fix race condition in `useFormNavigationState` where errors state lagged `formErrors` by one render cycle
- [Doesn't switch from Execution to Script when it should]: Fix `BROWSER_CHECK_FIELDS` registering 'instance' instead of 'target', causing the target field to be unhandled by `sectionByErrors` on Browser checks. 

https://github.com/user-attachments/assets/a69b5f90-e6e8-4912-974a-7945eea775a5


ckbedwell:
Great job! Can we update it so that the test button engages the validation in the stepper too like it does with the save button and manual navigation?

image

g3john:
> Great job! Can we update it so that the test button engages the validation in the stepper too like it does with the save button and manual navigation?
> 
> image

Updated the PR description video with the change

PR #1610, opened 2026-03-07
feat: add screenshot thumbails to timepoint explorer

peterschretlen:
## Problem

Browser check users have no way to see visual snapshots of what a page looked like during a check execution. When debugging failures (e.g. "timeout waiting for a selector"), they have to rely on text logs alone, which makes it difficult to understand what the browser actually rendered. This is a highly requested capability, particularly from customers running k6 browser checks at scale.

## Solution

Add experimental screenshot thumbnail support to the timepoint explorer and adhoc check views, gated behind the `synthetic-monitoring-screenshots` feature flag.

The feature works via a two-log-line contract: the k6 script captures a screenshot, pushes the data to Loki (either as chunked base64 or as a URL reference to external storage like GCS), and writes a `screenshot:<uuid>` discovery line to the execution logs. The frontend detects these UUIDs, fetches the screenshot data, and renders inline thumbnails with a click-to-enlarge modal.

Two example scripts are included in the browser check editor's Examples tab (also gated behind the FF): one for Loki storage (base64 chunks) and one for GCS storage (image uploaded to a bucket, URL reference in Loki). Both storage approaches are handled transparently by the same frontend code.

## Additional context

- The feature is behind an experimental FF (`synthetic-monitoring-screenshots`) and is not visible to any customers unless explicitly enabled.
- The screenshot module is isolated under `src/scenes/components/LogsRenderer/screenshots/` with its own types, utilities, hooks, and component. Swapping the storage backend in the future only requires changes to the hook.
- Storage implications have been discussed with the team. For a controlled POC with a handful of customers, Loki ingestion is manageable. For broader rollout, the Loki team should be consulted around log line sizes (~200KB per chunk) and ingestion volume.


src/components/BrowserExamplesMenu/snippets/screenshotsGCS.js
```diff
@@ -0,0 +1,119 @@
...
+// and logs a URL reference. The image bytes never pass through Loki — only a small
+// reference log line is stored. Best for large-scale checks where Loki ingestion
+// volume is a concern.
+//
+// IMPORTANT: The bucket must allow public read access for the stored objects.
+// The screenshots are rendered via <img> tags in the browser, which cannot attach
+// storage credentials. The object URLs contain UUIDs and are not guessable.
+//
+// Required secrets (Synthetic Monitoring > Config > Secrets):
+//   - sm-screenshot-loki-host: your Loki push endpoint
```
ckbedwell: So, this was a source of confusion for me when I was following along. I put in my push endpoint, like:

```
https://logs-dev-005.grafana-dev.net/loki/api/v1/push
```

Rather than just adding:

```
https://logs-dev-005.grafana-dev.net
```

Can we make it clearer what is expected here or change it to literally expect the push endpoint? It's also worth noting that because it's a http request it doesn't get redacted in the logs. Is there much point in having this as a secret -- I suppose it makes reusability easier?

Makes me think we should have a snippet library...

VikaCep: > Makes me think we should have a snippet library.

We just talked about this with Mark and the possibility of adding some of this example code as functions in jslib. It would be a good next iteration I believe.

src/scenes/components/LogsRenderer/screenshots/screenshots.hooks.ts
```diff
@@ -0,0 +1,58 @@
...
+      return;
+    }
+
+    let cancelled = false;
+
+    const fetchScreenshots = async () => {
+      try {
+        const uuidPattern = newUUIDs.join('|');
+        const expr = `{source="${SCREENSHOT_LOG_SOURCE}"} |~ "${uuidPattern}" | json`;
+        const result = await dataSource.queryLogsV2(expr, 'now-1h', 'now');
```
ckbedwell: The time range here is hard-coded so it is just looking in the last hour, that doesn't seem right? The time range should be passed in dependent on what you are looking at.

src/scenes/components/LogsRenderer/LogsEvent.tsx
```diff
@@ -1,14 +1,67 @@
...
+      const screenshotData = screenshotDataByUUID.get(match[1]);
+
+      if (!screenshotData) {
+        return log;
+      }
+
+      return {
+        ...log,
+        labels: { ...log.labels, ...screenshotData },
+      } as T;
```
ckbedwell:
```suggestion
      };
```

src/components/BrowserExamplesMenu/constants.ts
```diff
@@ -118,6 +133,7 @@ export const BROWSER_EXAMPLE_CHOICES: SelectableOptGroup[] = [
   { label: 'Keyboard Interaction', options: KEYBOARD },
   { label: 'Mouse Interaction', options: MOUSE },
   { label: 'Querying', options: QUERYING },
+  { label: 'Screenshots', options: SCREENSHOTS },
```
ckbedwell: I'd be tempted to put these first in the examples. Maybe get the new badge next to them?

src/scenes/components/LogsRenderer/screenshots/screenshots.hooks.ts
ckbedwell: This is totally optional but we need to standardise on the log querying strategy. Thomas and I were working on CheckEditor v2 and Time point Explorer simultaneously so I ended up making a hook called useInfiniteLogs and Thomas added dataSource.queryLogsV2 and they duplicate a bunch of work. I have a preference for my approach but we can discuss that in follow up work 😄 (but right now I'd be tempted to use useInfiniteLogs instead of the queryLogsV2 because it makes debugging in the network panel easier)

ckbedwell (changes requested):
Okay, so the big one for me...

It doesn't work -- I don't see screenshots being rendered in Time Point Explorer or the test panel 😅 

Add yourself to my Stack and investigate why the screenshots aren't showing (I used new Graft to force the feature toggle -- notice the UI for screenshot toggling is present)

Image

Image

[Stack](https://ckbedwellksix.grafana-dev.net/a/grafana-synthetic-monitoring-app/checks/8640?from=now-3h&to=now&timezone=browser&var-probe=$__all&var-job=Screenshot%20test%20%28loki%29&var-instance=quickpizza)
[Screenshot_base64 in Explore though](https://ckbedwellksix.grafana-dev.net/explore?schemaVersion=1&panes=%7B%22nk3%22%3A%7B%22datasource%22%3A%22grafanacloud-logs%22%2C%22queries%22%3A%5B%7B%22expr%22%3A%22%7Bsource%3D%5C%22synthetic-monitoring-agent-screenshot%5C%22%7D+%7C%7E+%5C%223d5cd0f6-0f8e-4c51-ba7e-a152c9cefe41%5C%22+%7C+json%22%2C%22refId%22%3A%22A%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22grafanacloud-logs%22%7D%2C%22editorMode%22%3A%22code%22%2C%22queryType%22%3A%22range%22%2C%22direction%22%3A%22backward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%221776268500000%22%2C%22to%22%3A%221776268800000%22%7D%2C%22panelsState%22%3A%7B%22logs%22%3A%7B%22sortOrder%22%3A%22Descending%22%7D%7D%2C%22compact%22%3Afalse%7D%7D&orgId=1)

Image

VikaCep:
> Okay, so the big one for me...
> 
> It doesn't work -- I don't see screenshots being rendered in Time Point Explorer or the test panel 😅
> 
> Add yourself to my Stack and investigate why the screenshots aren't showing (I used new Graft to force the feature toggle -- notice the UI for screenshot toggling is present)

There were actually two bugs:

- The hard-coded `now-1h` time range meant the screenshot query missed anything older than an hour. Now the timepoint's actual time range is used. 

- `extractFrameLines` was looking for a field named line, but newer Grafana returns body. It worked locally but not in cloud envs (should've used Graft from the beginning 😅) Related to https://github.com/grafana/synthetic-monitoring-app/pull/1540

All fixed now.

AWS example:
image

src/components/BrowserExamplesMenu/constants.ts
VikaCep: Done
image

VikaCep:
GCS [example](http://localhost:3000/a/grafana-synthetic-monitoring-app/checks/8391?from=now-3h&to=now&timezone=browser&var-probe=$__all&var-job=screenshots_test&var-instance=screenshots_test):

image

ckbedwell (dismissed):
We're having a [small discussion internally](https://raintank-corp.slack.com/archives/C0175SS6SA3/p1776697923069809?thread_ts=1776695805.077389&cid=C0175SS6SA3) related to the GCS example but otherwise...

LGTM! Great work getting this tidied up, secured and battle-tested @VikaCep 

And @peterschretlen for hacking it together as the original POC 👏

VikaCep:
> We're having a [small discussion internally](https://raintank-corp.slack.com/archives/C0175SS6SA3/p1776697923069809?thread_ts=1776695805.077389&cid=C0175SS6SA3) related to the GCS example but otherwise...

As Chris noted in that thread, a better alternative for the GCS [example](https://github.com/grafana/synthetic-monitoring-app/pull/1610/commits/e92fdfc30a7e216bd5476d6d3be773c5d3e04d16) is to use HMAC keys which are long lived, as opposed as the token that expired in a short amount of time. I've updated the example for GCS with this strategy instead so it provides a more adequate auth method.

image

src/components/BrowserExamplesMenu/snippets/screenshotsGCS.js
```diff
@@ -24,6 +25,7 @@ import http from 'k6/http';
 import secrets from 'k6/secrets';
 import { browser } from 'k6/browser';
 import { uuidv4 } from 'https://jslib.k6.io/k6-utils/1.4.0/index.js';
+import { AWSConfig, S3Client } from 'https://jslib.k6.io/aws/0.14.0/s3.js';
```
ckbedwell, line 28: Genuine WTF. So the **A**mazon **W**eb **S**ervices Config and the S3 (the proprietary Object storage services from Amazon Web Services) Client methods are compatible with GCS?

I've just looked up the [documentation](https://grafana.com/docs/k6/latest/javascript-api/jslib/aws/awsconfig/) on our website and I don't think it is documented at all.

Asking internally what the deal is here -- if it works it works but the naming is super confusing.

VikaCep: yes, I had this same concern but the thing is GCS offers a S3 compatible API (https://docs.cloud.google.com/storage/docs/interoperability) so they intentionally speack the same protocol as AWS S3 (including Signature V4 HMAC signing).  So any S3 client library works with GCS if you point it at storage.googleapis.com instead of s3.amazonaws.com.
I believe k6 doesn't have a dedicated GCS client library so that's why I went with that route.

> The V4 signing process lets you make signed header requests to the Cloud Storage XML API. After creating a signature using the V4 signing process, you include the signature in the Authorization header of a subsequent request, which provides authentication. You can create a signature using an RSA signature or your Amazon S3 workflow and HMAC credentials. For more details about authenticating requests, see [Signatures](https://docs.cloud.google.com/storage/docs/authentication/signatures).

PR #1612, opened 2026-03-09
Feat: integrate SM with Knowledge Graph

VikaCep:
## Problem

Synthetic Monitoring checks operate in isolation from the rest of a user's observability stack. When a check detects that an endpoint is down or latency is high, there is no built-in connection to the underlying service, its dependencies, or its current health in the broader environment. Users must manually switch between SM and other tools to understand *why* a problem is happening.

For users who have the Knowledge Graph (grafana-asserts-app) enabled, this is a missed opportunity as the KG already maps services, pods, and infrastructure with automatic insight detection, but SM checks are invisible to it.

## Solution

This PR is a **proof of concept** for connecting Synthetic Monitoring to the Knowledge Graph. It introduces four capabilities:

1. **Checks - Service integration:**  Introduces the integration between SM and the KG by modelling checks as `SyntheticCheck` entities that can be linked to KG Service entities via a `MONITORS` relationship.
image

2. **Auto-provisioning:** On plugin initialisation, SM automatically creates the `SyntheticCheck` entity rule and log config in the KG via the Asserts API if they don't already exist. This ensures SM checks are discoverable as KG entities. Full integration (the `MONITORS` relationship) requires users to set a `service_name` label on their checks.
image
image

3. **Service linking in the check editor:** Two combobox fields (`service_name`, `namespace`) in the labels section of the check form, with autocomplete from known KG services. Setting these creates a `MONITORS` relationship between the check and the service in the entity graph.
image

5. **"View in Knowledge Graph" on the check dashboard:** A deep-link button and an entity detail modal in the dashboard header. The modal fetches live entity data and displays service health, active alerts, connections, and service properties.
image
image

## Future improvements

There are a few things we'd want to explore with the Asserts team to take this further:

- **Built-in provisioning**: Right now we're creating the entity rule, log config, and potentially metrics config via the Asserts API on every plugin load. Ideally SM would be a built-in KG dataset so this configuration is provided out of the box.
- **Automatic service discovery via traces**: Currently users have to manually set the `service_name` label on a check to create the `MONITORS` relationship. If we could leverage trace IDs from check executions, the KG could discover which service a check targets and create that connection automatically.

## Notes

Video demo in this thread: 

https://raintank-corp.slack.com/archives/C0175SS6SA3/p1773109804407679?thread_ts=1770394450.006799&cid=C0175SS6SA3

### DIFF
--- a/src/components/Checkster/components/form/sections/LabelSection.tsx
+++ b/src/components/Checkster/components/form/sections/LabelSection.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { KnowledgeGraphServiceLink } from 'features/knowledgeGraph/KnowledgeGraphServiceLink';
 
 import { FormSectionName } from '../../../types';
 import { useTenantCostAttributionLabels } from 'data/useTenantCostAttributionLabels';
@@ -35,6 +36,7 @@ export function LabelSection() {
         calNames={calData?.names ?? []}
         labelLimit={maxAllowedMetricLabels}
       />
+      <KnowledgeGraphServiceLink />
     </FormSection>
   );
 }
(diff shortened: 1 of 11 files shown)


VikaCep:
This was a POC. Closing in favor of https://github.com/grafana/synthetic-monitoring-app/pull/1736

Issue #1569, opened 2026-01-30
Associate a check with existing CALs


VikaCep:
## Goal

Enable users to associate a Synthetic Monitoring check with one or more CALs during create/edit.

This involves: 

- Fetch available CAL names from SM API
- Display CAL inputs in the check form
- Support create/update flows

### Open questions

- API contract for fetching CALs
- Proposed: GET `/cals` (polled periodically) which will return just the keys if they exist.
- Shape of CAL data in the check payload
    - Empty list vs pre-populated keys with empty values
- Validation strictness for MVP (required vs optional)
- UX considerations
- What do we show if no CALs are defined for the tenant?

PR #1614, opened 2026-03-10
feat: add cost attribution labels to check create/edit

g3john:

Adds Cost Attribution Labels (CALs) to the check form. CALs are tenant-level labels fetched from a new `/sm/tenant/cals` endpoint and displayed in a dedicated section above custom labels. Users can assign values to each CAL or leave them blank (unattributed). Only CALs with values are included in the payload on save.

- New `useTenantCostAttributionLabels` hook with polling and feature flag gating (`FeatureName.CALs`)
- New `CostAttributionLabelsField` component with read-only names and editable values
- `calLabels` managed as a separate form field from `labels` to avoid React Hook Form conflicts, merged at submission time
- When editing a check, existing labels matching a CAL name are pre-filled into the CAL section
- MSW handlers and test fixtures for the new endpoint
- Integration tests covering both feature flag enabled/disabled states

Fixes https://github.com/grafana/synthetic-monitoring-app/issues/1569

- The backend endpoint is not yet available — `useTenantCostAttributionLabels` currently uses a hardcoded mock response (`Team`, `Service`). This will be replaced with `smDS.getTenantCostAttributionLabels()` once the backend is ready (see `TODO` comments in the file).
- The entire feature is gated behind `FeatureName.CALs` — when disabled, the CAL section is hidden and the hook doesn't fire.

https://github.com/user-attachments/assets/2e8be6d6-e6a7-4636-8d1e-e418ab1a0c8a

### DIFF
--- a/src/components/Checkster/components/form/layouts/GenericLabelContent.test.tsx
+++ b/src/components/Checkster/components/form/layouts/GenericLabelContent.test.tsx
@@ -139,6 +139,69 @@ describe('GenericLabelContent', () => {
       });
     });
 
+    describe('label limit accounts for CALs', () => {
+      const labelLimit = 10;
+
+      it('allows up to limit minus CAL count custom labels (2 CALs → 8 custom)', async () => {
+        renderGenericLabelContent({ calNames, labelLimit });
+
+        await waitFor(() => {
+          calNames.forEach((name) => {
+            expect(screen.getByDisplayValue(name)).toBeInTheDocument();
+          });
+        });
+
+        const addButton = screen.getByRole('button', { name: /label/i });
+        expect(addButton).not.toBeDisabled();
+      });
+
+      it('allows full limit when there are no CALs (0 CALs → 10 custom)', async () => {
+        renderGenericLabelContent({ calNames: [], labelLimit });
+
+        const addButton = screen.getByRole('button', { name: /label/i });
+        expect(addButton).not.toBeDisabled();
+      });
+
+      it('disables adding labels when custom labels fill remaining slots (2 CALs + 8 custom)', async () => {
+        const customLabels = Array.from({ length: 8 }, (_, i) => ({
+          name: `label${i}`,
+          value: `value${i}`,
+        }));
+
+        renderGenericLabelContent(
+          { calNames, labelLimit },
+          { labels: [...customLabels, { name: 'Team', value: 'team-a' }, { name: 'Service', value: 'svc-a' }] }
+        );
+
+        await waitFor(() => {
+          expect(screen.getByDisplayValue('label0')).toBeInTheDocument();
+        });
+
+        const addButton = screen.getByRole('button', { name: /label/i });
+        expect(addButton).toBeDisabled();
+      });
+
+      it('allows one more when a CAL-matching label frees a custom slot (2 CALs + 8 labels, 1 is a CAL)', async () => {
+        const customLabels = Array.from({ length: 7 }, (_, i) => ({
+          name: `label${i}`,
+          value: `value${i}`,
+        }));
+
+        renderGenericLabelContent(
+          { calNames, labelLimit },
+          { labels: [...customLabels, { name: 'Team', value: 'team-a' }] }
+        );
+
+        await waitFor(() => {
+          expect(screen.getByDisplayValue('label0')).toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


src/data/useTenantCostAttributionLabels.ts
```diff
@@ -0,0 +1,35 @@
...
+export function useTenantCostAttributionLabels() {
+  const { isEnabled: isCALsEnabled } = useFeatureFlag(FeatureName.CALs);
+
+  // TODO: Re-enable once backend is available
+  // const smDS = useSMDS();
+
+  return useQuery({
+    queryKey: QUERY_KEYS.list,
+    // TODO: Replace with smDS.getTenantCostAttributionLabels() once backend is available
+    queryFn: () => Promise.resolve(MOCK_CAL_RESPONSE),
```
VikaCep: I don't think we should return hardcoded mocks in the app code, instead I think it should just call `smDS.getTenantCostAttributionLabels()` and rely on the MSW handler for `/sm/tenant/cals` and the fixture data only when running `yarn dev:msw` (or using the Graft interceptors with `yarn:dev` otherwise)

src/data/useTenantCostAttributionLabels.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useQuery } from '@tanstack/react-query';
+
+import { FeatureName } from 'types';
+import { ONE_SECOND_IN_MS } from 'utils.constants';
+import { ListTenantCostAttributionLabelsResponse } from 'datasource/responses.types';
+import { useFeatureFlag } from 'hooks/useFeatureFlag';
+// TODO: Re-enable once backend is available
+// import { useSMDS } from 'hooks/useSMDS';
+
+export const CAL_REFETCH_INTERVAL = 10 * ONE_SECOND_IN_MS;
```
VikaCep: 10s feels a bit aggressive for this tenant-level config that probably won't change that much mid-session. Should we bump this to 30s+ to reduce unnecessary requests?

src/schemas/general/Label.ts
```diff
@@ -33,3 +33,10 @@ export const labelsSchema = z.array(labelSchema).superRefine((labels, ctx) => {
     });
   }
 });
+
+const calLabelSchema = z.object({
+  name: z.string(),
+  value: z.string().max(MAX_LENGTH, { message: `Value must be ${MAX_LENGTH} characters or less` }),
+});
```
VikaCep, line 40: Should we make the validation a bit more strict?  The way it works now allows entering whitespace-only values:

Image

Also, it allows saving custom labels that have the same name as the cost attribution labels. I think we should forbid this scenario. At the `BaseCheckSchema`  level you could add some refinement like the following to prevent it:

```
 .superRefine((data: CheckFormValuesBase, ctx) => {
      const calNames = new Set(data.calLabels.map((l) => l.name));
      const conflicts = data.labels.filter((l) => calNames.has(l.name));

      conflicts.forEach((label) => {
        ctx.addIssue({
          code: 'custom',
          path: ['labels'],
          message: `Label name "${label.name}" is already used as a cost attribution label`,
        });
      });
    });
```
And then display a validation message like: 
Image

(in order to show that message you might need to add  `<FieldValidationMessage>{errors.labels.root.message}</FieldValidationMessage>` to `GenericLabelContent`

src/components/Checkster/components/form/generic/CostAttributionLabelsField.tsx
```diff
@@ -0,0 +1,67 @@
...
+    formState: { disabled },
+  } = useFormContext<CheckFormValues>();
+  const styles = useStyles2(getStyles);
+
+  if (calNames.length === 0) {
+    return null;
+  }
+
+  return (
+    <StyledField label="Cost attribution labels" description={CAL_DESCRIPTION} required emulate>
```
VikaCep: Since values are optional, should this **not** be required?

src/components/Checkster/components/form/generic/CostAttributionLabelsField.tsx
```diff
@@ -0,0 +1,67 @@
...
+      {/* @ts-expect-error Totally valid spacing value */}
```
VikaCep: Can't we use `gap={0.5}` or `gap={1}` instead? to get rid of the error. Perhaps we should also change it in `GenericNameValueField` if we want them to be the same.

VikaCep (changes requested):
Great work!! Left a few comments/questions

PR #1637, opened 2026-03-31
feat: Add UI to list and revoke API access tokens

The-9880:
Relates to https://github.com/grafana/synthetic-monitoring/issues/529

Implements a basic UI with defined permissions for access token management:
image


ckbedwell:
@The-9880 -- is this reviewable (well, what I'm really asking is does this PR have any other upstream dependencies)?

I/we tend to ignore stuff in draft? 😬

The-9880:
> @The-9880 -- is this reviewable (well, what I'm really asking is does this PR have any other upstream dependencies)?
> 
> I/we tend to ignore stuff in draft? 😬

It's not ready for review yet - depends on https://github.com/grafana/synthetic-monitoring-api/pull/2019

src/page/ConfigPageLayout/tabs/AccessTokensTab.tsx
```diff
@@ -1,40 +1,60 @@
...
-      const token = await smDS.createApiToken();
-      setToken(token);
+  const [newToken, setNewToken] = useState<string | undefined>();
+
+  // Delete confirmation state
+  const [tokenToDelete, setTokenToDelete] = useState<TokenInfo | null>(null);
+
+  const { data, isLoading, isError, fetchNextPage, hasNextPage, isFetchingNextPage } = useTokens(
+    PAGE_SIZE,
+    canViewList
```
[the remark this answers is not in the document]
g3john: Will there ever be a case where a user has `access-tokens:write` but not `access-tokens:read`?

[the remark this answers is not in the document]
VikaCep: Yes, I believe write-without-read is possible. Not through the roles we bundle (both "Access tokens writer" and "Admin" grant all three actions together), but org admins could compose custom RBAC roles from individual plugin actions.

The problem is the fallback can't help that user anyway: the `sm/token/list` proxy route requires `access-tokens:read`, so a write-only user still triggers the query, gets a 403, and sees the "Failed to load access tokens" alert. 

Maybe we could just set `const canViewList = canReadTokens;` and drop the fallback comment above it.

src/page/ConfigPageLayout/tabs/AccessTokensTab.tsx
```diff
@@ -44,20 +64,100 @@ export function AccessTokensTab() {
...
-          disabled={!canWriteTokens}
-          onClick={() => showTokenModal()}
+          disabled={!canWriteTokens || createToken.isPending}
+          onClick={() => createToken.mutate()}
         >
           Generate access token
         </Button>
       </ConfigContent.Section>
 
+      <ConfigContent.Section title="Existing tokens">
```
VikaCep: Nit: users without `canViewList` see the "Existing tokens" heading with nothing under it. Consider gating the whole section instead so it disappears entirely for them.

src/page/ConfigPageLayout/tabs/AccessTokensTab.test.tsx
```diff
@@ -59,11 +70,216 @@ describe('AccessTokensTab', () => {
...
+
+      const revokeButtons = screen.getAllByRole('button', { name: /revoke/i });
+      await user.click(revokeButtons[0]);
+
+      const confirmButton = screen.getByTestId('data-testid Confirm Modal Danger Button');
+      await user.click(confirmButton);
+
+      // The DELETE request was made.
+      await waitFor(() => {
+        expect(deleteCallCount).toBe(1);
```
VikaCep, line 321: We could also assert the request URL contains the token id

src/page/ConfigPageLayout/tabs/AccessTokensTab.tsx
```diff
@@ -44,20 +64,100 @@ export function AccessTokensTab() {
...
+        {isLoading && canViewList && <Spinner />}
+        {isError && canViewList && (
+          <Alert severity="error" title="Failed to load access tokens">
+            Refresh the page to try again.
+          </Alert>
+        )}
+        {!isLoading && !isError && canViewList && tokens.length === 0 && <span>No tokens found.</span>}
+        {canViewList && tokens.length > 0 && (
+          <>
+            <table className={styles.table}>
```
VikaCep: We have a generic `src/components/Table`, which provides client-side pagination. Perhaps we can use it here instead, for consistency.

src/page/ConfigPageLayout/tabs/AccessTokensTab.tsx
```diff
@@ -44,20 +64,100 @@ export function AccessTokensTab() {
...
+                    >
+                      Revoke
+                    </Button>
+                  );
+
+                  return (
+                    <tr key={t.id} className={styles.tr}>
+                      <td className={cx(styles.td, styles.tdId)}>{t.id}</td>
+                      <td className={styles.td}>{formatUnixSeconds(t.created)}</td>
+                      <td className={styles.td}>{formatUnixSeconds(t.lastUsed)}</td>
```
VikaCep: I think it would look better if we show Last used as "x time ago", with a tooltip revealing the exact timestamp (dateTimeFormatTimeAgo from @grafana/data handles the relative part). "2 hours ago" vs "8 months ago" is the real liveness signal, and much easier to scan than a wall of absolute timestamps.

VikaCep (changes requested):
Left some small comments, looking great!

VikaCep:
Oh, another thing I saw: after revoking a token I don't see any confirmation in the UI.

src/page/ConfigPageLayout/tabs/AccessTokensTab.tsx
```diff
@@ -43,8 +41,70 @@ export function AccessTokensTab() {
...
+      name: 'Last used',
+      cell: (row) =>
+        row.lastUsed === 0 ? (
+          'Never'
+        ) : (
+          <Tooltip content={formatUnixSeconds(row.lastUsed)}>
+            <span>{dateTimeFormatTimeAgo(row.lastUsed * 1000)}</span>
+          </Tooltip>
+        ),
+    },
```
VikaCep, line 76:
```suggestion
    {
      name: 'Created',
      selector: (row) => row.created,
      format: (row) => formatUnixSeconds(row.created),
      sortable: true,
    },
    {
      name: 'Last used',
      selector: (row) => row.lastUsed,
      sortable: true,
      cell: (row) =>
        row.lastUsed === 0 ? (
          'Never'
        ) : (
          <Tooltip content={formatUnixSeconds(row.lastUsed)}>
            <span>{dateTimeFormatTimeAgo(row.lastUsed * 1000)}</span>
          </Tooltip>
        ),
    },
```

VikaCep (changes requested):
LGTM! Can we add sorting though?

VikaCep (approved):

Issue #525, opened 2023-03-12
SynMon app configuration expects a scheme for the API URL


Tristan971:
**What happened**:

Trying to add the SynMon app provisioning, I get an error when clicking "Initialize the plugin":
```
[...] msg="Proxy request failed" err="unsupported protocol scheme \"synthetic-monitoring-grpc-ap-southeast-0.grafana.net\""
```

Just in case it was a parsing issue and ignored, I also tried adding https:// in front, but it does (expectedly) fail when trying that:
```
msg="Proxy request failed" err="net/http: HTTP/1.x transport connection broken: malformed HTTP response \"\\x00\\x00\\x06\\x04\\x00\\x00\\x00\\x00\\x00\\x00\\x05\\x00\\x00@\\x00\""
```

**What you expected to happen**:

I expect the provisioning configuration to be usable on a selfhosted instance of Grafana

**How to reproduce it (as minimally and precisely as possible)**:

Stand up a Grafana instance with latest Grafana image at the time of writing (v9.4.3), install the synmon plugin, then try to provision it with the following (assuming the 2 datasources were provisionned correctly too, which is the case for me):

```
apiVersion: 1
apps:
  - type: grafana-synthetic-monitoring-app
    name: grafana-synthetic-monitoring-app
    disabled: false
    jsonData:
      apiHost: synthetic-monitoring-grpc-ap-southeast-0.grafana.net:443
      stackId: $redacted
      logs:
        grafanaName: grafanacloud-$redacted-logs
        hostedId: $redacted
      metrics:
        grafanaName: grafanacloud-$redacted-prom
        hostedId: $redacted
    secureJsonData:
      publisherToken: $redacted
```

**Anything else we need to know?**:

**Environment**:
- Grafana version (if running on prem): 9.4.3
- SM plugin version: 1.9.35
- OS Grafana is installed on: official docker image, looks like Alpine 3.15.7
- User OS & Browser: Windows and Edge 110
- Others: N/A

Tristan971:
Fwiw the solution is to use `https://synthetic-monitoring-api-ap-southeast-0.grafana.net` instead.

Which goes against the documentation here:

https://github.com/grafana/synthetic-monitoring-api-go-client/blob/58afdc8087623db5438bd7caecf55833a3350bc4/docs/API.md?plain=1#L7-L11

[...]

Issue #519, opened 2023-02-08
Extend the timeout from max 10 sec to 30 sec


yoramshai:
**What would you like to be added**:
Currently, the max timeout for a check is 10-sec max which results in HTTP response status equal to 0 and a latency of 10sec. It would be great if you could extend it to 30sec like in other checks platforms.

  **Why is this needed**:
When it happens, it hides the numbers we are looking for with the high latency. 
In the end, it is not only to measure up/down but also to make latency improvements, which is challenging at this moment.  
Note that when it happened, the latency was around ±11-15 sec.
In addition, it influences the reachability percentage, as the timeout counts as "down" even if the request is completed after 12 seconds.

**Additional context**
<img width="1023" alt="image" src="https://user-images.githubusercontent.com/11707803/217471617-a0a24ad7-fcb3-4f79-a093-0b409947affa.png">

peterschretlen:
This is addressed in https://github.com/grafana/synthetic-monitoring-app/pull/731
<img width="666" alt="image" src="https://github.com/grafana/synthetic-monitoring-app/assets/7903188/1101b2c3-effd-4358-baab-531efc144107">

yoramshai:
Thank you so much !

PR #1639, opened 2026-04-01
feat: folder assignment in check forms

VikaCep:
## Folder assignment in check forms

Part of [#519](https://github.com/grafana/synthetic-monitoring/issues/519) — M1: Core Folder Assignment
Closes: [#520](https://github.com/grafana/synthetic-monitoring/issues/520)

### What

Add folder assignment to all check creation and edit forms, allowing users to organize checks into Grafana folders. Gated behind `FeatureName.Folders`.

### Changes

- Add `folderUid` to the Check data model, form schema, and payload transformations
- Build React Query hooks for the Grafana Folders API with recursive fetch for nested folders
- Add a `FolderSelector` component with nested path display, loading state, and inline folder creation
- Integrate the folder field in all check form layouts (including Browser)
- **Per consensus (D3), all checks belong to a folder.** New checks and checks without a `folderUid` default to the existing "Grafana Synthetic Monitoring" root folder. The folder is resolved dynamically (UID lookup, title fallback, or created if absent)
- Add test infrastructure: MSW handlers, fixtures, and integration tests for feature flag gating, form rendering, and payload submission

### How to test

1. Enable the feature flag: add `?features=synthetic-monitoring-folders` to the URL
2. Create a new check — the folder selector should show "Grafana Synthetic Monitoring" as default
3. Edit an existing check — the folder selector should show the check's current folder (if any)
4. Create a subfolder via the "Create folder" button
5. Disable the feature flag — the folder field should not appear

image
image
image
image

### Note

- The SM backend does not yet support `folderUid` ([#521](https://github.com/grafana/synthetic-monitoring/issues/521)). The field is sent in the payload but will be ignored until the backend is updated. The frontend is fully functional with mocks.

- Deleted or missing folders currently display as `<uid> (folder not found)` in the selector. Full orphan detection and automatic reassignment to the default folder will be handled in [#526](https://github.com/grafana/synthetic-monitoring/issues/526).
image

### What's next

Following PRs in the [Folders delivery plan](https://github.com/grafana/synthetic-monitoring/issues/519):

- **M2**: Folder badge and folder view in check list ([#522](https://github.com/grafana/synthetic-monitoring/issues/522)), folder filtering ([#523](https://github.com/grafana/synthetic-monitoring/issues/523))
- **M3**: Combined permission model + permission-aware UI ([#524](https://github.com/grafana/synthetic-monitoring/issues/524))
- **M4**: Bulk move to folder ([#525](https://github.com/grafana/synthetic-monitoring/issues/525)), orphan detection ([#526](https://github.com/grafana/synthetic-monitoring/issues/526))


src/types.ts
```diff
@@ -420,6 +420,19 @@ export interface CheckBase {
...
+export interface GrafanaFolder {
+  uid: string;
+  title: string;
+  url: string;
+  parentUid?: string;
+  canEdit?: boolean;
+  canDelete?: boolean;
+  canAdmin?: boolean;
+  canSave?: boolean;
+  parents?: GrafanaFolder[];
```
g3john: This field is unused, should we remove it?

g3john (dismissed):
LGTM!

src/components/FolderSelector/FolderSelector.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const [showCreateModal, setShowCreateModal] = useState(false);
+
+  const isLoading = isDefaultLoading || isChildrenLoading;
+  const isError = isDefaultError || isChildrenError;
+
+  useEffect(() => {
+    if (value === undefined && defaultFolderUid) {
+      onChange(defaultFolderUid);
+    }
+  }, [value, defaultFolderUid, onChange]);
```
ckbedwell: Can we change this so it gets set in the `ChecksterProvider` in the defaultValues for the form so this becomes fully controlled. We would block rendering of the form until the API call but I don't think that's a big deal.

I know lifting it is a pain when it is behind a feature flag but I think it is worthwhile because I see this as tech debt but agents don't recognise this as that so will think this is an acceptable pattern that they can reuse elsewhere so we will accumulate tech debt elsewhere.

src/data/folders.constants.ts
```diff
@@ -0,0 +1,4 @@
+export const DEFAULT_FOLDER_TITLE = 'Grafana Synthetic Monitoring';
+export const DEFAULT_FOLDER_UID = 'grafana-synthetic-monitoring-app';
+export const FOLDERS_STALE_TIME = 5 * 60 * 1000;
+export const MAX_FOLDER_DEPTH = 8;
```
ckbedwell: Out of interest, why eight? (Why have a depth at all -- you'd have to be insane to have eight levels of nesting, though...).

src/data/folders.constants.ts
```diff
@@ -0,0 +1,4 @@
+export const DEFAULT_FOLDER_TITLE = 'Grafana Synthetic Monitoring';
```
ckbedwell, line 1: Booo 😂

src/data/folders.constants.ts
```diff
@@ -0,0 +1,4 @@
...
+export const DEFAULT_FOLDER_UID = 'grafana-synthetic-monitoring-app';
+export const FOLDERS_STALE_TIME = 5 * 60 * 1000;
```
ckbedwell: This seems too long to me. Why five minutes?

src/page/CheckList/components/CheckListFolderView.tsx
```diff
@@ -0,0 +1,385 @@
...
+        tabIndex={0}
+        aria-expanded={isExpanded}
+        aria-label={`${isExpanded ? 'Collapse' : 'Expand'} folder ${node.folder?.title ?? node.folderUid}`}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter' || e.key === ' ') {
+            e.preventDefault();
+            toggleFolder(node.folderUid);
+          }
+        }}
+      >
```
ckbedwell: Why isn't this a button?

src/page/CheckList/components/CheckListItemRow.tsx
```diff
@@ -22,6 +23,8 @@ export const CheckListItemRow = ({
   onStatusSelect,
   selected,
   onToggleCheckbox,
+  foldersMap,
+  foldersLoading,
```
ckbedwell: A lot of prop drilling going on here. Could be worth us thinking about a folders context?

ckbedwell (changes requested):
Super exciting and so glad to see this come to life! I've got a fair few comments on the general experience, though:

## Checks page

1. We have to account for the unhappy path if the folders API call fails. Right now it is quite alarming 😅 

Image

We should make it clearer than it failed to do the folder look-up and display a retry button. We don't ever want to render folder deleted when we can't confirm that.

2. I know I lost this fight about the default folder but this is so confusing and needs fixing.

Image

They aren't unassigned? They are in the default folder? 

Or are they?

Image

On the left I've assigned no folder but on the right I assigned them to the default but they are both in unassigned in the UI? 

This gets especially confusing when I've renamed the default folder and it still says unassigned.

Image

3. Speaking of which on the above -- why is the default folder special? I notice when I make new folders they render in a nested view but not for the default folder? I find this really confusing.

Image

4. The folder view should be the first choice and the default. I find it frustrating having to click on it every time I go to the checks page. Honestly, I'd take it a step further and not even have it as an option to opt out of -- if you are using folders, you are using folders.

As a user I want the list view (which really I think of as the 'condensed' view) to also be organised via folders.

5. I find the rendering of the folder label confusing in this view. I think if we just force the folders view we can drop this so we don't have to worry about having an inconsistent card experience?

Image

6. The focus styles of the card aren't the same as all the other focus styles? There could also be a bigger gap between the check card and the nested parent folder as there isn't one currently. 

Image

7. Should we just kill the pagination? This is how my page looks when I have a nested folder with 16 checks in it inside a parent folder with 17 checks in it.

Image

In the least we need to improve the nested folder view so it is easier to trace what we are looking at. If we do keep the pagination we could duplicate the controls so they appear at the top and bottom of the folder?

8. This label linking through to the actual folder (in dashboards... 🤦) added zero to the experience for me as a user. I found it more confusing because I was wondering where all my checks were... I think we should make folder view mandatory and then just kill this label as I mentioned above.

Image

On the above note -- maybe not for this PR but we should give some thought to folder management. As a user my expectation is that now I've embarked on organising things via folders I should very easily be able to rename / edit and move them around.

9. I know there is a bulk edit PR so maybe this is feedback for that but the global bulk edit button floats awkwardly in the UI besides the view selector with a very unclear purpose. Each folder should have one akin to the folder management in the dashboard view. There might even be room to argue we can remove the global one? This will need more thought.

## Checkster (_we need to rename this..._)

1. When I clicked '+ Create folder' and submitted my new folder in the checks edit view it submitted the form and took me to the checks page but I wasn't finished editing my check.

If my check had errors (or was a new check) it made the folder then immediately tried to submit the check.

My expectation here is that the folder wouldn't be made until I submitted the check. We'd have to have a similar waterfall to the alerts API.

2.  When making a new folder I couldn't nest it in an existing folder. I had to go to dashboards to move it.

3. The spacing in this modal is off and could be a bit more breathable.
Image

markjmeier:
Cherry picking some things to comment on:

>I know I lost this fight about the default folder but this is so confusing and needs fixing.

Looking at the knock on comments @ckbedwell left, here are some thoughts:

1. Unassigned is a bad name, at least make it "default" for clarity
2. There's a question if users should be able to rename this folder, I lean no - because if you let people rename you now need some UI to signal its the default folder and youll eventually get requests to "change the default folder". 

So i think we should basically be very opinionated on the name of the default folder, not let users change the name from whatever that is, and use the name where appropriate in the UI - i think this solves most of what chris commented on in regards to this.

I plan to hop on the branch and take a deeper look, but wanted to share the above.

ckbedwell:
> So i think we should basically be very opinionated on the name of the default folder, not let users change the name from whatever that is, and use the name where appropriate in the UI - i think this solves most of what chris commented on in regards to this.

I don't think it does. I'll reiterate I don't understand why there is a default folder seeing we don't have a default folder as of today.

This has just prompted me to check and I've just noticed that I can't put checks in any pre-existing folder I like. They **have** to be a child of the default folder (which I've since renamed in my stack -- which I like that I can do).

I've just checked and that wasn't part of the design document, is there a reason all folders have to be a child of the default one?

VikaCep:
> I've just checked and that wasn't part of the design document, is there a reason all folders have to be a child of the default one?

no, in fact it would be simpler to just place newly created folders **not** nested under the default one by default. I interpreted our decision from the design document as that everything needed to be under the default folder. But happy to modify this assumption (and implementation) if this is not accurate.

VikaCep:
## Checks page

> We have to account for the unhappy path if the folders API call fails. Right now it is quite alarming 😅

I've changed it to:

- folders can't be fetched, message indicating so and button to retry:
image

- folder actually deleted:
image

----------------------------------------------------------------------------------------

> I know I lost this fight about the default folder but this is so confusing and needs fixing.
> On the left I've assigned no folder but on the right I assigned them to the default but they are both in unassigned in the UI?
> This gets especially confusing when I've renamed the default folder and it still says unassigned.
> Speaking of which on the above -- why is the default folder special? I notice when I make new folders they render in a nested view but not for the default folder? I find this really confusing.

The "Unassigned" section is now gone. It now shows the actual default folder name with a "(default)" suffix, rendered as a collapsible folder node just like custom folders. If you've renamed the default folder, the real name shows.
image

----------------------------------------------------------------------------------------

> The folder view should be the first choice and the default. I find it frustrating having to click on it every time I go to the checks page. Honestly, I'd take it a step further and not even have it as an option to opt out of -- if you are using folders, you are using folders.

I had done this in a separate PR that have just merged into this branch: https://github.com/grafana/synthetic-monitoring-app/pull/1658

----------------------------------------------------------------------------------------
> I find the rendering of the folder label confusing in this view. I think if we just force the folders view we can drop this so we don't have to worry about having an inconsistent card experience?
> This label linking through to the actual folder (in dashboards... 🤦) added zero to the experience for me as a user. I found it more confusing because I was wondering where all my checks were... I think we should make folder view mandatory and then just kill this label as I mentioned above.

I've removed the folder badge (and the link with it) from card/list views. For folder management (rename, move, etc.), we lean on Grafana's existing folder management in Dashboards rather than replicating CRUD in SM. Happy to discuss in-app folder management down the line as a follow-up.

----------------------------------------------------------------------------------------

> The focus styles of the card aren't the same as all the other focus styles? There could also be a bigger gap between the check card and the nested parent folder.
> image

Fixed. After:
https://github.com/user-attachments/assets/c4f6d9e5-625f-4081-8be4-ae786f817f2e

----------------------------------------------------------------------------------------

> In the least we need to improve the nested folder view so it is easier to trace what we are looking at. If we do keep the pagination we could duplicate the controls so they appear at the top and bottom of the folder?

I've kept the pagination but duplicated the controls at the top and bottom of the folder. Also added a left border guide line on nested folder content to make it easier to trace which checks belong to which folder:

image

----------------------------------------------------------------------------------------

> I know there is a bulk edit PR so maybe this is feedback for that but the global bulk edit button floats awkwardly in the UI besides the view selector with a very unclear purpose. Each folder should have one akin to the folder management in the dashboard view. There might even be room to argue we can remove the global one? This will need more thought.

Addressed here: https://github.com/grafana/synthetic-monitoring-app/pull/1657#issuecomment-4299322332

VikaCep:
## Check form

> When I clicked '+ Create folder' and submitted my new folder in the checks edit view it submitted the form and took me to the checks page but I wasn't finished editing my check.
> If my check had errors (or was a new check) it made the folder then immediately tried to submit the check.
> My expectation here is that the folder wouldn't be made until I submitted the check. We'd have to have a similar waterfall to the alerts API.

I fixed the issue where creating a folder also submitted the check form. The folder is now created independently and the form stays open for the user to continue editing. I considered the waterfall approach (defer creation until check submit) but it adds significant complexity for what's a minor edge case (user creates a folder then cancels the check edit). Happy to revisit if this becomes a real user pain point/as a follow-up though.

https://github.com/user-attachments/assets/e78e26e0-f46d-496d-b46f-bda0cbcb4a21

> When making a new folder I couldn't nest it in an existing folder. I had to go to dashboards to move it.

Added a parent folder selector so we can choose where to place it.


https://github.com/user-attachments/assets/77a15135-5439-413d-8e4c-e5f86ee61250


> The spacing in this modal is off and could be a bit more breathable.

The modal uses standard `@grafana/ui` `Field` and `ButtonRow` components -- the spacing comes from their defaults. 
image

src/components/FolderSelector/FolderSelector.tsx
VikaCep: I've moved the resolution of the default folder to `ChecksterProvider` via `useDefaultFolder` and set in the form's `defaultValues` before it renders. Removed hte `useEffect` in `FolderSelector`.

src/data/folders.constants.ts
VikaCep: Reduced to 30 seconds.

src/page/CheckList/components/CheckListFolderView.tsx
VikaCep: Changed it to button 👍

src/data/folders.constants.ts
VikaCep: Removed the depth limit. It was a recursion guard, but I've addressed it differently using a `visited` set that tracks `UIDs` already fetched.

src/page/CheckList/components/CheckListItemRow.tsx
VikaCep: After removing the folder badge from card/list views, the folder props only pass one level deep now, from `CheckList` to `CheckListFolderView`. The previous multi-level drilling is gone, so a context for one level feels a bit too much perhaps.

VikaCep:
@ckbedwell Thanks for your thorough review! I've addressed all the comments. I think the outstanding point that we need to decide is whether to nest new folders under the Grafana default one or not.

src/components/FolderSelector/FolderSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+  };
+
+  const selectedValue = value ?? defaultFolderUid ?? null;
+
+  if (isLoading) {
+    return <LoadingPlaceholder text="Loading folders..." />;
+  }
+
+  if (isError) {
+    return <Alert title="Unable to load folders" severity="warning" />;
```
ckbedwell: I know I'm a stickler for the unhappy paths but can we put a retry button in here? It's not obvious as a user what to do next, especially if wanting to use the folders feature. I also noticed that it refetches on window focus which I'm iffy on -- I know that is the default for react-query but it always makes me jittery when I see the UI shift when I refocus a tab.

Image

ckbedwell (changes requested):
Image

I was testing on ops and realised I couldn't make folders. We might have to give some more thought about RBAC and folders control here. I just have the default permissions on ops, so you should be able to repro easily with Graft.

I don't know what level of RBAC folders has, so worth exploring and just making sure a few different situations are covered -- in particular I think you can get quite granular with having write permissions for some folders but not others?

Image

My expectation is that I could make a folder in the root here. If I went to the folders UI I could make a folder in root then it would appear here so that makes sense to me.

Really close now! Ping me when you've made the changes and I'll rereview ASAP!

src/components/FolderSelector/FolderSelector.tsx
VikaCep: Done! 
image

VikaCep:
> I was testing on ops and realised I couldn't make folders. We might have to give some more thought about RBAC and folders control here. I just have the default permissions on ops, so you should be able to repro easily with Graft.

So the thing is we don't have write permissions in ops for folders. I've hidden the button to create a folder when this is the case. 

image

-------

> I don't know what level of RBAC folders has, so worth exploring and just making sure a few different situations are covered -- in particular I think you can get quite granular with having write permissions for some folders but not others?

Regarding this, this other PR that branches off this one, focuses on the granular RBAC permissions https://github.com/grafana/synthetic-monitoring-app/pull/1653

ckbedwell (dismissed):
LGTM! AMAZING work, super excited to try this out!

Let's get this in then can you resolve the undoubted merge conflicts from the downstream PRs and I'll work my way through them tomorrow morning?

PR #1644, opened 2026-04-03
fix: prevent text overlap in check list view on narrow screens

seojcarlos:
## Summary

Fixes #1638 — Column text overlaps with adjacent columns in check list view when screen width is below ~1550px.

## Changes

**`CheckListItemRow.tsx`**:
- Adjusted grid column min widths: job name column from `220px` to `150px`, target column from `1px` to `100px`. This gives columns better space distribution on narrow viewports
- Added `overflow: hidden` and `minWidth: 0` to the target column container to enable proper text truncation within CSS Grid

**`CheckListItemDetails.tsx`**:
- Added `whiteSpace: nowrap` to detail items to prevent awkward line breaks in frequency/series/location text

## How it works

The root cause was that `minmax(1px, 1fr)` allowed the target column to shrink to almost nothing, while `auto` columns for details could grow unconstrained. The fix gives more balanced minimum widths and ensures text truncation via `overflow: hidden` + `minWidth: 0` (required for CSS Grid children to respect truncation).

### DIFF
--- a/src/page/CheckList/components/CheckListItemDetails.tsx
+++ b/src/page/CheckList/components/CheckListItemDetails.tsx
@@ -116,6 +116,7 @@ const getStyles = (theme: GrafanaTheme2) => ({
   detailItem: css({
     minWidth: 0,
     fontVariantNumeric: 'tabular-nums',
+    whiteSpace: 'nowrap',
   }),
   wrapDetailItem: css({
     backgroundColor: theme.colors.background.primary,
--- a/src/page/CheckList/components/CheckListItemRow.tsx
+++ b/src/page/CheckList/components/CheckListItemRow.tsx
@@ -88,7 +88,7 @@ const getStyles = (theme: GrafanaTheme2) => ({
   }),
   listCardWrapper: css({
     display: 'grid',
-    gridTemplateColumns: 'auto minmax(220px, 1.25fr) minmax(1px, 1fr) auto auto auto',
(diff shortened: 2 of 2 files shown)


seojcarlos:
@g3john Sure! Here are the visual references:

### Before (bug)
The screenshot from issue #1638 shows the overlap clearly:

Before — text overlap

At widths below ~1550px, the target column (`minmax(1px, 1fr)`) shrinks to nearly zero, causing text to overlap with adjacent `auto` columns.

### After (fix)
I don't have access to a running Grafana Synthetic Monitoring instance to capture a live screenshot, but I created an **interactive CSS Grid demo** that reproduces the exact grid layout before and after the fix:

🔗 **[Live demo (HTML)](https://gist.github.com/seojcarlos/3dc2ddb5cd463382ba74fc56bfdb6f62)** — open the raw HTML file to see both states

**What the fix does:**
- `minmax(220px, 1.25fr)` → `minmax(150px, 1.25fr)` — job name column allows more room for others
- `minmax(1px, 1fr)` → `minmax(100px, 1fr)` — target column no longer collapses to zero
- Added `overflow: hidden` + `minWidth: 0` on `.checkTarget` — enables proper text truncation in CSS Grid
- Added `whiteSpace: nowrap` on `.detailItem` — prevents awkward line breaks in frequency/locations text

The result is that all columns maintain readable minimum widths and long URLs truncate with ellipsis instead of overlapping adjacent columns.

Happy to set up the dev environment and provide live screenshots if needed!

ckbedwell (approved):

PR #1646, opened 2026-04-07
feat: create SLO from check

markjmeier:

This PR allows a user to create SLOs directly from a synthetic monitoring check.  They can create an SLO from a single check OR they can create SLOs based on matching labels.  We should be aware of the ongoing work to improve labels and considering holding this PR until that work is complete, since it will change some of the queries we would use here.

Upon creating a SLO a user can view the SLO directly in the SLO app.

image

See sm-dev slack for a full video demo!

Fixes #

If we do decide to move quickly on this, we should consider making it experimental to gather initial feedback. e.g. on the type of SLO query a customer would prefer


VikaCep:
@ckbedwell as discussed with @markjmeier, the current labels work may impact this, so we should factor that in when deciding when to release this feature.

markjmeier:
Made a few updates based on feedback from Pepe and Deyan from the SLO team:

- Improved links after SLO creation, users can go directly to the dashboard or edit the SLO, makes it easier to edit/configuure the SLO they just created
- gave instructions that they may want to set up burn-rate alerts, labels, and folders in the SLO app
- dropped SLO prefix to name
- improved name to be {job} reachability
- hide label group SLO for implementation at a later date

The idea of uptime based SLO is interesting as customers may want that - i didnt dig into it, but we should consider presenting that option in the near future

VikaCep:
This is looking great! 
image

I pushed some small changes on removing dead code/small refactor but other than that I think it looks much better.

VikaCep:
@markjmeier I wonder if we should show the link to the SLO dashboard on the check's dashboard itself (when it exists) and not only in the creation modal, as when it's closed all references to it are lost from SM's point of view.
image

Not sure if it's worth it though, as I believe we need to fetch all SLOs from their API and filter them based on the check id on our side to do that. Maybe something to revisit if we see adoption pick up.

markjmeier:
@VikaCep That does sound nice, but as you say we either ned to fetch all SLOs, or store the id somewhere so we can keep track of it.

Agree that we can skip and revisit if we see adoption pick up - im sure we will get that feedback eventually. "where did it go?". I think we should also pull SLO metrics back into a check dashboard if one is set up, but now we are talking about a deeper/more binding relationship between checks and SLOs

VikaCep:
Should we close this one in favor of https://github.com/grafana/synthetic-monitoring-app/pull/1669 ?

markjmeier:
@VikaCep Yes, i'll close it out.

 I believe there is some validity to the ease of use/just works in the method i provide. But i know @ckbedwell is working to improve the exposed component from the SLO app and its probably better to use their component for various reasons.

PR #1653, opened 2026-04-10
feat: combined folder permission model for checks

VikaCep:
# feat: combined folder permission model for checks

Addresses: https://github.com/grafana/synthetic-monitoring/issues/524

## Problem

When checks are assigned to Grafana folders, users can see and interact with all checks regardless of their folder-level permissions. A user with View-only access to a folder can still edit and delete checks in that folder. There is no enforcement of the Grafana folder permission model on check operations, which breaks the expected access control behavior established by other Grafana Cloud plugins like SLOs.

## Solution

Implements the combined permission model (D5): a user's effective permission on a check is the minimum of their SM RBAC role and their Grafana folder permission level. This follows the SLO model used across Grafana Cloud.

The folder permission mapping:
- **Folder View** (canEdit=false) → check read only
- **Folder Edit** (canEdit=true) → check read + write
- **Folder Admin** (canAdmin=true) → check read + write + delete

All changes are gated behind the `FeatureName.Folders` feature flag. When disabled, behavior is unchanged — SM RBAC applies directly.

### Check list
- Checks in forbidden folders (403) are hidden immediately
- Checks in orphaned folders (404, deleted) appear after background resolution
- Action buttons (edit, delete, enable/disable) respect per-check effective permissions
- Bulk actions are disabled if any selected check lacks the required permission

### Edit page
- Form is disabled when effective permissions don't allow write
- Redirects if the check's folder is forbidden

### Dashboard page
- Redirects if the check's folder is forbidden

### Architecture
- Pure decision functions in `folderPermissions.ts` (independently testable, no React dependencies)
- `useFolderPermissions` hook batch-fetches `GET /api/folders/:uid` (the only endpoint returning canEdit/canAdmin)
- `useCheckFolderAccess` hook encapsulates the full flow (accessible set filtering, background resolution, permission computation)
- `CheckFolderAccessContext` eliminates prop drilling — descendants consume permissions via hooks

### API constraint
The `GET /api/folders` list endpoint does not return permission fields. Only `GET /api/folders/:uid` returns `canEdit` and `canAdmin`. The hook fetches permissions progressively per folder, with buttons disabled during loading.

### Orphaned checks (deleted folders) are hidden for non-Admin users

When a folder is deleted, checks assigned to it become orphaned. The Grafana folder API (`GET /api/folders/:uid`) returns different responses depending on the user's org role:

- **Admin org role**: receives **404** (folder not found) → our code shows the check with a "Folder deleted" badge
- **Editor/Viewer org role**: receives **403** (access denied) → our code cannot distinguish "deleted folder" from "no access", so it hides the check as a safe default

This is a Grafana security behavior — it does not reveal folder existence to users without broad folder permissions. The result is that orphaned checks are only visible to Admin users until they are reassigned.

**M4 will address this** with backend-side orphan detection and auto-move: when the SM backend detects a check references a folder that no longer exists, it automatically moves the check to the default "Synthetic Monitoring" folder, making it visible to all users again.

### Folder permissions cannot be restricted below parent level

Grafana's folder permission model is additive — child folders inherit permissions from parents and can only escalate, not restrict. To test View-only access on a specific folder, the parent folder's Viewer/Editor role permissions must be removed first, then explicit per-user/team permissions set on each child folder.

### DIFF
--- a/src/data/folderPermissions.test.ts
+++ b/src/data/folderPermissions.test.ts
@@ -3,7 +3,7 @@ import { runTestAsRBACEditor, runTestAsRBACReader } from 'test/utils';
 
 import {
   CheckFolderStatus,
-  computeEffectiveCheckPermissions,
+    computeCheckPermissions,
   FolderAccessState,
   isCheckVisible,
   resolveCheckFolderStatus,
@@ -98,12 +98,12 @@ describe('isCheckVisible', () => {
   });
 });
 
-describe('computeEffectiveCheckPermissions', () => {
+describe('computeCheckPermissions', () => {
   describe('no-folder-context (folders disabled or no folderUid)', () => {
     const status: CheckFolderStatus = { type: 'no-folder-context' };
 
     it('returns full SM RBAC for a writer', () => {
-      expect(computeEffectiveCheckPermissions(smWriter(), status)).toEqual({
+      expect(computeCheckPermissions(smWriter(), status)).toEqual({
         canRead: true,
         canWrite: true,
         canDelete: true,
@@ -111,7 +111,7 @@ describe('computeEffectiveCheckPermissions', () => {
     });
 
     it('returns read-only for a reader', () => {
-      expect(computeEffectiveCheckPermissions(smReader(), status)).toEqual({
+      expect(computeCheckPermissions(smReader(), status)).toEqual({
         canRead: true,
         canWrite: false,
         canDelete: false,
@@ -123,7 +123,7 @@ describe('computeEffectiveCheckPermissions', () => {
     const status: CheckFolderStatus = { type: 'orphaned' };
 
     it('returns full SM RBAC — folder no longer restricts', () => {
-      expect(computeEffectiveCheckPermissions(smWriter(), status)).toEqual({
+      expect(computeCheckPermissions(smWriter(), status)).toEqual({
         canRead: true,
         canWrite: true,
         canDelete: true,
@@ -131,7 +131,7 @@ describe('computeEffectiveCheckPermissions', () => {
     });
 
     it('returns read-only for a reader', () => {
-      expect(computeEffectiveCheckPermissions(smReader(), status)).toEqual({
+      expect(computeCheckPermissions(smReader(), status)).toEqual({
         canRead: true,
         canWrite: false,
         canDelete: false,
@@ -141,7 +141,7 @@ describe('computeEffectiveCheckPermissions', () => {
 
   describe('loading (permissions not yet resolved)', () => {
     it('allows read but disables write and delete', () => {
-      expect(computeEffectiveCheckPermissions(smWriter(), FOLDER_LOADING)).toEqual({
+      expect(computeCheckPermissions(smWriter(), FOLDER_LOADING)).toEqual({
         canRead: true,
         canWrite: false,
         canDelete: false,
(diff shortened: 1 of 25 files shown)


VikaCep:
## Manual validation — combined folder permission model

### Setup

Tested against a Grafana Cloud instance with two users:
- **vikaAdmin user** — org Admin
- **vikaEditor user** —  org Editor, also member of `Team Editor`

Folder permission hierarchy configured by restricting the default `Grafana Synthetic Monitoring` folder so child folders can impose tighter limits:

Grafana Synthetic Monitoring default folder:
image

Admin Only folder (child of default):
image

Team Edit Folder (child of default):
image

Computed permissions:
| Folder | **vikaEditor**'s effective permission | Expected behavior |
|---|---|---|
| Admin Only Folder | View (no team grant) | Read only — edit/delete/enable/disable disabled |
| Team Edit Folder | Edit (via Team Editor) | Read + write — edit/enable/disable enabled, delete disabled |
| Unassigned (no folder) | SM RBAC only (Editor) | Full access — folder permissions not applied |

### Results as **vikaEditor** user

**Restricted folder — action buttons disabled:**
image

Checks in `Admin Only Folder` have edit, delete, and enable/disable buttons disabled. The user can only view dashboard.

**Permitted folder — action buttons enabled:**

image

Checks in `Team Edit Folder` have edit and enable/disable enabled. Delete is disabled (user has Edit, not Admin on this folder).

**Unassigned checks — SM RBAC applies directly:**

image

Checks without a `folderUid` bypass folder permissions entirely. The Editor user retains full SM RBAC access. This is by design — folder permissions are opt-in when a check is assigned to a folder.

### Summary

| Scenario | Expected | Actual |
|---|---|---|
| Checks in forbidden folder (no access) | Hidden from list | :white_check_mark: |
| Checks in View-only folder | Visible, all actions disabled | :white_check_mark: |
| Checks in Edit folder (no Admin) | Edit/toggle enabled, delete disabled | :white_check_mark: |
| Checks without folder | Full SM RBAC access | :white_check_mark: |
| Feature flag off | All checks visible, SM RBAC only | :white_check_mark: (covered by tests) |

g3john (dismissed):
LGTM!

src/data/folderPermissions.test.ts
```diff
@@ -0,0 +1,225 @@
...
+  const runtime = require('@grafana/runtime');
+  jest.replaceProperty(runtime, 'config', {
+    ...runtime.config,
+    bootData: {
+      ...runtime.config.bootData,
+      user: { permissions: { ...FULL_WRITER_ACCESS, 'grafana-synthetic-monitoring-app.checks:delete': false } },
+    },
+  });
+  return getUserPermissions();
+};
```
ckbedwell, line 25: Should we extract this to be a util like the others?

src/hooks/useCheckFolderAccess.ts
```diff
@@ -0,0 +1,91 @@
...
+    checks.forEach((check) => {
+      if (check.folderUid) {
+        uids.add(check.folderUid);
+      }
+    });
+    return [...uids];
+  }, [checks, isFoldersEnabled, defaultFolderUid]);
+
+  const { folderDetailsByUid } = useFolderPermissions(folderUids);
+  const smPerms = getUserPermissions();
```
ckbedwell: This is fairly minor but I realised whilst I was playing with this that this creates a new object on every render so it makes some of the useCallbacks / useMemos pointless. We should probably make a `useUserPermissions` hook that wraps `getUserPermissions` and memoises it.

ckbedwell (dismissed):
Two minor comments but otherwise LGTM!

PR #1656, opened 2026-04-15
fix: preserve adhoc test panel state during probe refreshes

ckbedwell:
## Problem

The checks editor's Test panel could collapse open log details while a test was still visible. This happened when probe refreshes or new log entries caused the panel state to be rebuilt, which interrupted the debugging flow and made it harder to inspect test output.

## Solution

This change preserves adhoc test panel state across refreshes by avoiding duplicate state initialization for the same adhoc request and by using stable keys for rendered log entries.

It also updates the shared test render helper to return the mounted React Query client so tests can trigger refetches against the same client instance the UI is using. The adhoc panel regression coverage now uses the codebase's normal MSW-backed request flow instead of hook-level mocks.


g3john (approved):
LGTM!

PR #1657, opened 2026-04-15
Feat: add folders bulk assignment

VikaCep:
## Feat: add folders bulk assignment

Addresses: https://github.com/grafana/synthetic-monitoring/issues/525

## Problem

When users have many checks across different folders, there is no way to reassign multiple checks to a different folder at once. Users must edit each check individually to change its folder, which is tedious at scale. Additionally, the folder selector in both the check form and the bulk move modal shows all visible folders regardless of the user's permission level, allowing users to select folders they only have View access on, where saving or moving would be invalid.

## Solution

Adds a "Move to folder" action to the existing bulk actions bar. When checks are selected, users can pick a target folder from a combobox and move them all at once via the existing bulk update API (`POST /sm/check/update/bulk`). The action is gated behind `FeatureName.Folders` and respects the M3 permission model. The button is disabled when the user lacks write permission on any of the selected checks (determined by the combined SM RBAC + folder permission model).

The folder selector now filters out folders where the user does not have Edit permission. This applies to all consumers of `FolderSelector`: both the bulk move modal and the check edit form benefit from the same protection.

https://github.com/user-attachments/assets/08b12063-c960-4221-b267-b8c871d5705b

https://github.com/user-attachments/assets/1f937d1c-d5e1-4cfa-83af-d95cdd447f1d

### DIFF
--- a/src/components/FolderSelector/FolderSelector.tsx
+++ b/src/components/FolderSelector/FolderSelector.tsx
@@ -16,8 +16,8 @@ interface FolderSelectorProps {
 }
 
 export function FolderSelector({ value, onChange, disabled, autoSelectDefault = true, 'aria-label': ariaLabel }: FolderSelectorProps) {
-  const { defaultFolder, defaultFolderUid, isLoading: isDefaultLoading, isError: isDefaultError } = useDefaultFolder();
-  const { data: childFolders = [], isLoading: isChildrenLoading, isError: isChildrenError } = useFolderChildren(defaultFolderUid);
+  const { defaultFolder, defaultFolderUid, isLoading: isDefaultLoading, isError: isDefaultError, refetch: refetchDefault } = useDefaultFolder();
+  const { data: childFolders = [], isLoading: isChildrenLoading, isError: isChildrenError, refetch: refetchChildren } = useFolderChildren(defaultFolderUid);
   const [showCreateModal, setShowCreateModal] = useState(false);
 
   const allFolders = useMemo(() => (defaultFolder ? [defaultFolder, ...childFolders] : []), [defaultFolder, childFolders]);
--- a/src/components/FolderSelector/FolderSelector.tsx
+++ b/src/components/FolderSelector/FolderSelector.tsx
@@ -1,27 +1,27 @@
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { Alert, Button, Combobox, ComboboxOption, Field, Input, LoadingPlaceholder, Modal, Stack } from '@grafana/ui';
 import { trackFolderCreated, trackFolderSelected } from 'features/tracking/folderEvents';
 
 import { GrafanaFolder } from 'types';
 import { useDefaultFolder } from 'data/useDefaultFolder';
+import { useFolderPermissions } from 'data/useFolderPermissions';
 import { getFolderPathParts, useCreateFolder, useFolderChildren } from 'data/useFolders';
 
 interface FolderSelectorProps {
   value?: string;
   onChange: (folderUid: string | undefined) => void;
   disabled?: boolean;
+  autoSelectDefault?: boolean;
   'aria-label'?: string;
 }
 
-export function FolderSelector({ value, onChange, disabled, 'aria-label': ariaLabel }: FolderSelectorProps) {
-  const { defaultFolder, defaultFolderUid, isLoading: isDefaultLoading, isError: isDefaultError, refetch: refetchDefault } = useDefaultFolder();
-  const {
-    data: childFolders = [],
-    isLoading: isChildrenLoading,
-    isError: isChildrenError,
-    refetch: refetchChildren,
-  } = useFolderChildren(defaultFolderUid);
+export function FolderSelector({ value, onChange, disabled, autoSelectDefault = true, 'aria-label': ariaLabel }: FolderSelectorProps) {
+  const { defaultFolder, defaultFolderUid, isLoading: isDefaultLoading, isError: isDefaultError } = useDefaultFolder();
+  const { data: childFolders = [], isLoading: isChildrenLoading, isError: isChildrenError } = useFolderChildren(defaultFolderUid);
   const [showCreateModal, setShowCreateModal] = useState(false);
+
+  const allFolders = useMemo(() => (defaultFolder ? [defaultFolder, ...childFolders] : []), [defaultFolder, childFolders]);
+  const { folderDetailsByUid } = useFolderPermissions(allFolders.map((f) => f.uid));
 
   const isLoading = isDefaultLoading || isChildrenLoading;
   const isError = isDefaultError || isChildrenError;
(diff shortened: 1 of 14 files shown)


src/components/FolderSelector/FolderSelector.tsx
```diff
@@ -4,71 +4,74 @@ import { Alert, Button, Combobox, ComboboxOption, Field, Input, LoadingPlacehold
...
-  const {
-    data: childFolders = [],
-    isLoading: isChildrenLoading,
-    isError: isChildrenError,
-  } = useFolderChildren(defaultFolderUid);
+  const { data: childFolders = [], isLoading: isChildrenLoading, isError: isChildrenError } = useFolderChildren(defaultFolderUid);
   const [showCreateModal, setShowCreateModal] = useState(false);
 
+  const allFolders = useMemo(() => (defaultFolder ? [defaultFolder, ...childFolders] : []), [defaultFolder, childFolders]);
+  const { folderDetailsByUid } = useFolderPermissions(allFolders.map((f) => f.uid));
```
g3john, line 24: Should we memoize the mapping here as well?
```suggestion
  const allFoldersUids = allFolders.map((f) => f.uid));
  const { folderDetailsByUid } = useFolderPermissions(allFoldersUids);
```

src/page/CheckList/components/BulkActions.tsx
```diff
@@ -118,6 +139,13 @@ export const BulkActions = ({ checks, onResolved }: BulkActionsProps) => {
           onDismiss={() => setShowDeleteModal(false)}
         />
       )}
+      {showMoveToFolderModal && (
+        <BulkMoveToFolderModal
+          checks={checks}
+          isOpen={showMoveToFolderModal}
+          onDismiss={handleMoveResolved}
```
g3john, line 146: Maybe for dismissing the modal, without actually moving, we don't call `handleMoveResolved` which unselects all the checks. That way users don't have to reselect the checks if the wanted to do a different bulk action

src/page/CheckList/components/BulkActions.tsx
```diff
@@ -118,6 +139,13 @@ export const BulkActions = ({ checks, onResolved }: BulkActionsProps) => {
           onDismiss={() => setShowDeleteModal(false)}
         />
       )}
+      {showMoveToFolderModal && (
```
g3john, line 142: I don't think we need to have this conditional render since we are already passing `showMoveToFolderModal` as `isOpen`

VikaCep:
Bringing a comment from @ckbedwell in https://github.com/grafana/synthetic-monitoring-app/pull/1639 here to address in this PR:

> I know there is a bulk edit PR so maybe this is feedback for that but the global bulk edit button floats awkwardly in the UI besides the view selector with a very unclear purpose. Each folder should have one akin to the folder management in the dashboard view. There might even be room to argue we can remove the global one? This will need more thought.

VikaCep:
> Bringing a comment from @ckbedwell in https://github.com/grafana/synthetic-monitoring-app/pull/1639 here to address in this PR:
> 
> I know there is a bulk edit PR so maybe this is feedback for that but the global bulk edit button floats awkwardly in the UI besides the view selector with a very unclear purpose. Each folder should have one akin to the folder management in the dashboard view. There might even be room to argue we can remove the global one? This will need more thought.


I've added a new section on each folders header that allows us to perform bulk actions at that level:
image

Recording:
https://github.com/user-attachments/assets/4c3a243b-3c82-4323-99fc-53a32159e08d

g3john:
> > Bringing a comment from @ckbedwell in #1639 here to address in this PR:
> > I know there is a bulk edit PR so maybe this is feedback for that but the global bulk edit button floats awkwardly in the UI besides the view selector with a very unclear purpose. Each folder should have one akin to the folder management in the dashboard view. There might even be room to argue we can remove the global one? This will need more thought.
> 
> I've added a new section on each folders header that allows us to perform bulk actions at that level: image
> 
> Recording: https://github.com/user-attachments/assets/4c3a243b-3c82-4323-99fc-53a32159e08d

Awesome! I feel like the checkbox for it is better suited to be left-aligned, similar to the other checkboxes for consistency, so users won't miss it. Also the bulk actions as well, WDYT?

ckbedwell:
> Awesome! I feel like the checkbox for it is better suited to be left-aligned, similar to the other checkboxes for consistency, so users won't miss it. Also the bulk actions as well, WDYT?

I agree with John on the consistency aspect here -- especially as in the folders management checkboxes for Grafana dashboards is left-aligned -- but I _do_ like that it uses the space on the right which feels wasted otherwise.

I think what you've done here is great. I have a POC to re-imagine the filtering experience on this page so maybe we just get this in and then I take a stab at re-doing the UX on this page with the new filters I've been experimenting with. (I'll put the POC up this afternoon)

src/page/CheckList/components/BulkActions.hooks.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export function useBulkActions({ checks, onResolved }: UseBulkActionsOptions) {
+  const isFoldersEnabled = isFeatureEnabled(FeatureName.Folders);
+  const { canWriteAll, canDeleteAll } = useBulkCheckPermissions(checks);
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
+  const [showMoveToFolderModal, setShowMoveToFolderModal] = useState(false);
+  const { mutate: bulkUpdateChecks } = useBulkUpdateChecks({ onSuccess: onResolved });
+
+  const handleDeleteResolved = () => {
```
ckbedwell: Nit: I just default to useCallback'ing everything nowadays. I read a [compelling argument](https://attardi.org/why-we-memo-all-the-things/) many years ago that convinced me for complicated and ever-growing code-bases the trade-off of premature optimisation via memoization vs going hunting for the performance problem later on is not worth it.

src/page/CheckList/components/FolderBulkActions.tsx
```diff
@@ -0,0 +1,94 @@
...
+      />
+      {showDeleteModal && (
+        <ConfirmModal
+          isOpen={showDeleteModal}
+          {...deleteModalProps}
+          onConfirm={deleteChecks}
+          onDismiss={() => setShowDeleteModal(false)}
+        />
+      )}
+      <BulkMoveToFolderModal
```
ckbedwell, line 78: Because this isn't conditionally rendered its state persists when it is closed and re-opened. Is that intentional?

ckbedwell (dismissed):
Few questions but otherwise LGTM! 👏 ⭐

src/page/CheckList/components/FolderBulkActions.tsx
VikaCep: Fixed!

VikaCep:
I moved the checkbox to select within the folder to the left for consistency. 
image

> I think what you've done here is great. I have a POC to re-imagine the filtering experience on this page so maybe we just get this in and then I take a stab at re-doing the UX on this page with the new filters I've been experimenting with. (I'll put the POC up this afternoon)

I agree we can tackle bigger changes if we re-design the filtering experience as a whole 👍

PR #1664, opened 2026-04-20
fix: include k6/secrets types when version management is enabled

ckbedwell:
## Problem

When version management is enabled, the script editor fetches `@types/k6` definitions from the Unpkg CDN and replaces the bundled `k6.types.ts` fallback. The CDN loader's module list was missing `k6/secrets`, so once dynamic types loaded, imports from `k6/secrets` in scripts would be reported as errors by Monaco even when the Secrets Management feature was enabled — losing autocomplete and type checking for that module.

## Solution

- Add `k6/secrets` to the CDN module list in `k6TypesCdnLoader` so it is fetched alongside the other `@types/k6` submodules.
- Filter `k6/secrets` out of the type registration step in `CodeEditor` when the `SecretsManagement` feature flag is disabled, so the types only surface in Monaco for tenants that can actually use them. This keeps the flag gate in one place regardless of whether types come from the CDN or the bundled fallback.


VikaCep (approved):
LGTM!

PR #1665, opened 2026-04-22
chore: use grafana secrets api

g3john:

Replaces SM's own secrets passthrough (`/api/v1alpha1/secrets` on the SM backend) with direct calls to Grafana's secrets management API at `apis/secret.grafana.app/v1beta1/namespaces/stacks-{stackId}/securevalues`.

The SM passthrough was built before the shared Grafana secrets backend was available. Now that the backend (owned by `#grafana-operator-experience-squad`) is in place, SM should talk to it directly like `grafana-k6-app` already does. This also fixes a defect in the current implementation where secrets with decrypters other than `synthetic-monitoring` were visible to SM even though SM cannot actually use them.

**Changes**

- New client module `src/data/clients/SecretsManagerClient/` that mirrors the [k6-app pattern](https://github.com/grafana/grafana-k6-app/blob/b2f5675fe75414a1ab4375c39a8fe4ac7a5b3d16/src/data/clients/SecretsManagerClient/SecretsManagerClient.ts):
  - `SecretsManagerClient` class with `fetchAll / fetchSecret / createSecret / updateSecret / deleteSecret`
  - Scopes list calls server-side via `fieldSelector=spec.decrypter=synthetic-monitoring`
  - Normalizes Kubernetes-style responses into the existing `SecretWithMetadata` shape so UI consumers don't need to change
  - `useSecretsManagerClient()` hook bound to `useMeta().jsonData.stackId`, returning `{ client, stackId }`
- `src/data/useSecrets.ts` rewired to use the new client; hook signatures (`useSecrets`, `useSecret`, `useSaveSecret`, `useDeleteSecret`) unchanged. `QUERY_KEYS` are now stack-scoped.
- Removed `getSecrets/getSecret/saveSecret/deleteSecret` from `SMDataSource` along with their imports.
- MSW handlers and fixtures updated to the new routes and k8s response shape.

**Behavior diff worth highlighting**

- **Before**: editing a secret with extra decrypters silently dropped them on save (SM's passthrough only knew about its own field set).
- **After**: editing echoes back the full decrypter list; dropping decrypters requires going through Administration > Secrets Management explicitly.

https://github.com/user-attachments/assets/01d24d37-8932-4148-95d4-97b896a70a8b

### DIFF
--- a/src/data/clients/SecretsManagerClient/constants.ts
+++ b/src/data/clients/SecretsManagerClient/constants.ts
@@ -1,0 +1,12 @@
+/**
+ * Decrypter identifier used by Synthetic Monitoring in Grafana's secrets
+ * management API. Secrets that include this decrypter are visible to SM
+ * and can be referenced by SM checks.
+ */
+export const SM_SECRET_DECRYPTER = 'synthetic-monitoring';
+
+/**
+ * Base path for the Grafana secrets management API.
+ * Full list URL: `${SECRETS_API_BASE}/namespaces/stacks-{stackId}/securevalues`.
+ */
+export const SECRETS_API_BASE = '/apis/secret.grafana.app/v1beta1';
--- a/src/data/useSecrets.ts
+++ b/src/data/useSecrets.ts
@@ -1,31 +1,28 @@
-import { useMutation, useQuery } from '@tanstack/react-query';
+import { type QueryKey, useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 
-import { SMDataSource } from 'datasource/DataSource';
-import { useSMDS } from 'hooks/useSMDS';
+import { useSecretsManagerClient } from 'data/clients/SecretsManagerClient';
+import { CreateSecretFormValues } from 'data/clients/SecretsManagerClient/SecretsManagerClient.utils';
 import { SecretWithMetadata } from 'page/ConfigPageLayout/tabs/SecretsManagementTab';
 import { SECRETS_EDIT_MODE_ADD } from 'page/ConfigPageLayout/tabs/SecretsManagementTab/constants';
 import { SecretFormValues } from 'page/ConfigPageLayout/tabs/SecretsManagementTab/SecretsManagementTab.utils';
 
-import { queryClient } from './queryClient';
-
-export interface SecretsResponse {
-  secrets: SecretWithMetadata[];
-}
-
+/**
+ * React-query keys for secrets. All keys are scoped by `stackId` so multiple
+ * stacks in the same session cannot share cache entries.
+ */
 export const QUERY_KEYS = {
-  list: ['secrets'],
-  byName: (name: string) => ['secrets', name],
+  list: (stackId: number): QueryKey => ['secrets', stackId],
+  byName: (name: string, stackId: number): QueryKey => ['secrets', stackId, name],
 };
 
-function secretsQuery(api: SMDataSource) {
-  return {
-    queryKey: QUERY_KEYS.list,
-    queryFn: () => api.getSecrets(),
-    throwOnError: true,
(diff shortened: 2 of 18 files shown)


src/data/clients/SecretsManagerClient/constants.ts
```diff
@@ -0,0 +1,12 @@
...
+ * management API. Secrets that include this decrypter are visible to SM
+ * and can be referenced by SM checks.
+ */
+export const SM_SECRET_DECRYPTER = 'synthetic-monitoring';
+
+/**
+ * Base path for the Grafana secrets management API.
+ * Full list URL: `${SECRETS_API_BASE}/namespaces/stacks-{stackId}/securevalues`.
+ */
+export const SECRETS_API_BASE = '/apis/secret.grafana.app/v1beta1';
```
g3john, line 12: Will this ever change? If so, how do we make sure we keep track of that?

VikaCep (approved):
This looks great!

Image

I notice with this new endpoint now we show the user id... is there any way we can show it in a friendlier way? not sure how much info this brings. In Query Library seems they hit the same thing (grafana/grafana#89230). Not blocking, but just worth considering as a follow-up.

src/data/clients/SecretsManagerClient/constants.ts
VikaCep: It will change when the API graduates from `v1beta1` to `v1`, but there's no automated way to track that today I believe. Might be worth adding a comment about this.

ckbedwell: Can we set up a SM check in ops targeting an environment that monitors when it breaks? 🤔

src/data/useSecrets.ts
```diff
@@ -35,11 +32,16 @@ function secretsQuery(api: SMDataSource) {
...
-  return useQuery<SecretsResponse, unknown, SecretWithMetadata[]>({
-    ...secretsQuery(smDS),
-    enabled,
+  return useQuery<SecretWithMetadata[], unknown, SecretWithMetadata[]>({
+    // eslint-disable-next-line @tanstack/query/exhaustive-deps -- `ctx` is derived from stackId, which IS in the key
+    queryKey: QUERY_KEYS.list(stackId),
+    queryFn: () => ctx!.client.fetchAll(),
+    throwOnError: true,
+    enabled: enabled && Boolean(ctx),
+    select: (secrets) => secrets.map((secret) => ({ ...secret, labels: secret.labels ?? [] })),
```
ckbedwell, line 44: The secrets client normalises the secret before returning the result so this select seems unnecessary?

ckbedwell (approved):
This is great, well done! 👏 

I'm going to approve so it's up to you whether you want to change this or not but one thing to think about from a code architecture perspective is whether there is a need for a memoized `SecretsClient` (and derived hooks, etc) at all. 

To be clear, I like that there is a SecretsClient and I think we should use this pattern more often to group related API calls. What I am questioning is whether the construction of the client should just be invoked on demand rather than than having an ever-living object that all the call sites share.

To give you a more pragmatic example of what that would look like in this PR:

```
export function useSecrets(enabled: boolean, stackId: number) {
  return useQuery<SecretWithMetadata[], unknown, SecretWithMetadata[]>({
    queryKey: [...QUERY_KEYS.list, stackId],
    queryFn: () => new SecretsManagerClient(stackId).fetchAll(),
    throwOnError: true,
    enabled: enabled && typeof stackId === 'number',
  });
}
```

This means you no longer have to do any typecasting or add the eslint disables in.

As we use React-query to manage all of our network state and caching, there are few advantages of having an ever-living client as it has no state to manage of its own.

g3john:
Merging for now, and can address reviews as a follow up later

PR #1669, opened 2026-04-28
feat: SLO integration drawer with create, link, delete, and query-match banners

ckbedwell:
## Problem

Users have no way to manage SLOs associated with their Synthetic Monitoring checks from within the SM app. They must navigate to the separate SLO app to create, view, link, or delete SLOs — even when the SLO clearly belongs to a specific check. Additionally, SLOs matched only by query (job label) have no way to be explicitly linked to a check, and SLOs linked to a different check can appear under the wrong check with no warning.

## Solution

Adds a full SLO integration drawer accessible from each check's dashboard header:

- **Create SLO**: Opens the SLO wizard inline within the drawer, pre-populated with the check's metrics and labels.
- **SLO detail tab**: Shows SLI, remaining error budget, and burn rate stat cards alongside objective, labels, and a link to the SLO dashboard.
- **Link to check**: When an SLO is matched by query only (no `sm_check_id` label), a banner prompts the user to explicitly link it. A second banner variant warns when an SLO is linked to a different check but query-matches the current one.
- **Delete SLO**: A destructive button with a confirmation modal calls the SLO plugin's `deleteSlo` API. SLOs with `readOnly.status.type === 'deleting'` are filtered out immediately so the count and drawer reflect the deletion.
- **Edit SLO**: Opens the SLO wizard inline for editing an existing SLO.

All features are wired through the `grafana-slo-app/slo-api/v1` extension point using `usePluginFunctions` and managed via React Query for caching and invalidation.

## Testing

Integration tests cover:
- Drawer opening, tab switching, and SLO count display
- Query-match and linked-to-other-check banner rendering
- Link-to-check action calling `updateSlo` and invalidating queries
- Delete confirmation flow calling `deleteSlo` and invalidating queries
- SloDetailTab stat cards, labels, dashboard link, edit, and delete button states
- `getMatchingSlosForSmCheck` filtering out deleting SLOs
- `linkSloToCheck` stripping `readOnly` and setting `sm_check_id`


VikaCep:
Is this one ready to review?

src/scenes/Common/useSLOCheckLinks.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+function useSLOPluginApi() {
+  const { functions, isLoading } = usePluginFunctions<GetSLOApi>({
+    extensionPointId: SLO_APP_API_EXTENSION_POINT_ID,
+  });
+  return { getSLOApi: functions[0]?.fn, isLoading };
+}
+
+export function useUpdateSLO() {
```
VikaCep, line 119: Seems `useUpdateSLO` and `useChecksForSLO`  have no consumers after the banner flow was removed, they're only referenced from their own tests.

src/scenes/Common/SLOIntegration.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const handleOpenDrawer = () => {
+    const firstSLOKey = slos.length > 0 ? slos[0].uuid : undefined;
+    setActiveTabKey(firstSLOKey ?? NEW_SLO_TAB_KEY);
+    setShowNewSLOTab(!firstSLOKey);
+    setEditingUuid(undefined);
+    setDrawerOpen(true);
+  };
+
+  const handleCloseNewSLOTab = () => {
```
VikaCep, line 124: With zero SLOs, Cancel clears `showNewSLOTab` but leaves `activeTabKey` as `'new-slo'`, and the wizard renders on `activeTabKey` alone so the tab disappears but the wizard keeps rendering and Cancel appears to do nothing. When `slos` is empty I'd expect Cancel to close the drawer entirely.

src/scenes/Common/SLOIntegration.tsx
```diff
@@ -0,0 +1,216 @@
...
+    () => queryClient.invalidateQueries({ queryKey: sloQueryKeys.all }),
+    [queryClient]
+  );
+
+  const handleDeleteSLO = useCallback(
+    async (slo: SLO) => {
+      setDeletingUuid(slo.uuid);
+      try {
+        const result = await deleteSLO(slo.uuid);
+        if (!result.error) {
```
VikaCep, line 80: If there's an error, can we surface it, e.g. via an appEvents error alert?

src/scenes/Common/useSLOCheckLinks.ts
```diff
@@ -0,0 +1,164 @@
...
+    extensionPointId: SLO_APP_API_EXTENSION_POINT_ID,
+  });
+
+  const getSLOApi = functions[0]?.fn;
+  const canFetch = pluginInstalled && !functionsLoading && typeof getSLOApi === 'function';
+
+  const query = useQuery({
+    queryKey: [...sloQueryKeys.all, getSLOApi],
+    queryFn: () => {
+      if (!getSLOApi) {
```
VikaCep, line 72: Nit: `getSLOApi` is a function, and react-query hashes keys via `JSON.stringify`, so it serialises to `null` and contributes nothing to cache identity. If the intent was to refetch when the plugin function re-registers, this doesn't achieve that — and if not, it's misleading and can be dropped from the key.

src/scenes/Common/SLOIntegration.utils.ts
```diff
@@ -0,0 +1,70 @@
...
+      (label) => label.key === SM_OBJECTIVE_KIND_LABEL_KEY && label.value === REACHABILITY_OBJECTIVE_KIND_VALUE
+    )
+  );
+}
+
+export function buildSLOWizardInitialValuesForCheck(
+  check: Check,
+  linkedSLOs: LinkedSLOLabels[]
+): SLOWizardInitialValues {
+  if (linkedSLOsHaveReachabilityObjectiveKind(linkedSLOs)) {
```
VikaCep, line 54: This guard looks unreachable at the moment: the create wizard only renders when `slos` is empty (the new-SLO tab only appears via `handleOpenDrawer` when there are no linked SLOs), so `linkedSLOs` is always `[]` when these initial values are actually consumed. Can we drop this guard?

VikaCep (changes requested):
Looking great! 🔥 
Left some comments. Also, the screen when editing service level indicators seems to need some spacing above the Ratio/Advanced buttons. Is this something we'd need to report to SLOs?

Image

ckbedwell:
Noticed that can't edit SLOs when reviewing this, put up an upstream PR for them here: https://github.com/grafana/slo/pull/4781

VikaCep (dismissed):
LGTM! 🔥

VikaCep:
>Noticed that can't edit SLOs when reviewing this, put up an upstream PR for them here: https://github.com/grafana/slo/pull/4781 

Should we disable the edit button while this issue is open?

VikaCep (approved):
LGTM!

PR #1671, opened 2026-05-04
feat: replace inline screenshot boilerplate with sm-screenshots jslib

VikaCep:
## Problem

The browser screenshot example snippets (`screenshotsLoki.js` and `screenshotsGCS.js`) contain ~200 lines of inline boilerplate — base64 encoding, chunking, Loki push logic, GCS upload via S3-compatible HMAC auth, and discovery log lines. This makes the examples hard for users to understand and copy into their own checks, and creates a maintenance burden if the log format ever changes.

## Solution

Replace both snippets with imports from the new [`sm-screenshots` jslib](https://github.com/grafana/k6-jslib-sm-screenshots), published at `https://jslib.k6.io/sm-screenshots/0.1.0/index.js`. The jslib produces the exact same log format as the previous inline code, so no frontend parsing changes are needed and secret names are unchanged.

The result is two clean ~35-line example scripts that clearly show users what they need to do: import the helpers, call `captureScreenshot`, and choose a storage backend (`loki()` or `gcs()`).

## Additional context

- Both snippets have been tested end-to-end against real SM infrastructure — screenshots appear correctly with thumbnails in execution logs.
- The required secrets are documented in script comments at the top of each snippet, which users see when the example is inserted into the editor.
- Relates to #1610 which originally added these snippets.
- PRs that adds the lib to the repo: https://github.com/grafana/k6-jslib-sm-screenshots/pull/2 and https://github.com/grafana/jslib.k6.io/pull/229

image
image
image

### DIFF
--- a/src/components/BrowserExamplesMenu/snippets/screenshotsGCS.js
+++ b/src/components/BrowserExamplesMenu/snippets/screenshotsGCS.js
@@ -1,117 +1,30 @@
 // Screenshots (External storage — GCS)
 //
-// Captures a screenshot, uploads the image to a Google Cloud Storage bucket
-// using S3-compatible HMAC authentication, and logs a URL reference to Loki.
-// The image bytes never pass through Loki — only a small reference log line
-// is stored. Best for large-scale checks where Loki ingestion volume is a concern.
-//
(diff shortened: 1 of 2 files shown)


g3john (approved):
LGTM!

PR #1682, opened 2026-05-19
feat: update k6 channel deprecation UX for v2 launch

VikaCep:
## Problem

When a k6 version channel was deprecated, it was completely hidden from the channel dropdown for new checks. This meant users couldn't select it even if they had a valid reason to (e.g. compatibility testing, gradual migration). With v2 becoming the new default channel, we need a gentler deprecation experience that informs users rather than blocking them — especially since v1 isn't deprecated yet but will be soon.

## Solution

Deprecated channels are now always visible in the dropdown regardless of whether the check is new or existing. Instead of hiding them, we show contextual messaging:

- **When v1 is selected (before its deprecation date):** a blue info alert explains that v2 is the recommended default, v1 will be deprecated in the coming weeks, and support is expected to end with k6 v3 (~May 2027).

image

- **When a deprecated channel is selected:** a yellow warning alert with a copy asking users to migrate their checks.

image

Test fixtures and assertions are updated to reflect v2 as the new default channel, with v1's deprecation date set to July 1st 2026 (~4 weeks from release).

## Context

This aligns with the team decision to launch v2 as the default while keeping v1 available with progressive warnings rather than hard-blocking users. v1's deprecation date is set to July 1st 2026. The hard cutoff (disabling v1 execution) will be handled separately when v3 ships.

### DIFF
--- a/src/components/CheckEditor/FormComponents/ChannelDetails.tsx
+++ b/src/components/CheckEditor/FormComponents/ChannelDetails.tsx
@@ -39,9 +39,11 @@ export function ChannelDetails({ channelId, channels }: ChannelDetailsProps) {
 
       {isDeprecated && (
         <Alert severity="warning" title="Deprecated k6 version channel">
-          This k6 version channel was deprecated on {dateTimeFormat(new Date(channel.deprecatedAfter))}. Support for
-          this version is expected to end with the release of k6 v3 (~May 2027). Please migrate your checks to a newer
-          channel.
+          This k6 version channel was deprecated on {dateTimeFormat(new Date(channel.deprecatedAfter))}.{' '}
+          {channel.id === 'v1'
+            ? 'Support for this version is expected to end with the release of k6 v3 (~May 2027).'
+            : 'Support for this version will end with the release of the next k6 major version.'}{' '}
+          Please migrate your checks to a newer channel.
         </Alert>
       )}
 
--- a/src/components/CheckEditor/FormComponents/ChannelDetails.tsx
+++ b/src/components/CheckEditor/FormComponents/ChannelDetails.tsx
@@ -49,8 +49,8 @@ export function ChannelDetails({ channelId, channels }: ChannelDetailsProps) {
 
       {isV1PreDeprecation && (
         <Alert severity="info" title="Newer version available">
-          v2 is now the recommended default. v1 will be deprecated in the coming weeks and support is expected to end
-          with the release of k6 v3 (~May 2027).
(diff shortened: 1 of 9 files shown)


src/components/CheckEditor/FormComponents/ChannelDetails.tsx
```diff
@@ -41,8 +39,16 @@ export function ChannelDetails({ channelId, channels }: ChannelDetailsProps) {
 
       {isDeprecated && (
         <Alert severity="warning" title="Deprecated k6 version channel">
-          This k6 version channel is deprecated since {dateTimeFormat(new Date(channel.deprecatedAfter))}. Consider
-          migrating to a newer channel.
+          This k6 version channel was deprecated on {dateTimeFormat(new Date(channel.deprecatedAfter))}. Support for
+          this version will end with the release of the next k6 major version. Please migrate your checks to a newer
```
markjmeier: We should be consistent here and also estimate May 2027

markjmeier (changes requested):
Only one minor change so we keep the dates front and center for the end users

src/components/CheckEditor/FormComponents/ChannelDetails.tsx
VikaCep: Updated it:

image

g3john (dismissed):
LGTM!

src/components/CheckEditor/FormComponents/ChannelDetails.tsx
```diff
@@ -41,8 +39,18 @@ export function ChannelDetails({ channelId, channels }: ChannelDetailsProps) {
...
+          {channel.id === 'v1'
+            ? 'Support for this version is expected to end with the release of k6 v3 (~May 2027).'
+            : 'Support for this version will end with the release of the next k6 major version.'}{' '}
+          Please migrate your checks to a newer channel.
+        </Alert>
+      )}
+
+      {isV1PreDeprecation && (
+        <Alert severity="info" title="Newer version available">
+          v2 is now the recommended default. v1 will be deprecated in the coming weeks and support is expected to end
```
markjmeier: I like the language from the other message better, lets drop the "and" and mirror it exactly.

v1 will be deprecated in the coming weeks. Support for this version is expected to end with the release of k6 v3 (~May 2027).

markjmeier (changes requested):

markjmeier (approved):
LGTM

PR #1689, opened 2026-05-21
fix: Update workflows to use more recent versions

mem:
This is still a work in progress that needs validation because the workflow is changing.

Take the existing "push" workflow and move it aside. Use a simpler version to validate the migration.

Fixes #


mem:
policy-bot: approve: migrating workflows

vmgrlab (dismissed):

mem:
policy-bot: approve

mem:
policy-bot: approve: updating worflows

VikaCep (changes requested):
A few comments:

- The `other_stuff` workflow should be renamed to something more meaningful. Maybe we can remove that file and just move the auto label job to `validate_pr`
- I noticed you removed `renovate_reviewer`. Can you explain why and if we can't apply any security measures to make it safer instead of removing it?
- I believe the DangerJS task won't work with these changes:
    - The actual file was renamed to `publish-main-bundle-size.yml`, not `main-bundle-size.yml`. `dawidd6/action-download-artifact `matches on the workflow filename, so it will never find the artifact.
    - `hashFiles` is missing the `.json` extension

.github/workflows/dangerJS.yml
```diff
@@ -27,22 +27,24 @@ jobs:
...
-          workflow: push.yml
+          workflow: main-bundle-size.yml
           workflow_conclusion: success
-          branch: main
+          if_no_artifact_found: ignore
           name: mainBundleSize
           path: artifacts
 
       - name: DangerJS
+        if: ${{ hashFiles('artifacts/mainBundleSize') != '' }}
```
VikaCep:
```suggestion
        if: ${{ hashFiles('artifacts/mainBundleSize.json') != '' }}
```

.github/workflows/dangerJS.yml
```diff
@@ -27,22 +27,24 @@ jobs:
...
+
+      - name: Generate current bundle size
+        run: yarn run stats artifacts/currentBundleSize.json
 
       - name: Download mainBundleSize
+        id: download-bundle
         uses: dawidd6/action-download-artifact@ac66b43f0e6a346234dd65d4d0c8fbb31cb316e5 # v11
         with:
-          workflow: push.yml
+          workflow: main-bundle-size.yml
```
VikaCep:
```suggestion
          workflow: publish-main-bundle-size.yml
```

mem:
> * The `other_stuff` workflow should be renamed to something more meaningful. Maybe we can remove that file and just move the auto label job to `validate_pr`

That's removed.

> * I noticed you removed `renovate_reviewer`. Can you explain why and if we can't apply any security measures to make it safer instead of removing it?

To be blunt: because that's the kind of thing that got us into this mess in the first place.

This PR is already much much much bigger than I was hoping for. I don't want to add "secure automerging renovate PRs" on top of everything else.

> * I believe the DangerJS task won't work with these changes:
>   
>   * The actual file was renamed to `publish-main-bundle-size.yml`, not `main-bundle-size.yml`. `dawidd6/action-download-artifact `matches on the workflow filename, so it will never find the artifact.
>   * `hashFiles` is missing the `.json` extension

That should be fixed now. The workflow is disabled (`if: false`) because it needs the artifact to be created in main first, so you have a chicken and egg problem: this PR needs merging for that to work, the workflow in this PR won't pass until this PR is merged. After merging this PR the conditional can be removed (in another PR) so that everything is enabled.

mem:
policy-bot: approve: updating workflows

VikaCep:
policy-bot: approve

guicaulada:
policy-bot: approve: updating policy-bot config

.github/workflows/validate-terraform.yml
```diff
@@ -1,32 +1,34 @@
...
+  pull_request:
+
+permissions: {}
 
 jobs:
   validate-terraform-generation:
     runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
```
VikaCep, line 13: This adds a comment on the PR when validation fails. Should we also add `pull-requests: write`?

.github/workflows/validate-probe-api-server-mappings.yml
```diff
@@ -1,12 +1,9 @@
...
 
 on:
-  workflow_call:
+  pull_request:
 
-permissions:
-  contents: read
-  issues: write
-  pull-requests: write
+permissions: {}
```
VikaCep, line 6: This one also comments on the PR when there is an error, should we add `pull-requests: write`?

.github/workflows/validate_main.yml
```diff
@@ -0,0 +1,37 @@
...
+      - main
+
+permissions: {}
+
+jobs:
+  # Main CI/CD workflow (shared workflow handles: lint, test, build, typecheck)
+  ci:
+    name: CI
+    # Documentation: https://enghub.grafana-ops.net/docs/default/component/grafana-plugins-platform/plugins-ci-github-actions/010-plugins-ci-github-actions
+    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@ci-cd-workflows/v8.0.0
```
VikaCep: The old `push.yml` used `cd.yml` here, which handled both CI and deployment to dev on merge to main. This now uses `ci.yml`.

As `ci.yml` doesn't deploy, merges to main will stop auto-deploying to dev, and the only way to deploy will be manually through `publish.yml`. To maintain our deployment to dev process we should be calling uses: `grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/v8.0.0` instead (cd.yml runs ci.yml internally before deploying, so CI checks are still covered).

VikaCep (changes requested):
With these changes we won't be auto-deploying to dev on merge to main. Left a comment with details and a suggested fix. 

Also two questions about permissions that I think will break the current terraform and probe-api-server-mappings workflows.

.github/workflows/validate-terraform.yml
mem: Thanks, this one should be fixed now.

.github/workflows/validate-probe-api-server-mappings.yml
mem: Same here.

.github/workflows/validate_main.yml
mem: I see.

Let's see if this makes sense:

- merges to main run validations (CI) and publish to dev (CD)
- the publish workflow publishes to whatever is specified in the input (CD), no validation (CI)

.github/workflows/validate-terraform.yml
mem: image

.github/workflows/validate_main.yml
VikaCep: That makes sense.
Just a heads-up: `cd.yml` already [runs ci.yml internally](https://github.com/grafana/plugin-ci-workflows/blob/ci-cd-workflows/v8.0.1/.github/workflows/cd.yml#L758) before any of the deploy jobs, every downstream job (`publish-to-catalog`, `trigger-argo-workflow`, etc.) depends on it. So with both ci and cd as separate jobs in `validate_main.yml`, the full CI suite will run twice on every merge to main. I think we could just keep the `cd` job and remove the standalone `ci` one.

VikaCep: > the publish workflow publishes to whatever is specified in the input (CD), no validation (CI)

`publish.yml` uses `cd.yml` too, and since `cd.yml` always runs `ci.yml` internally before any publishing/deploy steps, the publish workflow does run validation CI validation.

.github/workflows/validate-terraform.yml
```diff
@@ -0,0 +1,84 @@
...
+            else
+              echo "⚠️  Could not determine resolved provider version" | tee -a "$GITHUB_STEP_SUMMARY"
+            fi
+          fi
+          popd
+
+          yarn verify:terraform-test-config 2>&1 | tee /tmp/terraform_output.txt
+          exit_status=$?
+
+          if test "${exit_status}" -eq 0 ; then
```
VikaCep:
```suggestion
          if yarn verify:terraform-test-config 2>&1 | tee /tmp/terraform_output.txt ; then
```

.github/workflows/validate-terraform.yml
```diff
@@ -0,0 +1,84 @@
...
+          	4. **Update the transformation functions** in `src/components/TerraformConfig/`
+
+          	## 🧪 To test locally
+
+          	```bash
+          	yarn build:generate-terraform-test-config
+          	yarn verify:terraform-test-config
+          	```
+          	EOT
+          fi
```
VikaCep, line 82:
```suggestion
          	EOT
          	exit 1
          fi
```

.github/workflows/validate-probe-api-server-mappings.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+
+      - name: Setup Plugin Environment
+        uses: ./.github/actions/setup-env
+
+      - name: Verify probe mappings are up to date
+        id: verify-mappings
+        continue-on-error: true
```
VikaCep:
```suggestion
```

.github/workflows/validate-probe-api-server-mappings.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: |
+          # Run the verification and capture output to a temp file
+          yarn verify:probe-api-mappings 2>&1 | tee /tmp/verification_output.txt
+          status=$? # This should capture the failure from yarn
+          if test "$status" -eq 0 ; then
```
VikaCep:
```suggestion
          if yarn verify:probe-api-mappings 2>&1 | tee /tmp/verification_output.txt ; then
            :
```

VikaCep (changes requested):
I left some suggestion changes on `validate-terraform.yml` and `validate-probe-api-server-mappings.yml`:

The error handling code (`GITHUB_STEP_SUMMARY` messages) is currently dead on the failure path as the script exits before reaching it. For probe API, `continue-on-error: true` also means the job always passes. The suggestions wrap the command in `if` which lets the failure message get written before exiting.

VikaCep (dismissed):
👏 👏 👏

.github/workflows/validate_main.yml
mem: sorry, my bad, fixed.

mem: I removed the redundant ci.yml again, sorry 😞

VikaCep (approved):

PR #1692, opened 2026-05-21
feat: delete folder when all checks in folder are deleted

VikaCep:
## Problem

When a user selects all checks in a folder and deletes them, the Grafana folder is left behind empty. The user then has to navigate to the Dashboards section to clean it up. This is an unnecessary extra step, especially for users who are reorganising their monitoring setup.

## Solution

When all checks in a folder are selected and the user clicks delete, the confirmation modal now offers to delete the folder as well. The modal reads **Delete folder "X" + N checks**. On confirm, the checks are deleted first via the SM API, then the Grafana folder is deleted via `DELETE /api/folders/:uid`.

https://github.com/user-attachments/assets/e9ff4b80-87c1-4de2-83fc-1ef195b56b69

Folder deletion is safely skipped when:
- Only some checks in the folder are selected (standard "Delete N checks" behaviour)
- The folder is the default SM folder
- The folder is orphaned (already deleted in Grafana)
- The user lacks Grafana `folders:delete` permission on that folder

If the checks are deleted successfully but the folder deletion fails, a warning toast is shown so the user knows the folder remains.

https://github.com/user-attachments/assets/88683478-b3bc-4a5f-95d7-b070232decb7


vesnadean:
The interaction works well! 

We do have a pluralization problem with the delete modal copy. It says: "This will permanently delete the folder with all **1 checks** inside it."

I suggest we change it to:
"This will delete the folder, including 1 check. This action cannot be undone."

This way we can add as many checks as we want without breaking grammar. And I'm proposing to split the information that: _n number of objects will be deleted_, and _this action is permanent_, into two sentences to make it easier to parse.

VikaCep:
Nice catch!

I changed the copy to:

- Multiple checks:
image

- 1 check
image

vesnadean (dismissed):

g3john:
Should we also show the confirmation modal when deleting the check from the individual delete button? Otherwise the folder still exists

https://github.com/user-attachments/assets/0af25ee7-9a1b-42eb-9b02-ff0825eb17c0

VikaCep:
> Should we also show the confirmation modal when deleting the check from the individual delete button? Otherwise the folder still exists

@g3john We approached this from the folder's perspective. When a user is acting on all checks within a folder, offering to clean up the folder itself feels like a natural extension. For the individual check delete button, the user's intent is more focused on that specific check, so prompting about the folder might feel out of place. That delete component is also decoupled from any folder context today, so wiring it up would add extra complexity for a fairly small scenario. Happy to revisit if you feel strongly about it though.

src/page/CheckList/components/BulkActions.hooks.ts
```diff
@@ -1,16 +1,24 @@
-import { useCallback, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 
 import { Check, FeatureName } from 'types';
 import { useBulkCheckPermissions } from 'contexts/CheckFolderAccessContext';
 import { isFeatureEnabled } from 'contexts/FeatureFlagContext';
 import { useBulkDeleteChecks, useBulkUpdateChecks } from 'data/useChecks';
+import { deleteFolder as deleteFolderApi } from 'data/useFolders';
```
g3john: Should we use the useDeleteFolder hook instead? That was we can leverage the `invalidateAllFolders` attached to it

g3john:
> > Should we also show the confirmation modal when deleting the check from the individual delete button? Otherwise the folder still exists
> 
> @g3john We approached this from the folder's perspective. When a user is acting on all checks within a folder, offering to clean up the folder itself feels like a natural extension. For the individual check delete button, the user's intent is more focused on that specific check, so prompting about the folder might feel out of place. That delete component is also decoupled from any folder context today, so wiring it up would add extra complexity for a fairly small scenario. Happy to revisit if you feel strongly about it though.

Makes sense! 🚀

g3john (dismissed):

src/page/CheckList/components/BulkActions.hooks.ts
VikaCep: Great catch!

g3john (approved):
:shipit:

PR #1693, opened 2026-05-21
feat: add Grafana Assistant page context for Synthetic Monitoring

markjmeier:
## Problem

Users browsing the Synthetic Monitoring app and opening Grafana Assistant don't get help that's aware of which page they're on or what SM concepts are in scope. Assistant has to infer everything from the URL, which leads to generic answers that don't reference SM-specific concepts (checks, probes, k6 conventions, secrets, per-check alerts), and doesn't surface useful starter prompts at the right moments.

## Solution

This PR registers SM-specific page context with the `@grafana/assistant` SDK so Assistant knows where the user is in the SM app and can offer relevant first prompts.

Specifically:

- Static `structured` context per route (Home, Checks list, Choose check type, the four New Check sub-types by `CheckTypeGroup`, Check dashboard, Edit check, Probes list, View probe, Edit probe, New probe, Alerts, Config) with a short description of the page and what Assistant should help with.
- Starter questions per page that pre-fill the chat input with focused prompts and link canonical SM docs where relevant.
- Hidden "SM k6 conventions" context attached to the k6 scripted/browser/edit flows that tells Assistant to use `expect()` from the [k6-testing assertions library](https://grafana.com/docs/k6/latest/using-k6/assertions/) (rather than the classic `check()` API), avoid load-testing constructs, reference credentials via the SM Secrets module, and not suggest features that aren't currently functional (browser `page.screenshot()`).
- A small `<AssistantContext />` component mounted at the app root that registers all of the above when Assistant is available and unregisters on unmount.

The integration is a no-op when the `grafana-assistant-app` plugin isn't installed.

## How it works

- All registrations live in `src/components/AssistantContext/AssistantContext.constants.ts`, keyed off the existing `AppRoutes` enum and `PLUGIN_URL_PATH` constant so URLs don't drift.
- Routes with path parameters use regex patterns so e.g. `/checks/:id` does not also match `/checks/:id/edit`.
- The schema supports `createContextItems` and an optional `createQuestions` per entry, so adding new prompts or context for a page is a single-file edit.
- Tests assert that every `AppRoutes` value is covered (or in the explicit exclusion set for deep-link redirect routes), ids are unique, registration → unregistration lifecycle is correct, no starter prompt contains an unresolved bracket placeholder, hidden k6 conventions are attached to the right entries and recommend `expect()` over `check()`, and that "screenshot" only appears in DO-NOT framing.

## Verification

- Confirmed against a dev instance that the appropriate context chip appears in the Assistant sidebar on each SM route (and no chip appears on non-SM pages).
- Confirmed starter questions surface as clickable prompts on SM pages.
- Confirmed that scripts suggested by Assistant on the scripted / browser new-check pages use the k6-testing assertions library and SM Secrets module rather than perf-testing constructs.
- 14 new unit tests in `AssistantContext.test.tsx` covering constants coverage, registration lifecycle, k6-specific guardrails, and prompt hygiene. The full repo test suite still passes.

## Future work — input welcome from the frontend team

This PR is intentionally static — it gives Assistant the "where am I" awareness without flowing any live state through. The most impactful follow-ups, roughly ranked:

1. **Dynamic per-check context.** On the check dashboard and check edit pages, register a second context item alongside the static one containing `checkId`, `checkType`, `target`, `frequency`, `timeout`, probe count, and (for scripted/browser checks) the current script content. This unlocks "why is *this* check failing", "tune this timeout", and "help me fix this script" — questions the current static context cannot answer. Should use the SDK's `useProvidePageContext` hook wired into the existing react-query data hooks.
2. **`OpenAssistantButton` placements.** The SDK ships a self-hiding component that opens Assistant with a pre-filled prompt. Likely high-value placements: above the k6 script editor on the scripted/browser new and edit forms, on the check dashboard when the check is in a failing state, and near the per-check alert configuration. Each placement should use a distinct `origin` namespace (e.g. `grafana/sm/script-editor`) so entry points are distinguishable.
3. **Analytics on Assistant opens.** Wire `createSMEventFactory` Rudderstack events to fire when Assistant is opened from SM with `origin` as a property, so we can measure adoption and decide where to invest in deeper integration.
4. **Iterate on `help` and prompt copy.** The current copy was drafted without product input. A short pairing with an SM SE per area is likely the single biggest quality lever on answer relevance.

### Open questions worth raising with the Assistant team

Not blockers for this PR, but they'll shape the Phase 2 scope:

- **Write-back from Assistant into our editors.** The public SDK docs describe one-way page context → chat. The package's TypeScript definitions reference a `CALLBACK_EXTENSION_POINT` and a `tools` namespace that *appear* to enable functions Assistant could invoke (e.g. "insert this script into the editor"), but these aren't documented publicly. Worth confirming the supported pattern before scoping a script-insertion UX.
- **SM-specific k6 authoring.** Assistant currently uses a general k6 authoring tool oriented toward performance testing — different defaults, assertion patterns, and credential handling than synthetic monitoring needs. We've added hidden conventions to nudge the LLM toward transforming that tool's output for SM, but a dedicated SM-aware k6 authoring surface (registered by this plugin or owned elsewhere) would be a cleaner fix.
- **Browser screenshot support.** `page.screenshot()` is not currently functional in SM browser checks. The hidden conventions instruct Assistant not to suggest screenshots; when that feature ships, those rules should be relaxed.

### DIFF
--- a/src/components/AssistantContext/AssistantContext.constants.ts
+++ b/src/components/AssistantContext/AssistantContext.constants.ts
@@ -1,0 +1,518 @@
+import { type ChatContextItem, createAssistantContextItem, type Question } from '@grafana/assistant';
+
+import { PLUGIN_URL_PATH } from 'routing/constants';
+import { AppRoutes } from 'routing/types';
+
+/**
+ * Routes that are intentionally not registered with the Grafana Assistant
+ * because they have no user-facing surface area (they immediately redirect
+ * elsewhere on mount; see SceneRedirecter).
+ */
+export const ASSISTANT_CONTEXT_EXCLUDED_ROUTES: ReadonlySet<AppRoutes> = new Set([
+  AppRoutes.Redirect,
+  AppRoutes.Scene,
+]);
+
+interface AssistantPageContextEntry {
+  /** Stable id used for diffing the registration list in tests. */
+  id: string;
+  /** The AppRoutes value this entry covers. Used by the coverage test. */
+  route: AppRoutes;
+  /**
+   * URL pattern matched against `window.location.pathname` by the Assistant SDK.
+   * Regex is used wherever a route has path params, so that, e.g.,
+   * `/checks/:id` does not accidentally also match `/checks/:id/edit`.
+   */
+  urlPattern: string | RegExp;
+  /**
+   * Factory that builds the context items for this page. A factory (rather
+   * than a static array) keeps the items lazily constructed so the Assistant
+   * SDK is only invoked when a consumer needs the value.
+   */
+  createContextItems: () => ChatContextItem[];
+  /**
+   * Optional factory that builds starter questions for this page. These
+   * surface as clickable prompts in the Assistant UI to help users discover
+   * what they can ask. Bracketed `[placeholders]` are intentional — Assistant
+   * pre-fills the chat input with the prompt so the user can edit before
+   * sending.
+   */
+  createQuestions?: () => Question[];
+}
+
+function structured(
+  title: string,
+  data: {
+    name: string;
+    pageType: string;
+    capabilities: string[];
+    help: string;
+  }
+): ChatContextItem {
+  return createAssistantContextItem('structured', {
+    title,
+    bypassLimits: true,
+    data,
+  });
+}
+
+/**
+ * Builds a `structured` context item that is sent to the LLM but does NOT
+ * appear as a visible chip in the Assistant UI. Useful for rules-of-the-road
+ * that should always influence the model's answer without cluttering the UI.
+ */
+function hiddenStructured(name: string, data: Record<string, unknown>): ChatContextItem {
+  return createAssistantContextItem('structured', {
+    title: name,
+    hidden: true,
+    bypassLimits: true,
+    data: { name, ...data },
+  });
+}
+
+function question(title: string, prompt: string): Question {
+  return { title, prompt };
+}
+
+/**
+ * Hidden context attached to any page where Assistant is likely to draft or
+ * edit a k6 script for Synthetic Monitoring. Tools like `k6-authoring` are
+ * perf-testing oriented and emit load-test-shaped scripts; this context tells
+ * the model loudly to transform tool output to the SM single-VU model before
+ * presenting it to the user.
+ */
+function smK6Conventions(): ChatContextItem {
+  return hiddenStructured('Synthetic Monitoring k6 conventions', {
+    pageType: 'sm-k6-conventions',
+    audience: 'k6-script-authoring',
+    summary:
+      'This script is for Grafana Synthetic Monitoring (single-VU probe check on a recurring schedule), NOT k6 load testing.',
+    rules: [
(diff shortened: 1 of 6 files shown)


src/components/AssistantContext/AssistantContext.constants.ts
```diff
@@ -0,0 +1,518 @@
...
+  {
+    id: 'sm-new-check-browser',
+    route: AppRoutes.NewCheck,
+    urlPattern: newCheckBrowserRegex,
+    createContextItems: () => [
+      structured('Create a new k6 browser check', {
+        name: 'New k6 browser check',
+        pageType: 'sm-new-check-browser',
+        capabilities: ['k6-browser', 'browser-automation', 'page-assertions', 'multi-page-flows', 'probe-selection'],
+        help: `User is creating a k6 browser check (real headless browser) for synthetic monitoring. This is NOT a load test — the script runs as a single-VU probe check on a recurring schedule. Help with: writing browser scripts using the k6 browser API (page.goto, locator, waitForSelector), asserting on page content and DOM state with the k6-testing assertions library's auto-retrying matchers (await expect(page.locator(...)).toBeVisible(), .toHaveText(...), etc. — see ${DOC_URLS.k6Assertions}), navigating multi-page flows, referencing credentials via SM Secrets, and avoiding load-test constructs (no stages/VUs/thresholds). Prefer expect() over the classic check() function. Do NOT suggest page.screenshot() — screenshot capture is not currently functional in SM.`,
```
VikaCep: Should we also link to the recommended practices doc or give it as context? https://grafana.com/docs/k6/latest/using-k6-browser/recommended-practices/

src/components/AssistantContext/AssistantContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Registers static page context and starter questions with the Grafana
+ * Assistant SDK so that the Assistant knows which Synthetic Monitoring
+ * page the user is currently on and can suggest relevant first prompts.
+ *
+ * This is Phase 1 of the Assistant integration: a fixed set of `structured`
+ * context items and questions keyed by URL pattern. No dynamic state is
+ * exposed yet (e.g. the current check id is not included). See
+ * `.cursor/references/grafana-assistant-phase-2.md` for the follow-up plan.
```
VikaCep:
```suggestion
 * exposed yet (e.g. the current check id is not included).
```

VikaCep (dismissed):
This is super cool! 👏 👏 👏

src/components/AssistantContext/AssistantContext.constants.ts
markjmeier: Yeah this would be good!

markjmeier: we should send it to the markdown version of the docs: https://grafana.com/docs/k6/latest/using-k6-browser/recommended-practices.md

VikaCep (dismissed):

markjmeier:
@ckbedwell can you give @VikaCep and I a quick review of this?

ckbedwell (approved):
This is great so let's get it in! It should be moved to within pages so it has better context but let's just get this in an we'll iterate on it.

Issue #554, opened 2023-06-26
Summary dashboard link in Synthetics Home is broken (404)


adriansr:
**What happened**:

Home page link points to `/scenes/`, which results in a 404. 

Left navigation menu points to `/scenes`.

**What you expected to happen**:

Both work.

**How to reproduce it (as minimally and precisely as possible)**:

![sm404](https://github.com/grafana/synthetic-monitoring-app/assets/15056957/9b91d8c7-6830-4603-9f8d-212e4750cf3c)

**Anything else we need to know?**:

**Environment**:
- Grafana version (if running on prem):
- SM plugin version: 1.11.1
- OS Grafana is installed on:
- User OS & Browser:
- Others:

PR #1695, opened 2026-05-22
chore: migrate to yarn 4 with hardening settings

mem:
Yarn 1 reads .yarnrc (INI) and ignores .yarnrc.yml entirely, so the existing enableScripts: false setting in .yarnrc.yml was a no-op and the project had no enforced supply-chain controls beyond yarn 1's `ignore-scripts true` in the legacy .yarnrc.

Bump packageManager to yarn@4.15.0, regenerate yarn.lock in Berry format, and configure hardening in .yarnrc.yml:

  - enableScripts: false       (block dependency lifecycle scripts)
  - npmMinimalAgeGate: 4320    (reject npm packages younger than
    3 days; the field formerly called npmMinimumReleaseAge)
  - nodeLinker: node-modules   (keep node_modules layout for
    tooling that expects it)
  - approvedGitRepositories: [] (no current git deps; placeholder
    for future allowlist)

Remove the legacy .yarnrc file (yarn 1 only) and gitignore the new .yarn/install-state.gz while keeping releases/plugins/patches.

Update .github/actions/setup-env to enable corepack, use yarn 4's `yarn config get cacheFolder` for cache discovery, and drop the yarn 1-only `--prefer-offline` flag from `yarn install`.

Fixes #


VikaCep (changes requested):
The `corepack` approach breaks the shared `plugin-ci-workflows` -- the runners only have Yarn 1 globally and they've warned against using it in [#554](https://github.com/grafana/plugin-ci-workflows/issues/554#issuecomment-4517718863).

I've tested an alternative approach by vendoring the binary instead and pointing `yarnPath` to it, which seems to work: https://github.com/grafana/synthetic-monitoring-app/pull/1697

VikaCep:
> **TruffleHog Scan Results**
> 
> **Summary:** Found 1 potential secrets (0 verified, 1 unverified)
> 
> * **Possible secret** (Gitlab) at `.yarn/releases/yarn-4.15.0.cjs:711` → `npmP***ance`
> 
> **Review:** Check if unverified secrets are false positives.
> 
> **Ignoring False Positives:** To mark a false positive, add `# trufflehog:ignore` as an inline comment on the same line as the detected secret:
> 
> ```
> my_fake_secret = "AKIAIOSFODNN7EXAMPLE"  # trufflehog:ignore
> ```
> 
> This works for files that support line numbers (most source files). After adding the comment, push your changes and the scan will re-run.

The flagged string `npmP***ance` is `npmPublishProvenance`, a configuration key name inside the official Yarn 4.15.0 binary for npm's [provenance attestation feature](https://docs.npmjs.com/generating-provenance-statements), not a credential. The vendored file matches the official [@yarnpkg/cli-dist@4.15.0](https://www.npmjs.com/package/@yarnpkg/cli-dist/v/4.15.0) from npm.

ckbedwell (changes requested):
Looks good! But can we update the CONTRIBUTING.md to make it clear what devs need to do to get this to work? Maybe document the `.yarn/releases` folder?

VikaCep (approved):

ckbedwell (approved):
LGTM!

PR #1696, opened 2026-05-22
fix: add link to docs for v1 migration

VikaCep:
## Problem

The k6 v1 deprecation warning message was vague and gave no instructions on what to do to fix it.

## Solution

Update deprecation copy in the check editor (`ChannelDetails`) to match the version management rollout plan:

- State that **checks continue to run** after deprecation.
- **Recommend switching to the v2 k6 version channel** (v1-specific message).
- Reassure users that **most scripted and browser checks do not need script changes**.
- Link to the [k6 v2 migration guide](https://grafana.com/docs/k6/latest/get-started/migrating-to-v2/) only as troubleshooting if a check fails after switching—not as the primary migration path.

image


markjmeier (approved):
LGTM

PR #1699, opened 2026-05-26
chore(parseLokiLogs): replace raw data frame in error log with safe metadata

g3john:
## Problem

In `src/features/parseLokiLogs/parseLokiLogs.ts`, when `isLokiFields` rejects the post-normalization frame, we log the full `normalizedDataFrame` object to the console:

```ts
console.error('Failed to normalize LokiDataFrame fields', normalizedDataFrame);
```

That object's `fields[*].values` carry the actual Loki rows — log lines, user-supplied label keys and values, ids — i.e. potential PII and customer payloads.

This isn't just a hypothetical browser-console exposure. This app initialises Faro via `src/faro.ts`, and the Faro Web SDK ships with `ConsoleInstrumentation` enabled by default. Every `console.error(..., normalizedDataFrame)` call is therefore JSON-serialised and POSTed to the Grafana-Ops Faro collector with full row contents whenever this error path fires. That's an active information-disclosure path into observability infrastructure, not a speculative screen-share risk.

## Solution

Log only schema-level metadata that's useful for diagnosing why `isLokiFields` returned false, and that can never contain row data:

```ts
console.error('Failed to normalize LokiDataFrame fields', {
  fieldNames: normalizedDataFrame.fields.map((f) => f.name),
  length: normalizedDataFrame.length,
  refId: normalizedDataFrame.refId,
});
```

- `fieldNames` are schema constants (`labels`, `timestamp`, `body`, `Time`, `Line`, `tsNs`, `labelTypes`, `id`) — never user data — and tell us exactly which required field was missing.
- `length` is the row count, useful for distinguishing "empty result" from "malformed result".
- `refId` is the query identifier (`"A"`, `"B"`, …) set by the panel/scenes layer, useful for tracing the source query.
- `field.config`, `field.values`, frame-level `meta`, and the rest of the frame are excluded.

The error message string is unchanged so existing log searches and Faro alerts keyed on it keep working. No behaviour change beyond the console payload.

## Tests

A new test in `parseLokiLogs.test.ts` covers the error path:

- Constructs a malformed-but-normalized frame (new schema, missing `labelTypes` and `id`) so `normalizeLokiDataFrame` returns it unchanged and `isLokiFields` rejects it.
- Asserts `parseLokiLogs` returns `[]`.
- Asserts `console.error` was called with exactly `{ fieldNames, length, refId }` via Jest's strict deep-equal `toHaveBeenCalledWith` — which rejects any extra keys, so the metadata shape can't silently grow to include row data without failing this test.

### DIFF
--- a/src/features/parseLokiLogs/parseLokiLogs.test.ts
+++ b/src/features/parseLokiLogs/parseLokiLogs.test.ts
@@ -154,9 +154,17 @@ describe(`parseLokiLogs`, () => {
 });
 
 describe(`parseLokiLogs error path`, () => {
-  it(`returns [] and logs only safe schema metadata when required fields are missing`, () => {
-    const consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+  let consoleErrorSpy: jest.SpyInstance;
 
+  beforeEach(() => {
+    consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    consoleErrorSpy.mockRestore();
+  });
+
+  it(`returns [] and logs only safe schema metadata when required fields are missing`, () => {
     // New-schema frame missing labelTypes and id, so normalizeLokiDataFrame
     // returns it as-is and isLokiFields rejects it.
     const malformedFrame = {
@@ -170,13 +178,14 @@ describe(`parseLokiLogs error path`, () => {
     } as unknown as LokiDataFrame<Record<string, string>, Record<string, string>>;
 
     expect(parseLokiLogs(malformedFrame)).toEqual([]);
+    // toHaveBeenCalledWith uses strict deep-equal; any extra key (e.g. row data
+    // smuggled into the metadata object) would fail this assertion, which is
+    // the no-PII-in-logs guarantee we care about.
     expect(consoleErrorSpy).toHaveBeenCalledWith('Failed to normalize LokiDataFrame fields', {
       fieldNames: [LokiFieldNames.Labels, LokiFieldNames.TimeStamp, LokiFieldNames.Body],
       length: 0,
       refId: 'A',
     });
-
-    consoleErrorSpy.mockRestore();
   });
 });
 
--- a/src/features/parseLokiLogs/parseLokiLogs.test.ts
+++ b/src/features/parseLokiLogs/parseLokiLogs.test.ts
@@ -153,6 +153,73 @@ describe(`parseLokiLogs`, () => {
   });
 });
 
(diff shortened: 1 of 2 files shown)


src/features/parseLokiLogs/parseLokiLogs.test.ts
```diff
@@ -153,6 +153,42 @@ describe(`parseLokiLogs`, () => {
...
+    // smuggled into the metadata object) would fail this assertion, which is
+    // the no-PII-in-logs guarantee we care about.
+    expect(consoleErrorSpy).toHaveBeenCalledWith('Failed to normalize LokiDataFrame fields', {
+      fieldNames: [LokiFieldNames.Labels, LokiFieldNames.TimeStamp, LokiFieldNames.Body],
+      length: 0,
+      refId: 'A',
+    });
+  });
+});
+
```
VikaCep, line 208: Can we add another test with non-empty field values?

```
it(`does not leak row values when fields contain data`, () => {
  const malformedFrame = {
    length: 1,
    refId: 'A',
    fields: [
      { name: LokiFieldNames.Labels, values: [{ secret: 'should-not-leak' }] },
      { name: LokiFieldNames.TimeStamp, values: [1234567890], nanos: [0] },
      { name: LokiFieldNames.Body, values: ['sensitive log line content'] },
    ],
  } as unknown as LokiDataFrame<Record<string, string>, Record<string, string>>;
  parseLokiLogs(malformedFrame);
  const loggedArg = JSON.stringify(consoleErrorSpy.mock.calls[0]);
  expect(loggedArg).not.toContain('should-not-leak');
  expect(loggedArg).not.toContain('sensitive log line content');
});
```

VikaCep (dismissed):
LGTM! Left a suggestion to add an extra test.

src/features/parseLokiLogs/parseLokiLogs.test.ts
g3john: Added in f5cff305 as a second `it()` in the same describe block.

Fair callout — when Copilot raised the same point earlier I dismissed it on technical grounds (the existing `toHaveBeenCalledWith` uses strict deep-equal, so any extra key in the actual call would already fail the assertion, including row data smuggled into the metadata object). That's still true and the existing test still encodes the guarantee.

But your independent ask of the same thing tells me the strict-deep-equal property reads as a guarantee to me but not to reviewers landing fresh on the file — what people naturally look for in a "no PII in logs" test is rows-in, rows-not-out. Kept both tests so we get the shape assertion and the explicit non-leak demonstration. Thanks for pushing on it.

VikaCep (dismissed):

VikaCep (approved):

PR #1700, opened 2026-05-26
chore(eslint): warn on console.* in src/ production code

g3john:
## Problem

`@grafana/eslint-config` ships `no-console` with an `allow` list covering `error`, `log`, `warn`, and `info`, which makes the rule effectively a no-op. Console calls in `src/` production code land with zero feedback at lint time.

## Solution

Override `no-console` in our `src/`-scoped flat-config block so every console method warns, with a test-files override that exempts tests. `scripts/` is unaffected (already outside our lint scope).

Severity is `warn`, so CI is not affected. Surfaces 13 existing call-sites for per-call-site triage in follow-ups.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -25,6 +25,7 @@ const config = defineConfig([
     },
     rules: {
       'custom-plugin/tracking-event-creation': 'error',
+      'no-console': ['warn', {}],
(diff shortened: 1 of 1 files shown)


VikaCep (approved):
LGTM!

PR #1701, opened 2026-05-26
fix: prevent sensitive data leaking through Faro error reports

VikaCep:
## Problem

The `reportError` function in `faro.ts` accepted arbitrary objects (`Error | Object | string`) and serialised them with `JSON.stringify` before sending to Faro. This meant any caller that passed a full error response object — including API responses containing tokens, credentials, or internal data — would have its entire contents logged as a Faro error.

One active caller in `useAppInitializer.ts` had a `?? err` fallback that passed the entire `SubmissionErrorWrapper` object (including `err.data`) to `reportError` when both `.msg` and `.err` were falsy.

## Solution

Narrowed the `reportError` and `sanitizeError` signatures from `Error | Object | string` to `Error | string`, removing the `JSON.stringify` branch entirely. If a non-Error, non-string value reaches `sanitizeError` at runtime (bypassing TypeScript), `String()` produces `[object Object]` rather than dumping the object's contents.

Fixed the `useAppInitializer.ts` catch block to extract the message string before calling `reportError`, falling back to `'Something went wrong'` instead of the raw error object.

Added tests covering the key security properties: Error passthrough, string wrapping, object rejection, optional type parameter, and exception swallowing.

### DIFF
--- a/src/faro.test.ts
+++ b/src/faro.test.ts
@@ -1,0 +1,57 @@
+import { faro } from '@grafana/faro-web-sdk';
+
+import { FaroEvent, reportError } from 'faro';
+
+const mockPushError = jest.fn();
+
+beforeEach(() => {
+  // faro is a getter-backed singleton that lazily initialises api after initializeFaro().
+  // In tests we haven't called initializeFaro(), so api is undefined.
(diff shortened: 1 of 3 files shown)


g3john (approved):
:shipit:

PR #1703, opened 2026-05-26
fix: Use `contents: write` in validate_main.yml

mem:
plugin-ci-workflows/cd.yml won't work unless you have `contents: write` permission. We don't need it because we don't publish docs to the website, but we must have it, otherwise this won't work.

Fixes #


VikaCep:
Merged https://github.com/grafana/synthetic-monitoring-app/pull/1702 instead as we also needed `pull-requests: read` on CD from`validate_main.yml`

PR #1705, opened 2026-05-26
fix: Document workflow permissions

mem:

Fixes #

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,10 +26,20 @@ jobs:
   cd:
     name: CD
     uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/v8.0.1
+    #  About permissions below.
+    #
+    # - attestations: write: publish attestations.
+    # - contents: write: checkout code (read); create a git tag (write, not
+    #   used); create a GitHub Release (write, not used); publish techdocs to
+    #   webiste (not used). Change to contents: read as soon as possible.
+    # - id-token: write: used to mint various tokens.
+    # - pull-requests: read: list PRs to determine if the branch from which the
+    #   release is created is an open PR. We don't need this because we set
+    #   branch=main. Remove this as soon as possible.
     permissions:
+      attestations: write
       contents: write
       id-token: write
-      attestations: write
       pull-requests: read
     with:
       branch: ${{ github.event.inputs.branch }}
--- a/.github/workflows/dangerJS.yml
+++ b/.github/workflows/dangerJS.yml
@@ -3,12 +3,7 @@ name: DangerJS
 on:
   pull_request:
 
-permissions:
-  actions: read         # dawidd6/action-download-artifact
-  contents: read        # actions/checkout
-  issues: read          # Why?
-  pull-requests: write  # Used to post a message to the PR with the results.
-  statuses: write       # Update the PR status
+permissions: {}
 
 jobs:
   danger-js:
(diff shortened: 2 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -27,10 +27,10 @@ jobs:
...
     uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/v8.0.1
     permissions:
-      contents: write
-      id-token: write
-      attestations: write
-      pull-requests: read
+      attestations: write  # writes artifact attestatins
+      contents: write      # write is only needed if we publish docs to the website, which we don't, but the workflow won't work without this.
+      id-token: write      # generates token
+      pull-requests: read  # why?
```
VikaCep: Because [cd.yml](url) [declares](https://github.com/grafana/plugin-ci-workflows/blob/ci-cd-workflows/v8.0.1/.github/workflows/cd.yml#L556) `pull-requests: read` in its top-level permissions block. Same [reason](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#nesting-reusable-workflows) as with other cases, the caller must match it due to the "maintained or reduced" rule, regardless of whether the code path executes.

Actual usage is [here](https://github.com/grafana/plugin-ci-workflows/blob/ci-cd-workflows/v8.0.1/.github/workflows/cd.yml#L722) where it lists open PRs for the branch to block prod deployments from branches with open PRs. This code only runs inside an if `(environments.includes('prod') || environments.includes('prod-canary'))` block, so for dev deployments it won't execute, but GitHub validates permissions at parse time, not runtime.

mem: > Because cd.yml declares pull-requests: read in its top-level permissions block. Same reason as with other cases, the caller must match it due to the "maintained or reduced" rule, regardless of whether the code path executes.

sure, but why is that needed?

mem: Answering my own question, this is covering a rather bizarre use case that does not apply to us:

1. You create release/1.2.0, push some last-minute changes.
2. You open a PR release/1.2.0 → main for review.
3. Before the PR is merged/reviewed, someone triggers cd.yml with branch=release/1.2.0, environment=prod.
4. The check in the code sees the open PR from release/1.2.0 to main and blocks the deploy.

That still doesn't make sense to me, but that's what the code does. Why would you point cd.yml at release/1.2.0 when that is a branch that you can use to create PRs?

The other case is even more bizarre.

There's another setting `allow-publishing-prs-to-prod: true`, which allows you to have an open PR in `it-is-my-party-i-will-do-whatever-i-want` *and* release that to prod 😱. You have to couple that with other settings, like this:

```
      uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/v8.0.1
      with:
        branch: ${{ inputs.pr-branch }}
        environment: prod-canary
 
        # Widen the release-reference regex so the PR's branch is treated
        # as a valid release reference. Without this, cd.yml fails with
        # "not a release reference" before it ever gets to the open-PR check.
        release-reference-regex: main|${{ inputs.pr-branch }}

        # Explicitly acknowledge that we're deploying from a branch
        # that still has an open PR. Required because of the regex above.
        allow-publishing-prs-to-prod: true
  
        # Suffix the version with the commit SHA so this prod-canary build
        # has a unique version and never collides with a real prod release.
        # (cd.yml does this by default for prod-canary, but being explicit
        # avoids surprises if you later change the environment value.)
        plugin-version-suffix: ${{ github.sha }}

        # Don't draft a GitHub release for an unmerged PR.
        disable-github-release: true
    
        # Don't publish docs — that's prod-only territory.
        disable-docs-publishing: true
```

If this is the intention, I don't understand why you cannot simply create a branch and merge to that as usual, without resorting to this "let's deploy unmerged code!" dance.

mem: For that last bit, 1) the workflow has to exist; 2) a GATB config with `branches: "*"` would be required.

mem: I have updated the permissions and the comment.

VikaCep (approved):
LGTM!

PR #1711, opened 2026-05-28
feat: show empty folders in check list folder view

VikaCep:
## Problem

Empty folders were invisible in the check list folder view. Users had no way to see or clean up folders that no longer contained any checks, leading to clutter in Grafana's folder structure that could only be managed outside of Synthetic Monitoring.

## Summary

- **Show empty folders** in the check list folder view. They appear below the default folder, sorted alphabetically after folders that contain checks.
- **Delete empty folders** via a trash icon button on the folder row. Clicking it opens a confirmation modal. Only shown when the user has Grafana-level `canDelete` permission on the folder.

image
image

https://github.com/user-attachments/assets/b5b9db71-0ea5-4335-bfea-0795ab1756ae

### DIFF
--- a/src/hooks/useCheckFolderAccess.ts
+++ b/src/hooks/useCheckFolderAccess.ts
@@ -46,8 +46,9 @@ export function useCheckFolderAccess<T extends Pick<Check, 'folderUid'>>(checks:
         uids.add(check.folderUid);
       }
     });
+    allFolders.forEach((folder) => uids.add(folder.uid));
     return [...uids];
-  }, [checks, isFoldersEnabled, defaultFolderUid]);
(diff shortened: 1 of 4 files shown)


vesnadean:
I think we should keep the delete pattern the same across all folders, empty or not. This means that the delete icon becomes visible once the checkbox is selected, and we don't disable anything. Right now it's a little bit off that only delete icons for empty folders are visible by default.

VikaCep:
Good call @vesnadean, I've made some changes to keep the delete icon consistent in all cases by only showing it after the checkbox is clicked, even if the folder is empty. Updated the screenshots on the PR description with these changes.

vesnadean (approved):

PR #1722, opened 2026-06-11
fix: handle spread operator in browser script options validation

VikaCep:
## Problem

Browser check scripts that build their exported `options` with a spread operator (e.g. `export const options = { ...defaultOptions }`) crash the check form with an uncaught `TypeError: Cannot read properties of undefined (reading 'type')`. The error is thrown on every validation pass, which makes it impossible to update or save the check.

The crash happens because the static script analysis assumes every entry in an object literal is a regular property, but spread elements have no `key` node.

## Solution

The script parser is now spread-aware:

- Spreads of objects declared in the same script are resolved, so their contents are validated exactly as if they were written inline — including the required chromium browser type and the `duration`/`vus`/`iterations` restrictions. Later properties correctly override earlier spreads, matching JavaScript semantics.
- Spreads that can't be statically resolved (e.g. computed values) fail validation with a clear message asking the user to define the browser options inline, instead of throwing.
- Circular references between spread objects are guarded against, so they can't hang or crash the form either.

## Testing

Added a unit test suite for `validateBrowserScript` covering all validation paths: valid scripts, resolved spreads (passing and failing), spread override semantics, unresolvable spreads, and the pre-existing rules (chromium type, duration, vus, iterations, browser import).

**Before**

https://github.com/user-attachments/assets/ae99ccbf-e56b-44cd-aa89-bbe4dbcb82b6

**After**

https://github.com/user-attachments/assets/6f45c9f1-d602-4349-bc90-0d97778586a3

### DIFF
--- a/src/schemas/forms/script/parser.ts
+++ b/src/schemas/forms/script/parser.ts
@@ -1,4 +1,4 @@
-import { Node, ObjectExpression, parse, Property } from 'acorn';
+import { Node, ObjectExpression, parse, Property, SpreadElement } from 'acorn';
 import { simple as walk, SimpleVisitors } from 'acorn-walk';
 
 export function parseScript(script: string) {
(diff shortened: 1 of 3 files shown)


src/schemas/forms/script/parser.ts
```diff
@@ -60,20 +96,64 @@ const optionExportMatcher: SimpleVisitors<{ options: ObjectExpression | null }>
...
-    return false;
+    return undefined;
   }
 
-  for (const scenario of scenarios.properties) {
-    if (scenario.key.type === 'Identifier') {
-      const propValue = getPropertyValueByPath(scenario.value, propPath);
+  for (const scenario of expandProperties(scenarios, declarations)) {
+    if (scenario.key.type === 'Identifier' && scenario.value.type === 'ObjectExpression') {
+      const propValue = getPropertyValueByPath(scenario.value, propPath, declarations);
```
cristianoventura, line 118: This isn't related to the use case we discovered, but I noticed this will resolve the first entry for properties spread inside the `scenario` object.

```ts
const defaultScenarios = {
  ui: {
    executor: 'shared-iterations',
    options: { },
  },
}

export const options = {
  scenarios: {
    ...defaultScenarios,
    executor: 'shared-iterations',
    ui: {             // does not override defaultScenarios, when it should
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  }
}

```

cristianoventura (dismissed):
LGTM, thanks for the quick fix 🙏

src/schemas/forms/script/parser.ts
VikaCep: Good catch! Addressed in the last commit 👍

cristianoventura (approved):
🚀

Issue #1588, opened 2026-02-11
Properly handle displaying data for custom retention periods


Pokom:
**What happened**:

Timepoint explorer on the check overview page does not respect custom log retention policies(>31 days), resulting in an error being rendered instead of the data. 

**What you expected to happen**:

Timepoint explorer to render data that is returned from Loki's datasource. Interestingly enough, if you view the network requests, Loki returns the data. The app makes the decision to not render anything, even if some data is returned. 

**How to reproduce it (as minimally and precisely as possible)**:

This came from an internal escalation, so please reach out and I can share the details from there. But generally speaking:

1. Have an org with retention policy >31 days
2. Have data >31 days old
3. View a check with a timeframe that is between 31 days and the retention period

---

This code here(https://github.com/grafana/synthetic-monitoring-app/blob/f9142d3afc7ce9ad1a0afe3ad0c5099b9682404c/src/data/useLogsRetention.ts#L5-L22) implies that we need to make gcom calls. I don't speak with authority, but I would be curious as to whether or not we can simply make the Loki calls and render data if it's returned. That's what is happening today, but we chose to ignore returned data. 

**Anything else we need to know?**:

**Environment**:
- Grafana version (if running on prem):
- SM plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

sapiensantiquus:
Hello, I'm the person who originally discovered this bug. Is it possible to get an update? Our customers need the information that is missing and are waiting on it.

PR #1724, opened 2026-06-12
fix: detect logs retention period from Loki instead of hardcoding it

VikaCep:
Addresses: https://github.com/grafana/synthetic-monitoring-app/issues/1588

## Problem

The Timepoint Explorer assumed every org has a logs retention period of 31 days (14 for free tier). For orgs with custom retention policies (> 31 days), the explorer refused to display anything older than 31 days and rendered an "Out of retention period" annotation instead — even though Loki returns the data when queried. There is no endpoint available to plugins which exposes a tenant's actual retention period (it lives in gcom), which is why the values were hardcoded.

## Solution

Instead of hardcoding the retention period, we now infer it from Loki itself. On mount the explorer sends a single "canary" query to the logs datasource with a time range that is deliberately far outside any plausible retention period (a 1-minute window 10 years in the past, cached for the session):

- **Loki rejects the query** with `this data is no longer available, it is past now - max_query_lookback (2160h)` — we parse the configured limit from the error message and use it as the retention period. Grafana Cloud keeps `max_query_lookback` in step with the tenant's retention, so custom retention orgs now get an accurate boundary. The rejection is deterministic when the whole range is out of bounds ([Loki docs](https://grafana.com/docs/loki/latest/query/troubleshoot-query/#error-data-is-no-longer-available)).
- **The query succeeds** (no lookback limit enforced for the tenant) — the explorer applies no clamping at all: it queries the full requested time range, renders whatever data Loki returns, and shows "No data" for the rest.
- **Anything else fails** — we fall back to the previous hardcoded values, so behaviour is never worse than before.

This is the same `max_query_lookback` error message the assertions table already parses for its "beyond retention" warning.

## Trade-offs and notes for reviewers

- On tenants where `max_query_lookback` is not enforced, the explorer can no longer distinguish "out of retention" from "no data" — old empty ranges show the "No data" annotation rather than "Out of retention period". This is the agreed fallback from the issue discussion.
- The canary query runs once per session and is cheap: it never returns data, as it is rejected by the query frontend before touching storage.

## Testing

- Unit tests cover the three canary outcomes (parsed limit, no limit enforced, unexpected failure) plus the error-message and Go-duration parsing.
- Verified manually against a real stack without `max_query_lookback` enforcement: a 90-day range now renders timepoints across the full range with "No data" where logs no longer exist, instead of cutting off at 31 days.
- Not yet verified against a tenant that both enforces `max_query_lookback` and has custom retention (the scenario from the original escalation) — reviewer with access to such a stack welcome.

---

> [!NOTE]
> **Medium Risk**
> Changes how Timepoint Explorer bounds log queries and retention UX (including no clamp when lookback is not enforced); failures fall back to prior hardcoded limits.
> 
> **Overview**
> Replaces **hardcoded** 14/31-day logs retention with a **session-cached Loki canary query** that probes `max_query_lookback`, so Timepoint Explorer can honor **custom retention** instead of falsely marking older data as out of retention.
> 
> `useLogsRetentionPeriod` no longer takes a time-range argument; it returns `{ retentionPeriod, isLoading }`. When Loki rejects the canary with the documented lookback error, the limit is parsed from Go duration strings; when the query succeeds, `retentionPeriod` is **null** and the explorer **does not clamp** the range. Unexpected failures still use the previous free/standard day fallbacks while loading or on error.
> 
> Timepoint Explorer waits on retention loading, treats **null** retention as “no clamp” (`logsRetentionFrom` → 0), and adds unit tests for the three canary outcomes plus parsing helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2bbb062c17821308377de136f7492f1fe63c59c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/data/useLogsRetention.test.ts
+++ b/src/data/useLogsRetention.test.ts
@@ -1,0 +1,114 @@
+import { renderHook, waitFor } from '@testing-library/react';
+import { apiRoute } from 'test/handlers';
+import { createWrapper } from 'test/render';
+import { server } from 'test/server';
+
+import { Result } from 'test/handlers/types';
+
+import {
+  parseGoDuration,
+  parseMaxQueryLookback,
+  REF_ID_LOGS_RETENTION_CANARY,
+  useLogsRetentionPeriod,
+} from './useLogsRetention';
+
+const MILLISECONDS_IN_HOUR = 60 * 60 * 1000;
+const STANDARD_RETENTION_PERIOD = 31 * 24 * MILLISECONDS_IN_HOUR;
+const NINETY_DAYS_IN_HOURS = 2160;
+
+const MAX_QUERY_LOOKBACK_ERROR_MESSAGE = `this data is no longer available, it is past now - max_query_lookback (${NINETY_DAYS_IN_HOURS}h0m0s)`;
+
+function mockCanaryQuery(result: Result<unknown>) {
+  server.use(
+    apiRoute(`getHttpDashboard`, {
+      result: () => result,
+    })
+  );
+}
+
+describe(`useLogsRetentionPeriod`, () => {
+  it(`returns the retention period parsed from Loki's max_query_lookback error`, async () => {
+    mockCanaryQuery({
+      status: 400,
+      json: {
+        results: {
+          [REF_ID_LOGS_RETENTION_CANARY]: {
+            error: MAX_QUERY_LOOKBACK_ERROR_MESSAGE,
+            status: 400,
+          },
+        },
+      },
+    });
+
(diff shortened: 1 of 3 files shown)


VikaCep:
cursor review

vmgrlab:
Bugbot review

VikaCep:
Bugbot review

ckbedwell (approved):
LGTM!

PR #1725, opened 2026-06-12
Feat/traceroute path health

markjmeier:
## What

Reworks the traceroute check dashboard to tell an accurate path-health story.
Presentation layer only — no agent, metric-name, or log-format changes; every
panel is built from existing Prometheus metrics and Loki logs.

| Problem with the old view | Change |
|---|---|
| "Average latency" stat showed multi-second execution walltime with alarming red thresholds — it was never network latency | Renamed **"Trace completion time"**, neutral coloring, description clarifies it is walltime for all mtr cycles |
| "Overall packet loss" averages loss across ALL hops; intermediate routers deprioritize ICMP TTL-exceeded replies, so 40%+ readings are normal on healthy paths | Retitled **"Overall packet loss — all hops"** with a caution description, complemented by per-hop views so users can see whether loss persists to the destination (only that is real forwarding loss) |
| Route hash plotted as a raw value — unreadable | **Path changes stat** (changes over selected range) + **bar timeseries** of changes per 15m window. The window is fixed rather than `$__interval`, which can be narrower than check frequency and renders empty |
| Per-hop data only visible one execution at a time in the log explorer | **mtr-style per-hop table** (hosts, avg loss %, avg RTT per TTL, aggregated over the time range, loss color-coded) + **per-hop loss timeseries** smoothed over 15m windows so ECMP path-hopping reads as bands instead of 0↔100% sawtooth |
| No way to correlate route changes with symptoms | **Path-change annotations** on all timeseries panels |

The node graph and Timepoint Explorer are unchanged.

## Annotation design

Path-change annotations only fire when route churn (>4 hash changes/15m)
coincides with `probe_success == 0`, and the toggle defaults to off. ECMP
networks flap the route hash between equally-valid paths constantly (~195
changes/3h measured on a healthy production target), so churn alone is not
an event — a path change is worth a marker when there is a symptom for it
to explain.

## Shared code changes

- `DashboardContainer` accepts an optional `extraAnnotations` prop (appended
  after alert annotations; no behavior change for other check types)
- New `ToggleableAnnotationLayer`: identical to the scenes-react
  `AnnotationLayer` except the initial toggle state honors `query.enable` —
  the library version always constructs layers enabled. Existing alert
  annotations are `enable: true`, so their behavior is unchanged.

## Validation

Verified against a production traceroute check (3 probes → pdt.grafana.net)
and a deliberately failing check:

- Walltime stat reads neutral; middle-hop ICMP "loss" (~40%) presents as
  informational while destination rows sit green at ~0%
- ECMP hop-count flapping (8–11) and constant path churn render as path
  characterization, not alarm
- Annotation verified firing exactly at a real failure window that coincided
  with a path change, with explanatory tooltip

## Rollout / open questions

- This replaces the old traceroute view outright. A commented
  `FeatureName.TraceroutePathHealth` entry exists if we'd rather gate it;
  the old layout is recoverable from git history.
- Not yet tested with real users — feedback welcome on whether this should
  land flag-gated first.
- Known gap: a permanent reroute on a *passing* check no longer annotates.
  Solving that well needs novel-path detection, which wants the route hash
  emitted as a label rather than a gauge value — an agent-side follow-up. this might not be needed?

## Screenshots

# Before state
Screenshot 2026-06-12 at 10-52-08

# after state
Screenshot 2026-06-12 at 10-55-10


markjmeier:
this PR definitely needs some user testing - it exists here for discussion purposes. 

I think there would be a real concern on suddenly changing the page on users - most of the same data should be captured and presented better, but could use some scrutiny

ckbedwell:
@markjmeier let's go over this in 1:1 at some point. A few things to change to make this more maintainable. @VikaCep and I just took a quick look.

markjmeier:
@ckbedwell sounds good - this was mostly exploration to improve what we have today and it has very light feedback. I think some validation would be good before we commit too much (even if its internal/ourselves)

Issue #1717, opened 2026-06-04
Migrate feature flags to OpenFeature


VikaCep:
# Migrate feature flags from Grafana feature toggles to OpenFeature

## Summary

Grafana is deprecating legacy feature toggles in favor of OpenFeature flags configured with Go Feature Flag (GOFF). We hit this when adding the CALs feature flag in deployment_tools and were nudged to migrate:

> ⚠️ Legacy feature toggles are discouraged. Adding new legacy feature toggles is no longer recommended; please use OpenFeature flags, configured with Go Feature Flag (GOFF), instead.

Ref: https://github.com/grafana/deployment_tools/pull/605954#issuecomment-4626003190

## Why

- Legacy Grafana feature toggles are binary (on/off) and require a deploy to change.
- OpenFeature + GOFF supports richer targeting: per-stack/org/tenant rules, percentage-based gradual rollouts, and runtime changes without a deploy.
- Aligns with the recommended direction across Grafana; avoids adding to a deprecated system.

## Current state

- The app reads flags via `config.featureToggles` (Grafana legacy toggles), centralized in:
  - `src/contexts/FeatureFlagContext.ts` (`isFeatureEnabled`, `getFeatureContextValues`)
  - `src/hooks/useFeatureFlag.ts` and `useFeatureFlagContext`
- `@openfeature/web-sdk` (`1.7.3`) is already a dependency but is **not yet wired in**.
- Existing flags include `FeatureName.CALs`, `FeatureName.Folders`, `FeatureName.TimepointExplorer`, etc.

## Scope / open questions to investigate

- [x] How OpenFeature flags are evaluated client-side in a Grafana plugin context (which provider, how the GOFF backend is reached).
- [x] Wire up an OpenFeature provider and expose it behind the existing `useFeatureFlag` hook so consumers don't change.
- [x] Migration path for each existing `FeatureName.*` toggle (and the `?features=` query-param override behavior in `isFeatureEnabled`), or decide whether they can coexist
- [x] Update test helpers (`mockFeatureToggles`) to mock OpenFeature evaluations.
- [ ] Coordinate flag definitions/config in deployment_tools (GOFF) with @grafana-feature-flags.
- [x] Decide rollout/cutover plan and whether toggles and OpenFeature coexist temporarily.

## Notes

- No hard blockers identified so far. The single `useFeatureFlag` abstraction means the consumer-side change should be contained; most of the work is provider wiring + GOFF config + per-flag migration.
[...]

PR #1726, opened 2026-06-12
feat: bootstrap OpenFeature provider behind useFeatureFlag

g3john:
Addresses: https://github.com/grafana/synthetic-monitoring-app/issues/1717

> **This PR doubles as the design doc for the OpenFeature migration.** The code here is a working bootstrap, but treat everything as a proposal open for discussion

## Problem

Grafana is standardizing frontend feature flags on OpenFeature (OFREP evaluation, GOFF configuration in deployment_tools), and legacy toggles are now actively discouraged at deployment_tools review — adding our CALs toggle triggered the deprecation nudge that prompted #1717. Synthetic Monitoring reads all of its flags from the legacy `config.featureToggles` system, so every new flag deepens our investment in a deprecated mechanism, and no flag value can change without a deploy.

## Proposed approach

Migrate gradually, not all at once. The existing `useFeatureFlag` API stays exactly as it is — every component keeps asking the same question the same way. What changes is what answers it: a routing map (`OPEN_FEATURE_KEYS`) behind the hook sends each flag to either OpenFeature or the legacy system, flag by flag. Migrating a flag is a one-line map entry — hosted-grafana auto-mirrors the legacy `features-{env}.libsonnet` definitions into the flag service, so existing flags are already served at the OFREP endpoint under their legacy names with no deployment_tools change. Moving each definition to a proper dotted GOFF flag in `goff/k6/` is a separate, decoupled step. None of the ~20 consumer files or ~25 test files change.

Three phases:

1. **Phase 1 (this PR)** — bootstrap the plumbing with **zero flags routed; behavior is unchanged**.
2. **Phase 2** — per flag: (a) frontend-only PR routing it through OpenFeature against its existing auto-mirrored legacy name, starting with CALs (dev/staging only, zero prod exposure); (b) decoupled deployment_tools PRs moving the definition to a dotted GOFF flag in `goff/k6/` and retiring the legacy entry.
3. **Phase 3** — delete the legacy path once the map covers every flag.

### How the bootstrap works

- An OFREP provider initializes when the app mounts against Grafana's flag endpoint (`/apis/features.grafana.app/v0alpha1/namespaces/<ns>`), registered under a plugin-scoped OpenFeature domain — isolated from Grafana core and other plugins. Same pattern as the SLO, drilldown, IRM, and pathfinder plugins in production.
- Init failures are caught and logged once (e.g. Grafana versions without the features API); flags then fall back to legacy values/defaults.
- Initialization deliberately does NOT happen at plugin preload time: that grew the preloaded `module.js` past danger's critical threshold and would fire OFREP requests on every Grafana page fleet-wide, even when SM is never opened. The preload bundle is unchanged from main.
- Tests run against an in-memory provider; `mockFeatureToggles` drives both backends so existing test call sites work unchanged before and after a flag migrates.
- Local development keeps the familiar `custom.ini` workflow: Grafana's OFREP endpoint is backed by a static provider seeded from the same `[feature_toggles]` ini section, so migrated flags can be toggled locally exactly like legacy ones (add the flag key, restart Grafana).
- The full per-flag checklist lives in `docs/development/openfeature-migration.md`.

## Proposed migration order (Phase 2)

| # | Flag | Legacy rollout | Notes |
|---|---|---|---|
| 1 | `CALs` | dev+staging only | Pilot — triggered the nudge, zero prod exposure |
| 2 | `Folders` | dev+staging only | Most consumers |
| 3 | `Screenshots`, `GRPCChecks` | nowhere | Decide migrate vs delete |
| 4 | `VersionManagement`, `TimepointExplorer` | prod | Route against mirrored legacy name; verify in dev/staging first |
| 5 | `SecretsManagement` | prod | Last — module-scope read in `k6.types.ts` needs surgery first |
| — | `synthetic-monitoring-check-editor` | prod (zombie) | No longer in the app — delete from deployment_tools |

## Additional notes

- **`?features=` URL override: dropped for migrated flags.** OpenFeature flags change at runtime via GOFF without a deploy, which removes most of the override's use case. It keeps working for flags still on the legacy path; an in-memory dev override can be added later if its absence hurts.
- **OSS Grafana keeps flag settability.** OSS/on-prem Grafana (>= 12.x) serves the same OFREP endpoint backed by a static provider seeded from `[feature_toggles]` in a `.ini` file. Operators set migrated flags via ini + restart, same as legacy toggles. What OSS doesn't get is GOFF: runtime changes and rollout targeting are Cloud-only, and GOFF-defined flags are default-off in OSS unless the operator opts in.


docs/development/openfeature-migration.md
```diff
@@ -0,0 +1,114 @@
...
+   `mockFeatureToggles({ [FeatureName.X]: true })` test calls keep working unchanged.
+3. Smoke test in dev/staging: the flag should resolve with the same value as before
+   (`reason: TARGETING_MATCH` from the mirrored definition).
+
+Note: the `?features=<flag>` URL override no longer applies once a flag is routed.
+
+### Step B — move the definition to the new pattern (deployment_tools PRs)
+
+1. Define the dotted key in
+   `ksonnet/environments/hosted-grafana/waves/feature-toggles/goff/k6/{env}.libsonnet`,
```
VikaCep: Shouldn't we have a path for synthetic monitoring instead of k6 here?
`ksonnet/environments/hosted-grafana/waves/feature-toggles/goff/synthetic-monitoring/{env}.libsonnet`

src/components/App.tsx
```diff
@@ -55,17 +56,19 @@ const App = (props: AppRootProps<ProvisioningJsonData>) => {
...
-          <SMDatasourceProvider>
-            <PermissionsContextProvider>
-              <DevTools>
-                <InitialisedRouter />
-              </DevTools>
-            </PermissionsContextProvider>
-            <ReactQueryDevtools />
-          </SMDatasourceProvider>
-        </FeatureFlagProvider>
+        <SmOpenFeatureProvider>
```
VikaCep: For consistency with other providers I'd name it `SMOpenFeatureProvider`

src/hooks/useFeatureFlag.ts
```diff
@@ -1,10 +1,21 @@
...
+// Mapped flags evaluate through OpenFeature; the rest fall through to legacy
+// config.featureToggles. See docs/development/openfeature-migration.md.
 export function useFeatureFlag(featureFlag: FeatureName) {
   const { isFeatureEnabled } = useFeatureFlagContext();
+  const openFeatureKey = OPEN_FEATURE_KEYS[featureFlag];
+  const openFeatureValue = useBooleanFlagValue(openFeatureKey ?? UNMAPPED_FLAG_KEY, false);
+
   return {
-    isEnabled: isFeatureEnabled(featureFlag),
+    isEnabled: openFeatureKey !== undefined ? openFeatureValue : isFeatureEnabled(featureFlag),
```
VikaCep: For mapped flags looks like this can't distinguish between `off` from `provider not ready` which might trigger a flickering effect.
We should consider also returning `isReady`/`isLoading` (from the provider status), trivially `true` for legacy flags, so consumers can distinguish those states for phase B.

docs/development/openfeature-migration.md
```diff
@@ -0,0 +1,114 @@
...
+   ```ts
+   export const OPEN_FEATURE_KEYS: Partial<Record<FeatureName, string>> = {
+     [FeatureName.CALs]: 'synthetic-monitoring-cost-attribution',
+   };
+   ```
+
+   All `useFeatureFlag` / `<FeatureFlag>` consumers flip to OpenFeature atomically. Swap any
+   non-React call sites (direct `isFeatureEnabled` calls) to `getBooleanFlag(...)`. Existing
+   `mockFeatureToggles({ [FeatureName.X]: true })` test calls keep working unchanged.
+3. Smoke test in dev/staging: the flag should resolve with the same value as before
```
VikaCep, line 75: tested this on dev and it didn't work for `synthetic-monitoring-cost-attribution`. Turns out there's a missing comma in `features-dev.libsonnet` for this flag and it makes it fail. It works fine on ops though!

VikaCep (commented):
👏 👏 👏 
Left a few comments but overall great work!

VikaCep (approved):
LGTM! 🔥

PR #1732, opened 2026-06-17
feat: auto-enable Synthetic Monitoring on check-creation deep-links

g3john:
Addresses: https://github.com/grafana/synthetic-monitoring/issues/565

When an uninitialized user lands on a route that signals intent to use Synthetic Monitoring, the plugin now runs the initialization flow automatically (no manual "Get started" click) and returns them to the exact page they intended.

This applies to the section and resource-creation routes:
- `checks` and `checks/new`, `checks/choose-type`
- `probes` and `probes/new`
- `alerts`

The Home page stays actionable (it is not auto-initialized), so we don't install for everyone who simply lands on SM. The previous per-section welcome pages (`ChecksWelcomePage`, `ProbesWelcomePage`, `AlertingWelcomePage`) are removed in favor of the auto-initialize flow.

This ships on by default (no feature flag). If auto-init fails, the user sees an error with a retry button; users lacking permissions get the existing contact-admin guidance.

### Tracking
Adds Rudderstack onboarding events so this can be tracked in dashboards:
- `synthetic-monitoring_onboarding_auto_initialized` on success (with the route)
- `synthetic-monitoring_onboarding_auto_initialize_failed` on failure (with the route and reason)

https://github.com/user-attachments/assets/8ac797b4-9295-420d-9bfd-7de06de01fcb


src/routing/types.ts
```diff
@@ -14,3 +14,7 @@ export enum AppRoutes {
   Redirect = 'redirect',
   Scene = 'scene',
 }
+
+export const AUTO_INITIALIZE_ROUTES = [AppRoutes.NewCheck, AppRoutes.ChooseCheckGroup, AppRoutes.NewProbe] as const;
```
g3john: Are these routes enough? Is it worth adding more? (i.e. /checks, /probes, /alerts etc.)

src/page/AutoInitializeResourceCreation.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Stack } from '@grafana/ui';
+
+import { FeatureName } from 'types';
+import { AppRoutes } from 'routing/types';
+import { getRoute } from 'routing/utils';
+import { useFeatureFlag } from 'hooks/useFeatureFlag';
+import { AppInitializer } from 'components/AppInitializer';
+
+export const AutoInitializeResourceCreation = () => {
+  const { isEnabled: isFeatureEnabled } = useFeatureFlag(FeatureName.AutoInitializeOnUrl);
```
VikaCep: I think we could release this feature without feature flagging it. Would be a clear benefit for everyone.

src/components/AppInitializer.tsx
```diff
@@ -38,7 +40,17 @@ export const AppInitializer = ({ redirectTo, buttonText }: PropsWithChildren<Pro
...
     setDataSouceModalOpen,
-  } = useAppInitializer(redirectTo);
+  } = useAppInitializer(redirectTo, autoInitialize);
+
+  const [autoInitializing, setAutoInitializing] = useState(autoInitialize);
+  const hasAutoInitialized = useRef(false);
+  useEffect(() => {
+    if (autoInitialize && !hasAutoInitialized.current && canReadDs && canInitialize) {
+      hasAutoInitialized.current = true;
+      reportEvent(FaroEvent.AutoInit);
```
VikaCep, line 50: I think we should also add rudderstack events for this so we can easily track it in our dashboards the way we do for other app interactions. We could add a tracking event under `src/features/tracking/` (maybe an onboarding feature) e.g. `auto_initialized` on success with the route it came from, and `auto_initialize_failed` with the reason.

src/routing/types.ts
VikaCep: I saw the [thread](https://raintank-corp.slack.com/archives/C0175SS6SA3/p1781724157542189?thread_ts=1775583841.040999&cid=C0175SS6SA3) where you settled on check-creation deep-links only, it makes sense! That said, I think it'd also be valuable to auto-init on `AppRoutes.Checks`, `AppRoutes.Probes` and `AppRoutes.Alerts`, since landing on those sections seems like a signal the user wants to use them. We could keep the home page actionable so we're not installing for everyone who just lands on SM. What do you think? If we do that, we can just remove the `ChecksWelcomePage`, `ProbesWelcomePage` and `AlertingWelcomePage`

VikaCep (commented):
Awesome work! Left a few comments

src/page/AutoInitializeResourceCreation.tsx
g3john: Sounds good!

src/routing/types.ts
g3john: Yeah that makes sense to me!

VikaCep (commented):
LGTM! I think you're missing a `yarn build:analytics-events` though

VikaCep (approved):
LGTM!

PR #1734, opened 2026-06-18
fix: distinguish multiple notification policy trees in routing preview

### LINKED ISSUE #1733: Alert Routing Summary labels every policy tree's root as "Default policy" when alertingMultiplePolicies is enabled
## Summary
With the `alertingMultiplePolicies` feature enabled (multiple notification policy trees), SM's **Alert Routing Summary** preview lists one row per tree but labels *every* tree's root as **"Default policy"**. This makes it look like there are several duplicate default policies when in fact each row is a distinct, named tree (e.g. `Default policy`, `hashicorp-vault`, `pam-incident-alert`).

## Root cause
Each tree's root node sits at journey `level === 0` with no matchers, so it's always classified as the default policy and given the hard-coded label `"Default policy"`, regardless of which tree it belongs to.

## Expected behaviour
Each tree's root should be labelled with the tree's actual name (e.g. `hashicorp-vault`, `pam-incident-alert`), with only the genuine default tree shown as "Default policy".

## Actual behaviour
All tree roots render as "Default policy", so the summary appears to contain several identical default policies and it's unclear which contact point(s) the alert actually reaches.

## Notes / clarifications to surface in the UI (optional follow-ups)
- "Sent to empty" = matched but no contact point configured (nothing sent). We could instead omit it for clarity.

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring-app/issues/1733

## What this PR does / why we need it

Fixes the per-check **Alert Routing Summary** so it reflects where a synthetic-monitoring alert is *actually* delivered.

SM per-check alert rules are provisioned by the sm-api ([`newAlertRule`](https://github.com/grafana/synthetic-monitoring-api/blob/main/internal/alerts/rules.go#L768-L794)) **without any notification settings**. Because that field is empty, the rule gets no `__grafana_managed_route__` label, and Grafana Alerting routes it through the **default (`user-defined`) tree only** — it never enters additional policy trees created via the `alertingMultiplePolicies` feature.

The preview, however, matched the alert's labels against **every** routing tree returned by `listRoutingTree`. When a stack had extra policy trees, this showed matches (and contact points) in trees the alert can never reach — a false positive. Customers saw the preview claim an alert would route to a custom tree, while in reality it fell through to the default tree's receiver and the custom tree showed 0 instances.

This PR:

- Scopes the preview's matching to the **default tree** .
- Keeps the earlier improvements to `RouteTreeDisplay` (correct per-tree labelling + per-tree route de-duplication) so the component can still render multiple trees **if** SM later supports per-check policy-tree selection.

## Special notes for your reviewer

- No change to alert delivery or backend routing — this is preview/UI only.
- Why the default tree is the only one that applies to SM alerts is verified in Grafana core: a rule reaches a named tree only by carrying `__grafana_managed_route__=<tree>`, which is produced solely by a rule's notification settings (`PolicyRouting`). SM sets none, so all SM alerts use the default tree. Refs: [notifications.go](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/models/notifications.go), [merge.go](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/notifier/merge/merge.go), [Manage multiple notification policies](https://grafana.com/docs/grafana/latest/alerting/configure-notifications/create-notification-policy/#manage-multiple-notification-policies).

### DIFF
--- a/src/components/CheckForm/AlertsPerCheck/alertRoutingUtils.test.ts
+++ b/src/components/CheckForm/AlertsPerCheck/alertRoutingUtils.test.ts
@@ -1,20 +1,19 @@
 import { USER_DEFINED_TREE_NAME } from '@grafana/alerting';
 
+import { routingTreeFactory } from '../../../test/factories/routingTree';
 import { getDefaultRoutingTree } from './alertRoutingUtils';
 
 describe('getDefaultRoutingTree', () => {
-  const userDefinedTree = { metadata: { name: USER_DEFINED_TREE_NAME } };
-  const namedTreeA = { metadata: { name: 'hashicorp-vault' } };
-  const namedTreeB = { metadata: { name: 'pam-incident-alert' } };
+  const userDefinedTree = routingTreeFactory.build({ metadata: { name: USER_DEFINED_TREE_NAME } });
+  const namedTreeA = routingTreeFactory.build({ metadata: { name: 'hashicorp-vault' } });
+  const namedTreeB = routingTreeFactory.build({ metadata: { name: 'pam-incident-alert' } });
 
   it('returns the user-defined tree regardless of its position in the list', () => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const trees = [namedTreeA, userDefinedTree, namedTreeB] as any;
+    const trees = [namedTreeA, userDefinedTree, namedTreeB];
     expect(getDefaultRoutingTree(trees)).toBe(userDefinedTree);
   });
 
   it('returns undefined when there is no user-defined tree', () => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    expect(getDefaultRoutingTree([namedTreeA, namedTreeB] as any)).toBeUndefined();
+    expect(getDefaultRoutingTree([namedTreeA, namedTreeB])).toBeUndefined();
   });
 });
--- a/src/components/CheckForm/AlertsPerCheck/alertRoutingUtils.test.ts
+++ b/src/components/CheckForm/AlertsPerCheck/alertRoutingUtils.test.ts
@@ -1,0 +1,20 @@
+import { USER_DEFINED_TREE_NAME } from '@grafana/alerting';
+
+import { getDefaultRoutingTree } from './alertRoutingUtils';
+
+describe('getDefaultRoutingTree', () => {
+  const userDefinedTree = { metadata: { name: USER_DEFINED_TREE_NAME } };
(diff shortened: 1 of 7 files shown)


src/components/CheckForm/AlertsPerCheck/alertRoutingUtils.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getDefaultRoutingTree } from './alertRoutingUtils';
+
+describe('getDefaultRoutingTree', () => {
+  const userDefinedTree = { metadata: { name: USER_DEFINED_TREE_NAME } };
+  const namedTreeA = { metadata: { name: 'hashicorp-vault' } };
+  const namedTreeB = { metadata: { name: 'pam-incident-alert' } };
+
+  it('returns the user-defined tree regardless of its position in the list', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const trees = [namedTreeA, userDefinedTree, namedTreeB] as any;
```
ckbedwell: I'd rather we have an RoutingTree factory that added the boilerplate to make this type satisfied rather than casting it as here lie dragons when we are just casually casting any and overriding the lint rules.

src/components/CheckForm/AlertsPerCheck/alertRoutingUtils.test.ts
```diff
@@ -0,0 +1,20 @@
...
+    expect(getDefaultRoutingTree(trees)).toBe(userDefinedTree);
+  });
+
+  it('returns undefined when there is no user-defined tree', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    expect(getDefaultRoutingTree([namedTreeA, namedTreeB] as any)).toBeUndefined();
```
ckbedwell: This could use the same factory.

src/test/fixtures/alerting.ts
```diff
@@ -266,6 +266,65 @@ export const MOCK_PROBE_FAILED_ROUTE_MATCH: InstanceMatchResult = {
...
+              receiver: 'PAM Incident Alert',
+            },
+            matchDetails: [],
+            matched: true,
+          },
+        ],
+      },
+    },
+  ],
+} as unknown as InstanceMatchResult;
```
ckbedwell: This could do with a factory too as this could also become a problem in the future.

ckbedwell (changes requested):
Just some factories for the test fixtures so we don't have to do casting :)

ckbedwell (approved):
LGTM!

Issue #1615, opened 2026-03-10
Feat: SM <-> KG integration


VikaCep:
## Description

Following up on the POC in https://github.com/grafana/synthetic-monitoring-app/pull/1612, which explored an initial integration of Synthetic Monitoring (SM) with the Knowledge Graph (KG), there are a few improvements and next steps we identified.

## Next steps

- **Built-in provisioning:** Contribute the rule(s) to the Asserts repositories so they are always active and the KG automatically discovers Synthetic Checks.

- **UX improvements:** Explore integrating the **asserts widget** and **entity explore widget** into the check UI headers.

  This would allow users to understand if a downstream service or infrastructure component is experiencing issues directly from the check view, or quickly navigate to the related entity in the Knowledge Graph.

- Reach out in `#asserts-knowledge-graph-engineers` if help or guidance is needed.

## References

- POC PR: https://github.com/grafana/synthetic-monitoring-app/pull/1612
- Slack conversation: https://raintank-corp.slack.com/archives/C0175SS6SA3/p1773109804407679?thread_ts=1770394450.006799&cid=C0175SS6SA3

PR #1736, opened 2026-06-19
feat: surface Knowledge Graph insights and service linking on checks

VikaCep:
Part of https://github.com/grafana/synthetic-monitoring-app/issues/1615

## Problem

Synthetic Monitoring checks are disconnected from the rest of a user's observability stack. When a check is failing or slow, there is no path inside the product from the check to the service it monitors, that service's dependencies, or its current health. Users have to jump between Synthetic Monitoring and other tools (Application Observability, Kubernetes Monitoring, alerts) to understand why. There is also no way, from within Synthetic Monitoring, to tell the Knowledge Graph which service a check is actually monitoring.

## Solution

This PR adds the Synthetic Monitoring side of the Knowledge Graph (KG) integration, in three places:

**1. Knowledge Graph insights on the check dashboard.** The check page embeds the KG assertions widget (the health rings), scoped to the check's `SyntheticCheck` entity and the current time range. Users get error, latency and health insights for the check inline, without leaving Synthetic Monitoring.

image

**2. Linking a check to a Knowledge Graph service.** The Labels section of the check form gains optional service name and namespace fields. Suggestions are fetched live from the Knowledge Graph, and the values are stored as the `service_name` and `namespace` labels on the check. This is what lets the Knowledge Graph draw the `MONITORED_BY` relationship between the service and the check. A live indicator shows whether the selected pair matches a known Service entity. When these labels are declared as cost attribution labels, both surfaces edit the same values and stay in sync.

image

**3. Connected services graph on the check dashboard.** A "Connected services" section renders the check's Knowledge Graph neighbourhood as an inline graph: the service the check monitors, plus that service's immediate callers and dependencies. Nodes carry the KG insight rings, so an unhealthy neighbour surfaces as a root cause hint right on the check page. Hovering a node shows its insights with a link to the entity in the Knowledge Graph. The data comes from a Cypher query against the KG datasource that walks `MONITORED_BY` to the linked check and one hop of `CALLS` in both directions. The section is expanded by default. When the check is not linked to a service yet, it shows a zero state whose button links straight to the Labels section of the edit form. An "Open in Knowledge Graph" link opens the full entity graph.

image
image

## Feature flag and rollout

All three surfaces render only when both of these are true:

* the Knowledge Graph app (`grafana-asserts-app`) is installed on the stack, and
* the `synthetic-monitoring-knowledge-graph` feature flag is enabled.

The flag exists because the Knowledge Graph app has a large installed base that predates this integration. `SyntheticCheck` entities only start appearing once the entity discovery rules land in the KG platform (the `asserts-adi` side of this project), so without a flag those stacks would see empty sections. With the flag, this PR is safe to merge as soon as it is approved: everything stays dormant until we enable the flag per stack, and the flag also works as a kill switch if anything misbehaves.

## Review guide

The diff is large, but more than half of it has already been reviewed:

* **Connected services graph** (`ConnectedServices.*`, `ConnectedServicesGraph.tsx`, `useKGDS.ts` and the KG test fixtures, about 1,600 lines): reviewed and approved separately in #1774, then merged into this branch unchanged. 
* **Service link in the check form** (`KnowledgeGraphServiceLink.*`, `KnowledgeGraphValueCombobox.tsx`, `knowledgeGraphApi.ts` and the changes to the label form components, about 1,100 lines, roughly half of it tests): this is the main part to review.
* **Insights widget** (`KnowledgeGraphInsights.tsx` plus two lines in each scene dashboard): small.
* **Feature flag** (`knowledgeGraph.hooks.ts`, `types.ts`): the shared gate described above.

## Design doc

This is the Synthetic Monitoring side of the [Design Doc: Synthetic Monitoring × Knowledge Graph Integration (v2)](https://docs.google.com/document/d/1JN8EhTbHWdOPp7zm9fN_kBqJJdCFiuL5n4It-J29MX8/edit?tab=t.52ch9tsbok18#heading=h.2c6v5q7ad1y9), covering work units 2.5, 2.6 and 2.7. The doc is approved.

## Additional context

This PR pairs with the KG platform changes in `asserts-adi` (the `SyntheticCheck` entity type, the entity discovery rules and the `MONITORED_BY` relationship) and the `asserts-app-plugin` changes that add the entity Metrics tab and the link back to Synthetic Monitoring ([#3461](https://github.com/grafana/asserts-app-plugin/pull/3461)). Until those rules are deployed and `SyntheticCheck` entities exist in the graph, everything in this PR stays dormant behind the feature flag.


VikaCep:
bugbot review

VikaCep:
bugbot review

src/features/knowledgeGraph/ConnectedServices.constants.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Insight-ring severity colors, pinned to the Knowledge Graph's own values
+ * (`assertsColors` in the KG plugin) so a check's ring reads identically on both surfaces.
+ */
+export const KG_SEVERITY_COLORS = {
+  critical: '#F2495C',
+  warning: '#FF9830',
+  info: '#5794F2',
+} as const;
```
ckbedwell: Is there a link to their file and line number we can add to this so it is quick to reference?

src/features/knowledgeGraph/ConnectedServices.hooks.ts
```diff
@@ -0,0 +1,75 @@
...
+    refId: 'A',
+    queryType: 'entityGraph',
+    queryMode: 'cypher',
+    cypherQuery: buildServiceNeighbourhoodQuery(checkEntityName),
+  };
+
+  const request: DataQueryRequest = {
+    requestId: 'sm-kg-service-neighbourhood',
+    targets: [query],
+    range: { from, to, raw: { from: 'now-1h', to: 'now' } },
```
ckbedwell: Doesn't this block the ability to do any historic look-up as it is always hard-coded to the last hour -- shouldn't it be synced with the selected time range on the dashboard?

src/features/knowledgeGraph/ConnectedServices.hooks.ts
```diff
@@ -0,0 +1,75 @@
...
+/** The KG datasource's entityGraph query in Cypher mode (see the KG plugin's kgdatasource types). */
+interface KGEntityGraphQuery extends DataQuery {
+  queryType: 'entityGraph';
+  queryMode: 'cypher';
+  cypherQuery: string;
+}
+
+// The KG resolves entities against a time window; one hour matches the convention used by the
+// service-link lookups in knowledgeGraphApi.ts.
+const NEIGHBOURHOOD_WINDOW_MS = 60 * 60 * 1000;
```
ckbedwell: I'd rather this was just ONE_HOUR_IN_MS imported from the utils.constants.ts file and leave a comment citing its a neighbourhood window.

src/features/knowledgeGraph/ConnectedServices.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+  const serviceName = findLabelValue(check.labels ?? [], KG_SERVICE_NAME_LABEL);
+  const namespace = findLabelValue(check.labels ?? [], KG_NAMESPACE_LABEL);
+
+  return (
+    <section className={styles.container} data-testid={CONNECTED_SERVICES_TEST_ID.section}>
+      <div className={styles.header}>
+        <IconButton
+          name={isOpen ? 'angle-down' : 'angle-right'}
+          aria-label={isOpen ? 'Collapse connected services' : 'Expand connected services'}
```
ckbedwell: Small nit: aria-labels should remain constant and just rely on aria-expanded to convey state. It's confusing for screen reader users when both expanded and label values change.

src/features/knowledgeGraph/ConnectedServicesGraph.tsx
ckbedwell, line 1: Are we recreating the graph view here? Do they not expose their own component?

ckbedwell (changes requested):
Few questions and nits -- the main one is looking at the syncing of the dashboard time range with the KG query.

src/features/knowledgeGraph/ConnectedServicesGraph.tsx
VikaCep: Yeah, I [asked](https://raintank-corp.slack.com/archives/C0B9EB3350Q/p1784926540935699?thread_ts=1782934995.822309&cid=C0B9EB3350Q) about this but unfortunately this is not available as an exposed component. Perhaps we can move this out as a shared utility between consumers as Russ mentioned in the conversation.

src/features/knowledgeGraph/ConnectedServices.constants.ts
VikaCep: Done!

ckbedwell (approved):
LFGOOOOO

PR #1739, opened 2026-06-23
fix: gracefully degrade when user lacks folders permissions

VikaCep:
## Problem

The **Grafana Synthetic Monitoring** folder is the default folder used by alerting rules and, with the folders feature flag, also used to group checks. When a user opens Synthetic Monitoring for the first time on a stack where this folder does not yet exist, the app auto-creates it. However, if the user lacks `folders:create` permission, the auto-creation fails and the app hides all checks with no explanation.

Users who permanently lack `folders:read` would also hit this problem without self-resolution, as they cannot access the folder even after it has been created by someone else.

## Solution

In this PR we add a fallback for when the default folder is unavailable. Instead of hiding checks, the app loads the check list without folder grouping and displays an info banner explaining what needs to happen:

- **Folder not provisioned**: tells the user the **Grafana Synthetic Monitoring** folder has not been created yet and that a user with `folders:create` permission needs to open Synthetic Monitoring to initialize it.
- **Missing folders:read**: tells the user they lack the required permission and should contact their organization administrator.

image

image

### DIFF
--- a/src/components/Checkster/components/form/generic/GenericNameValueField.test.tsx
+++ b/src/components/Checkster/components/form/generic/GenericNameValueField.test.tsx
@@ -33,8 +33,7 @@ jest.mock('data/useDefaultFolder', () => ({
   useDefaultFolder: jest.fn(() => ({
     defaultFolder: undefined,
     defaultFolderUid: undefined,
-    isLoading: false,
-    isError: false,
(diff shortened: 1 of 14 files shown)


ckbedwell (approved):
LGTM!

🔥

PR #1742, opened 2026-06-24
fix: allow editing checks when users lack folders:read permissions

VikaCep:
## Problem

Users with `checks:write` but without `folders:read` could not edit their synthetic checks. Clicking **Edit** on an accessible check bounced them straight back to the timepoint explorer.

This affected customers managing permissions via Terraform/custom roles (e.g. teams granted SM check permissions but not folder permissions), who had no easy way to re-grant `folders:read` across every stack. The graceful-degradation fallback shipped in #1739 was meant to cover exactly this case, but the edit page never reached it.

## Solution

The edit page now waits for folder access to finish resolving before it decides whether to redirect. Previously it evaluated the check's read permission while folder data was still loading — and during the optimistic loading window a check whose folder returned `403` briefly looked forbidden, triggering an irreversible redirect out of the edit page before the SM-RBAC fallback could take over.

With the redirect now gated on folder access settling, a user who lacks `folders:read` lands on the editable form (SM-RBAC only) as intended, instead of being sent to the timepoint explorer.

In the recording, I'm loading the app as a user without `folders:read` access, but with `checks:read` and `checks:write`. This allows me to edit a check, and doesn't use the new folders view.

https://github.com/user-attachments/assets/b6c54c66-c798-48a6-a929-ef1cb14f093e

### DIFF
--- a/src/data/useDefaultFolder.ts
+++ b/src/data/useDefaultFolder.ts
@@ -83,6 +83,15 @@ export function useDefaultFolder(enabled = true) {
     },
     staleTime: FOLDERS_STALE_TIME,
     refetchOnWindowFocus: false,
+    // Once this query has settled (success or error) we must not re-fetch it
+    // every time a new observer mounts. With no data and a 403 error, React
+    // Query's `shouldLoadOnMount` is gated by `retryOnMount` (not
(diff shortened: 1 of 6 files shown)


g3john (approved):
LGTM!

PR #1744, opened 2026-06-25
feat: emit event bus event when check created

MattIPv4:

Adds a new operational event fired over the event bus for when checks are created, to support [product marketing work](https://raintank-corp.slack.com/archives/C0AJVPB9S2U/p1781191475571649) (🔐) via the [global messaging feature](https://github.com/grafana/grafana-setupguide-app/blob/main/docs/global-messaging.md) (🔐), to show a toast popup for users who have created a check.

N/A

https://github.com/user-attachments/assets/a94d65ae-869b-4b1c-b4c4-c6500af65a31

### DIFF
--- a/src/data/useChecks.ts
+++ b/src/data/useChecks.ts
@@ -1,5 +1,6 @@
 import { type QueryKey, useMutation, UseMutationResult, useQuery, useSuspenseQuery } from '@tanstack/react-query';
 import { isFetchError } from '@grafana/runtime';
+import { emitCheckCreatedEvent } from 'features/tracking/appEvents';
 import { trackCheckCreated, trackCheckUpdated } from 'features/tracking/checkFormEvents';
 
 import { type MutationProps } from 'data/types';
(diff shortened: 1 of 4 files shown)


VikaCep (changes requested):
The tests are failing because `emitCheckCreatedEvent` calls `getAppEvents().publish()`, but the app event bus is never initialized in Jest, so `getAppEvents()` is `undefined` and the call throws, which rejects the create mutation and skips the alerts request those tests assert on.

Since this is the first time we use the app event bus, I think we should mock it globally in `src/test/mocks/@grafana/runtime.tsx` so it's available everywhere. Something like:

```ts
const appEvents = {
  publish: jest.fn(),
  subscribe: jest.fn(() => ({ unsubscribe: jest.fn() })),
  getStream: jest.fn(),
  removeAllListeners: jest.fn(),
  newScopedBus: jest.fn(),
};

return {
  ...actual,
  getAppEvents: () => appEvents,
  // ...rest
};

MattIPv4:
Ahh, makes sense, hadn't considered runtime wouldn't have a stubbed implementation by default.

VikaCep (approved):
LGTM! 🎉

PR #1749, opened 2026-06-30
feat: Frontend Observability context for the Grafana Assistant on browser-check pages

markjmeier:
## What

Adds a `useDemAssistantContext` hook that registers structured Grafana Assistant **page context** on the pages where users work with browser checks — the home, check-list, and browser check dashboard pages. The context describes how Synthetic Monitoring browser checks (synthetic monitoring) relate to Grafana Frontend Observability (real-user monitoring), includes the user's own browser checks, and points at the Frontend Observability setup entry point.

It is only registered when the user has at least one browser check, and on the single-check dashboard it also flags the currently viewed check.

## Why

Browser checks tell you a synthetic path broke, but not who was affected or what real users experienced. When a user asks the Grafana Assistant about the real-user side of a browser check, it previously had to guess. This gives the Assistant accurate, grounded product context so it can explain how Frontend Observability complements their checks and reference their actual checks.

## How it works

- `src/hooks/useDemAssistantContext.ts` builds a single `structured` context item via `@grafana/assistant` and registers it with `useProvidePageContext`.
- Mounted in `SummaryDashboard` (home), `CheckList` (checks), and `BrowserDashboard` (check dashboard, which passes the focused check).
- Gated on the presence of browser checks — no context is registered otherwise.

## Testing

- Unit tests in `useDemAssistantContext.test.ts` cover the gating, that only browser checks are included, that the setup URL is present, and the focused-check behaviour.
- Manual: on the home or checks page with at least one browser check, open the Assistant and ask about the real-user impact of a browser check — it should reference your checks and link the Frontend Observability setup page. With no browser checks, no context is registered.

## Dependencies

Adds `@grafana/assistant`.

### DIFF
--- a/src/hooks/useDemAssistantContext.ts
+++ b/src/hooks/useDemAssistantContext.ts
@@ -1,0 +1,106 @@
+import { useEffect, useMemo } from 'react';
+import { createAssistantContextItem, useProvidePageContext } from '@grafana/assistant';
+
+import { Check, CheckType } from 'types';
+import { getCheckType } from 'utils';
+import { PLUGIN_URL_PATH } from 'routing/constants';
+
+// Onboarding entry point for Grafana Frontend Observability ("create web app" page).
+const FRONTEND_OBSERVABILITY_SETUP_URL = '/a/grafana-kowalski-app/apps/new';
+
+// The context is registered for any Synthetic Monitoring app page where this hook is
+// mounted (today: the home, check-list, and browser check dashboard pages).
+const URL_PATTERN = `${PLUGIN_URL_PATH}**`;
+
+interface DemAssistantContextOptions {
+  // The browser check the user is currently viewing, when on a single-check page.
+  focusedCheck?: Check;
+}
+
+/**
+ * Provides the Grafana Assistant with product context describing how Synthetic Monitoring
+ * browser checks (synthetic monitoring) relate to Grafana Frontend Observability (real-user
+ * monitoring). This grounds the Assistant so that questions about the real-user side of a
+ * browser check get an accurate, useful answer, and includes the user's browser checks so the
+ * Assistant can reference them and recommend instrumenting the same URLs.
+ *
+ * The context is only registered when the user has at least one browser check, and stays
+ * dormant unless the conversation is about real-user impact.
+ */
+export function useDemAssistantContext(checks: Check[], options: DemAssistantContextOptions = {}) {
+  const { focusedCheck } = options;
+
+  const browserChecks = useMemo(
+    () => checks.filter((check) => getCheckType(check.settings) === CheckType.Browser),
+    [checks]
+  );
+
+  const focusedBrowserCheck =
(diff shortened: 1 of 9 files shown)


markjmeier:
@VikaCep this is to support work with the PMM team. when working on this i was a little confused that it wanted to add grafana/assistant, i thought that was present already with our previous context work. 

Also it authored test coverage for me, so that would be worth looking at.  I tested assistant replies and it provides accurate/factual information

src/scenes/BrowserDashboard/BrowserDashboard.tsx
```diff
@@ -29,6 +30,8 @@ export const BrowserDashboard = ({ check }: { check: Check }) => {
   const styles = useStyles2(getStyles);
   const checkType = getCheckType(check.settings);
 
+  useDemAssistantContext([check], { focusedCheck: check });
```
VikaCep: It looks like we should pass all checks here and keep `focusedCheck` as the current one, to stay consistent with the hook's semantics: `checks` is meant to be the user's full browser-check set, while `focusedCheck` narrows to the one in view. Passing only `[check]` collapses those two into the same thing.

```ts
const { data: checks } = useSuspenseChecks();
useDemAssistantContext(checks, { focusedCheck: check });
```

src/hooks/useDemAssistantContext.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const context = useMemo(() => {
+    if (browserChecks.length === 0) {
+      return [];
+    }
+
+    return [
+      createAssistantContextItem('structured', {
+        title: 'Synthetic Monitoring browser checks & Frontend Observability',
+        bypassLimits: true,
```
VikaCep: `bypassLimits: true` goes against the SDK's guidance [here](https://github.com/grafana/grafana-assistant-app/blob/main/packages/%40grafana/assistant/src/context/base.ts#L13-L28). It says to only bypass limits for fixed-size content (static instructions, bounded metadata) and keep it `false` for anything variable-size. This payload includes `userBrowserChecks`, which grows with the number of browser checks, so I'd remove it if possible.

VikaCep (commented):
This looks great! I just left a few comments plus the following one:

The pages that mount this hook run the real `@grafana/assistant` SDK in their existing test suites. We could add a global mock like the other `@grafana/*` SDKs so tests stay isolated:

`src/test/mocks/@grafana/assistant.tsx`:

```ts
jest.mock('@grafana/assistant', () => ({
  useProvidePageContext: jest.fn(() => jest.fn()),
  createAssistantContextItem: jest.fn((type, params) => ({ type, ...params })),
}));
```

Then add `import 'test/mocks/@grafana/assistant';` to `src/test/jest-setup.tsx` next to the existing mocks.

markjmeier:
@VikaCep requested changes should be addressed!

VikaCep (approved):
LGTM!

PR #1667, opened 2026-04-27
feat: link to Frontend Observability sessions from timepoint viewer

ckbedwell:
Addresses: [grafana/synthetic-monitoring#507](https://github.com/grafana/synthetic-monitoring/issues/507)

Closes [grafana/synthetic-monitoring#509](https://github.com/grafana/synthetic-monitoring/issues/509)
Partially addresses [grafana/synthetic-monitoring#510](https://github.com/grafana/synthetic-monitoring/issues/510) (see *Known limitations* below)

## Problem

When investigating a Synthetic Monitoring browser check via the Timepoint Explorer, the per-execution context lives in two places: the probe-side k6 logs surfaced in the viewer, and the corresponding Frontend Observability session captured by Faro. Today there is no link between them — to inspect the user-side trace (page loads, web vitals, errors) for a specific execution a user has to leave the plugin, open the Frontend Observability app and manually correlate the k6 test run.

## Solution

The Timepoint Viewer's action bar now includes a "View Frontend Observability session" link button when a Faro session can be located for the selected timepoint and probe. The hook issues a Loki query against the configured logs datasource that filters Faro events by `job`, `instance` and `probe` (matched independently against `k6_testRunId`, so the upstream JSON key order doesn't matter), extracts `app_id` and `session_id` from the first matching record, and constructs a relative link to `/a/grafana-kowalski-app/apps/{app_id}/sessions/{session_id}`.

The query only fires for browser checks. If Frontend Observability isn't configured, no matching logs come back, the query errors, or no session is found, the action button is silently omitted — answering [#509's open question](https://github.com/grafana/synthetic-monitoring/issues/509) about whether to show a "no session found" state in favour of a silent omission.

## Known limitations

- **1:N session handling.** A single check execution may visit multiple instrumented apps and therefore produce multiple Faro sessions. This PR surfaces only the first matching session. Tracked as a follow-up under [grafana/synthetic-monitoring#510](https://github.com/grafana/synthetic-monitoring/issues/510).
- **Integration test for the rendered button** is not included; pure utilities are unit-tested. Happy to add one if reviewers want it.

## Screenshots

[Check with FE O11y available](https://dev.grafana-dev.net/a/grafana-synthetic-monitoring-app/checks/9786?from=now-3h&to=now&timezone=browser&var-probe=$__all&var-job=Quick%20pizza%20dev&var-instance=https:%2F%2Fquickpizza.grafana-dev.com%2F)
image

[Check without FE O11y available
[...]

PR #1756, opened 2026-07-09
feat: distinguish missing Faro sessions from uninstrumented checks

ckbedwell:
Follow-up to #1667 / [grafana/synthetic-monitoring#509](https://github.com/grafana/synthetic-monitoring/issues/509).

## Problem

Some browser-check executions for an instrumented app return no matching Frontend Observability session even when other executions for the same check do. A successful empty lookup was treated like “not instrumented,” so the Timepoint Viewer showed **Add RUM to your app** — which is misleading when RUM is already present for the check.

## Solution

Track check-level RUM availability and use it to choose the empty-state copy:

- **Passive:** any successful per-execution Faro session lookup marks the check as RUM-present for the explorer session.
- **Probe:** once loaded timepoints expose `execution_id`s, one capped Loki existence query over those IDs sets present/absent (errors leave availability unknown; `present` is never demoted).

Empty successful lookups then show **No session for this run** when RUM is present, and keep the **Add RUM to your app** docs CTA when it is not. Loading/error handling from the prior work is preserved (spinner while fetching, silent omit on error).

## Test plan

- [ ] Browser check with Faro sessions: open Timepoint Viewer → **View Frontend Session** still works
- [ ] Same check, execution with no session after RUM is known present → muted **No session for this run** (no docs CTA)
- [ ] Browser check with no Faro data at all → **Add RUM to your app** after successful empty lookup
- [ ] Arrow between executions → spinner / no flash of wrong empty state
- [ ] Loki failure on session lookup → button omitted (not Add RUM)

### DIFF
--- a/src/scenes/components/TimepointExplorer/FrontendO11yButton.tsx
+++ b/src/scenes/components/TimepointExplorer/FrontendO11yButton.tsx
@@ -8,12 +8,8 @@ import { getCheckType } from 'utils';
 import { useLogsDS } from 'hooks/useLogsDS';
 import { DocsLink } from 'components/DocsLink';
 import { useTimepointExplorerContext } from 'scenes/components/TimepointExplorer/TimepointExplorer.context';
-import {
-  useSelectedProbeNames,
-  useStatefulTimepoint,
-} from 'scenes/components/TimepointExplorer/TimepointExplorer.hooks';
+import { useStatefulTimepoint } from 'scenes/components/TimepointExplorer/TimepointExplorer.hooks';
 import { StatelessTimepoint } from 'scenes/components/TimepointExplorer/TimepointExplorer.types';
-import { getCouldBePending, getPendingProbeNames } from 'scenes/components/TimepointExplorer/TimepointExplorer.utils';
 import { useFaroSessionLink } from 'scenes/components/TimepointExplorer/TimepointViewerFaroSession.hooks';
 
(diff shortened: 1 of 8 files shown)


src/scenes/components/TimepointExplorer/FrontendO11yButton.tsx
```diff
@@ -52,81 +43,104 @@ export const FrontendO11yButton = ({ timepoint }: { timepoint: StatelessTimepoin
...
+  if (!executionId) {
     return null;
   }
 
-  if (canQueryFaro && (isFaroLoading || !isFaroFetched)) {
-    return <Spinner />;
+  // Prefer a spinner over flashing empty/Add-RUM states, and over briefly
+  // showing the previous execution's session while the new lookup is in flight.
+  if (isFaroInFlight) {
+    return <Tooltip content="Looking for Frontend Observability session..."><Spinner /></Tooltip>;
```
g3john: The tooltip isn't showing up for me when I hover over the spinner. Apparently `Tooltip` attaches a ref + hover handlers, but `Spinner` doesn't forward refs, so the tooltip never opens. What helps is wrapping the `Spinner` with a `span`
```suggestion
    return <Tooltip content="Looking for Frontend Observability session..."><span><Spinner /></span></Tooltip>;
```

g3john (dismissed):
LGTM! Just noticed one thing

PR #1757, opened 2026-07-10
fix: attach default folder UID to POST

g3john:

The Synthetic Monitoring app creates its default Grafana folder using only the folder title, allowing Grafana to generate a random UID.

The SM API alert ruler expects the same folder to use the canonical `grafana-synthetic-monitoring-app` UID. If alert synchronization runs later, it may create a second folder with the same title but the canonical UID.

Once both folders exist, the app selects the canonical folder tree. Checks assigned to the older random-UID folder can then disappear from the checks list even though their folder still exists and is accessible.

Addresses grafana/support-escalations#23263


VikaCep (approved):
LGTM!

PR #1758, opened 2026-07-10
feat: add Grafana Cloud org_id and stack_id to all tracking events

markjmeier:

Campaign and analytics tooling can't attribute our RudderStack events to a
Grafana Cloud org: `context_traits_org_id` carries the instance-internal org id,
which is set at the platform level and can't be
fixed in this plugin. Marketing needs the grafana.com org identity as event
properties to build cohorts (e.g. the SM → Frontend Observability campaign).

This adds two properties to **every** event created through the tracking
factory (`createSMEventFactory`), without touching any event definition:

- `org_id` — the grafana.com org id
- `stack_id` — the grafana.com stack id

Both are sourced from the SM tenant API (`TenantResponse.orgId` / `stackId`) —
not from `config.bootData.user.orgId`, which is the instance-internal value.
A new null-rendering `TrackingIdentity` component (mounted inside
`SMDatasourceProvider`) fetches the tenant via the existing `useTenant()` hook
and registers the ids as factory-level base properties once resolved.

Behavioral notes:

- **Async-safe:** events fired before the tenant resolves — or when there is no
  cloud tenant (local dev, uninitialized plugin, tenant fetch failure) — omit
  the properties entirely rather than reporting undefined values.
- **No contract changes:** existing event names and properties are unchanged;
  downstream BigQuery tables and campaign triggers keying off them are
  unaffected. The new properties are snake_cased in source (`org_id`, not
  `orgId`) so the BigQuery column names are exact and don't depend on
  RudderStack's key transformation.
- The grafana.com org **slug** is not available client-side, so it is
  deliberately not included rather than substituting an instance-internal or
  stack-level value. (The stack slug is derivable from `config.appUrl` if ever
  needed as a separate property.)

n/a — supports Grafana Cloud org attribution for marketing campaign tooling

- New `utils.test.ts` covers the factory: event-name composition, base
  properties present on sample/props-less events, cleanly absent when cloud
  identity isn't available, undefined values dropped, and event props winning
  over base props on key collision.
- `useTenant()` is react-query cached, so this adds at most one extra
  `/sm/tenant` request per session (it's already fetched by
  `useLegacyAlertsRestriction` in many flows).
- No UI changes, so no screenshots.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -5,6 +5,7 @@ import { AppRootProps } from '@grafana/data';
 import { getWebInstrumentations, initializeFaro } from '@grafana/faro-web-sdk';
 import { config, getAppPluginVersion } from '@grafana/runtime';
 import { css, Global } from '@emotion/react';
+import { TrackingIdentity } from 'features/tracking/TrackingIdentity';
 
 import { ProvisioningJsonData } from 'types';
 import { getFaroConfig } from 'faro';
@@ -63,6 +64,7 @@ const App = (props: AppRootProps<ProvisioningJsonData>) => {
             <SMDatasourceProvider>
               <PermissionsContextProvider>
                 <AssistantContext />
+                <TrackingIdentity />
                 <DevTools>
                   <InitialisedRouter />
                 </DevTools>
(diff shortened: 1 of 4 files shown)


VikaCep (approved):
LGTM

PR #1759, opened 2026-07-13
fix: show checks in folders outside the default folder

VikaCep:
Addresses: https://github.com/grafana/support-escalations/issues/23263

## Problem

Checks assigned to a folder that lives outside the default Synthetic Monitoring folder's subtree disappeared from the checks list, even though the folder still existed and the user could access it. The checks stayed reachable by direct link, so from the user's perspective their checks had silently vanished from the list with no explanation.

This surfaced in production when a tenant ended up with a second folder named "Grafana Synthetic Monitoring" under a non-canonical UID: the app loaded the canonical folder's tree and every check still pointing at the older folder was stranded and hidden. The same gap affected any check assigned to an arbitrary folder outside the default subtree (for example via the API or Terraform).

## Solution

The checks list now hides a check only when its folder is genuinely forbidden (a 403 from the folder permissions endpoint). Folders that exist and are readable but sit outside the default folder's subtree are treated as visible.

In the folder view these "external" folders are surfaced as top-level nodes with their real title and an "Outside default folder" badge that explains the folder is separate from the default Synthetic Monitoring folder (including its UID to disambiguate same-named folders) and prompts the user to move its checks. Folder-level permissions still apply to these checks, so nothing is granted beyond what the user already has.

image

### DIFF
--- a/src/data/folderPermissions.ts
+++ b/src/data/folderPermissions.ts
@@ -1,4 +1,4 @@
-import { Check } from 'types';
+import { Check, GrafanaFolder } from 'types';
 
 import { SMPermissions } from './permissions';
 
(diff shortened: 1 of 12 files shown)


g3john (dismissed):
LGTM!

g3john (approved):
Nice!

PR #1762, opened 2026-07-17
feat(e2e): validate pull requests against dem-dev

ckbedwell:
## Problem

Frontend changes in this plugin currently lack a realistic, reproducible pull-request check against the Synthetic Monitoring API, agent-backed scenario data, and the Grafana plugin runtime. When regressions appear, reproducing CI locally also requires translating workflow YAML into manual setup steps, which slows down iteration on dashboard and check-list behavior.

## Solution

This change adds a closed-loop dem-dev Playwright harness that builds the plugin, seeds a deterministic historical scenario through dem-dev, and exercises the UI in the browser. A local `scripts/dem-dev-e2e.sh` bridge mirrors the CI lifecycle so developers can configure, seed, test, and tear down the same stack independently.

The harness includes reusable dem-dev fixtures, historical read/write and hybrid journeys, feature-flag profile support, pinned runtime revisions, CI workflows with diagnostics and artifact upload on failure, and documentation for local setup and troubleshooting.

## Test plan

- [ ] CI `dem-dev-e2e` workflow passes on this branch
- [ ] `yarn typecheck:e2e`
- [ ] Local smoke: `yarn e2e:dem:doctor`, `yarn e2e:dem:test:historical-read`
- [ ] `shellcheck scripts/dem-dev-e2e.sh`

## Notes

This branch has been rebased onto current `main` and force-pushed to replace the earlier pre-release base. Follow-up PR `feat/check-dashboard-parity-contract` adds the legacy HTTP dashboard parity contract on top of this branch.

PR #1763, opened 2026-07-20
Experiment: prove dem-dev E2E catches a breaking API

ckbedwell:
## What this demonstrates

This stacked experiment pins the dem-dev E2E environment to the deliberately incompatible Synthetic Monitoring API commit from grafana/synthetic-monitoring-api#2337.

The API still passes its own build, lint, unit-test, and container-build matrix, but it returns `410 Gone` from the legacy check-update route used by the app.

## Expected CI result

This PR is intentionally expected to remain red:

- the plugin build should pass
- dem-dev startup and deterministic scenario seeding should pass
- historical read journeys should pass
- the hybrid handoff journey should pass
- the historical write journey should fail through the UI after clicking **Disable check**, because the card never reaches the **Disabled** state
- Playwright should upload its HTML report, screenshots, videos, and trace

The test does not inspect HTTP methods, URLs, requests, or responses. It detects the regression only through user-visible UI behavior.

## Relationship to the other PRs

- Base: #1762 — the green app-owned dem-dev E2E framework
- Breaking upstream API: grafana/synthetic-monitoring-api#2337

Experiment only — do not merge.


ckbedwell:
Proved POC, closing for now. Will revisit.

PR #1768, opened 2026-07-21
feat: Add Label Migration config tab for label_ prefix removal

The-9880:
Closes https://github.com/grafana/synthetic-monitoring/issues/543

## What — a config tab for migrating off the `label_` prefix

The agent prefixes user-defined check/probe labels with `label_` before they reach metrics and logs. grafana/synthetic-monitoring-api#2307 adds the API control plane for a tenant-driven migration to un-prefixed labels; this PR adds the UI that drives it: a new **Label Migration** tab under Config.

Two gates control it: the `synthetic-monitoring-label-migration` feature flag decides whether the tab (nav entry and route) exists on a stack, and `isAdmin` decides who can act — non-admins on an enabled stack see the current state with a contact-admin notice instead of action buttons.

## The mental model

A tenant is in one of three modes (`GET /api/v1/tenant/label-mode`). The tab renders one state per mode, and each state's single action is the transition the API allows from there:

| Mode | `env=prod` appears as | Tab shows |
|---|---|---|
| `PREFIXED` (0, default) | `label_env` | explanation + **Enable dual-write** |
| `DUAL_WRITE` (1) | `label_env` **and** `env` | migration guidance + **Finalize migration** |
| `UNPREFIXED` (2) | `env` | success + **Revert to dual-write** |

The state machine's reversibility is encoded in the UI:

- **Enable dual-write is irreversible** — the confirm modal says so, and there is deliberately no "revert to prefixed" action anywhere in the tab.
- **Finalize ↔ revert to dual-write** is the reversible pair: a tenant who finalizes before all their LBAC rules/alerts/dashboards are migrated can temporarily restore the prefixed form.
- Every transition goes through a `ConfirmModal`. When the API's collision pre-check rejects enabling dual-write (409), the tab renders the `collidingLabels` list with instructions to rename those labels and retry.

Supporting UI in all states:

- **Live series preview** — fetches one real `probe_success` series from the tenant's metrics datasource and shows how system + user labels look in the current mode (constructed example as fallback when there's no live data).
- **Reserved system label names** — collapsible list rendered from the API's `systemLabels` response, so the app never hardcodes the reserved set.

## How to review — layers first, then the review-response rounds as one diff

The first five commits are the feature, one layer each:

1. **`feat(datasource)`** — `getLabelMode`/`setLabelMode` on `SMDataSource`, plus MSW handlers and fixtures.
2. **`feat(ui)`** — the `LabelMigrationTab` component: per-mode states, confirm modals, collision error rendering, reserved-names list; component tests.
3. **`feat(routing)`** — nav entry and route registration.
4. **`feat(ui)`** — the live `probe_success` series preview.
5. **`fix(ui)`** — alignment with the final API state machine: the branch predates the reversibility flip (irreversible dual-write entry, revertible finalization), so copy/buttons/tests are updated, and the reserved-set fixture is refreshed to the 27 names shipped in agent v0.60.0.

**Everything after that (`bbce32f8..50a613f3`) is review response**:

- Data fetching migrated to react-query (`data/useLabelMode.ts`: query + mutation; the preview is a `useQuery`), replacing hand-rolled fetch state.
- Gating: the `synthetic-monitoring-label-migration` feature flag on nav + route; `isAdmin` handled in-tab with `ContactAdminAlert` (Secrets-tab pattern; no more 404 on non-admin deep-links).
- Double-submit protection via the app's async `ConfirmModal`; error alerts carry the API's `msg` (FetchError shape), render next to the actions that trigger them, and persist until replaced by a new attempt's outcome or dismissed.
- Preview correctness: live keys classified against the API's `systemLabels`, bounded `topk(1, …)` query, and distinct loading/empty/failed/no-datasource states.
- Corrected UNPREFIXED copy: reserved names are rejected at write time by the API (agent-side dropping is the backstop), so the tab no longer tells migrated tenants to audit for silent drops.
- Tests: 5 → 23, including all three transition flows end-to-end (with PUT body assertions), modal cancel, collision-alert persistence, non-admin rendering, and the preview state matrix.

## Follow Ups

- Wire the 409 collision alert to the bulk rename endpoint (`POST /api/v1/check/labels/rename`) in a separate PR; for now the alert instructs manual renaming.


src/page/ConfigPageLayout/tabs/LabelMigrationTab/LabelMigrationTab.tsx
g3john, line 1: This file's quite large. Let's break the Main component and Sub components into their own files

src/test/handlers/tenants.ts
```diff
@@ -64,3 +65,25 @@ export const getTenantCostAttributionLabels: ApiEntry<ListTenantCostAttributionL
...
+  route: `/sm/tenant/label-mode`,
+  method: `get`,
+  result: () => {
+    return {
+      json: TENANT_LABEL_MODE,
+    };
+  },
+};
+
+export const setLabelMode: ApiEntry<{ mode: number; systemLabels: string[] }> = {
```
g3john: Instead of `mode: number`, let's reuse `LabelMode`

src/page/ConfigPageLayout/tabs/LabelMigrationTab.tsx
```diff
@@ -0,0 +1,571 @@
...
+  // explicitly, so cancelling the modal doesn't discard rename guidance.
+  const openConfirm = (targetMode: number, title: string, body: string, confirmText = 'Confirm') => {
+    setConfirmModal({ isOpen: true, targetMode, title, body, confirmText });
+  };
+
+  return (
+    <ConfigContent title="Label Migration" loading={isLoading} ariaLoadingLabel="Loading label mode">
+      {!isAdmin && <ContactAdminAlert title="Contact your administrator to change the label migration mode" />}
+
+      {Boolean(loadError) && !state && (
```
g3john: We can just double exclamation to assert
```suggestion
      {!!loadError && !state && (
```

src/page/ConfigPageLayout/tabs/LabelMigrationTab.tsx
```diff
@@ -0,0 +1,571 @@
...
+    font-family: ${theme.typography.fontFamilyMonospace};
+    font-size: ${theme.typography.bodySmall.fontSize};
+  `,
+  tagDimmed: css`
+    white-space: break-spaces;
+    overflow-wrap: anywhere;
+    font-family: ${theme.typography.fontFamilyMonospace};
+    font-size: ${theme.typography.bodySmall.fontSize};
+    opacity: 0.45;
+  `,
```
g3john: These are practically the same with the exception of the `opacity`. Let's just trim `tagDimmed` down to just the `opacity`, and then in the `className` of `LabelTag` have it be conditionally added based on dimmed: `className={cx(styles.tag, { [styles.dimmed]: dimmed })}`
```suggestion
  tag: css`
    white-space: break-spaces;
    overflow-wrap: anywhere;
    font-family: ${theme.typography.fontFamilyMonospace};
    font-size: ${theme.typography.bodySmall.fontSize};
  `,
  dimmed: css`
    opacity: 0.45;
  `,
```

src/test/handlers/tenants.ts
```diff
@@ -64,3 +65,25 @@ export const getTenantCostAttributionLabels: ApiEntry<ListTenantCostAttributionL
     };
   },
 };
+
+export const getLabelMode: ApiEntry<{ mode: number; systemLabels: string[] }> = {
```
g3john: Instead of `mode: number`, let's reuse `LabelMode`

src/datasource/DataSource.ts
```diff
@@ -377,6 +377,22 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
     });
   }
 
+  //--------------------------------------------------------------------------------
+  // LABEL MODE
+
+  async getLabelMode() {
+    return this.fetchAPI<{ mode: number; systemLabels: string[] }>(
```
g3john: Instead of `mode: number`, let's reuse `LabelMode`

src/datasource/DataSource.ts
```diff
@@ -377,6 +377,22 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
...
+      `${this.instanceSettings.url}/sm/tenant/label-mode`
+    );
+  }
+
+  async setLabelMode(mode: number) {
+    return this.fetchAPI<{ mode: number; systemLabels: string[] }>(
```
g3john: Instead of `mode: number`, let's reuse `LabelMode`

src/page/ConfigPageLayout/ConfigPageLayout.tsx
```diff
@@ -59,6 +59,17 @@ export function ConfigPageLayout() {
       ],
     };
 
+    // Label Migration is feature-flagged for rollout. The tab itself limits
+    // mode changes to admins and shows a contact-admin notice otherwise.
+    if (isFeatureEnabled(FeatureName.LabelMigration)) {
+      navModel.children!.push({
+        icon: 'tag-alt',
+        text: 'Label Migration',
```
g3john: Let's sentence case to match the other tab "Access tokens"
```suggestion
        text: 'Label migration',
```

src/test/handlers/index.ts
```diff
@@ -58,11 +60,13 @@ const API_ROUTES = {
   getRangeMetrics,
   getSecret,
   getSMDS,
+  getLabelMode,
```
g3john: nit: Let's move this up to maintain the alphabetical ordering

src/test/handlers/index.ts
```diff
@@ -58,11 +60,13 @@ const API_ROUTES = {
...
   getTenant,
   getTenantCostAttributionLabels,
   getTenantLimits,
   getTenantSettings,
   installPlugin,
+  setLabelMode,
```
g3john: nit: Let's move this down to maintain the alphabetical ordering

g3john (commented):
Looking great! Couple things I noticed during my testing that's outside the scope of this PR so not a blocker, but we should address at some point. Both on the create/edit a check page:
1. Tested the reserved name enforcement and it works, the API rejects it properly. But the user just sees "Save failed / bad request". The real message (label "xyz" is reserved by the system...) is in `data.msg`, and `FormRoot.tsx:117-119` reads `data.message`, so it gets dropped
2. The check form still shows a hardcoded label_ prefix on the label input, with a tooltip saying the prefix is what prevents system label conflicts. Nothing outside this tab is mode aware, so it feels like a missing piece of the migration rather than something for this PR. Probably worth checking for other hardcoded `label_` assumptions too

Image

g3john (approved):
LGTM!

PR #1771, opened 2026-07-23
fix: show secrets side-panel tab for HTTP checks

mem:
Add CheckType.Http to SECRETS_CHECK_COMPATIBILITY so the Secrets feature tab appears in the check editor for HTTP checks, in addition to Browser and Scripted. HTTP credential fields (basic auth password, bearer token, TLS cert/key) already opt into secret references via allowSecrets, so no further wiring is needed.

Add tests for the previously-uncovered tab-filtering logic in FeatureTabsContext, asserting the Secrets tab is shown for compatible check types when the SecretsManagement flag is on and hidden otherwise.

Fixes #


VikaCep:
bugbot review

VikaCep (approved):
LGTM

PR #1772, opened 2026-07-24
fix: remove stale secrets datasource proxy routes

mem:
The datasource proxy declared GET/PUT/POST/DELETE routes for api/v1alpha1/secrets, pointing at the Synthetic Monitoring API. These date back to the "intermediate secrets management" work and are no longer used: the app now reads and writes secrets directly through Grafana's App Platform secrets API (secret.grafana.app) via SecretsManagerClient added in 4af0d71ef, never through the datasource proxy. No TypeScript code references these routes.

Drop the dead route declarations.

Fixes #

### DIFF
--- a/src/datasource/plugin.json
+++ b/src/datasource/plugin.json
@@ -76,54 +76,6 @@
           "content": "Bearer {{.SecureJsonData.accessToken}}"
         }
       ]
-    },
-    {
-      "path": "api/v1alpha1/secrets",
-      "method": "GET",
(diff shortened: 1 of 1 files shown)


VikaCep:
bugbot review

VikaCep (approved):
LGTM

mem:
I need to hold this one until I get confirmation from a customer that they no longer need something or that they have switched to a different solution.

ckbedwell:
@mem do you know if we have that confirmation yet?

PR #1774, opened 2026-07-24
feat: add Connected services graph to the check dashboard

VikaCep:
## Problem

When a synthetic check fails, users can't tell from the check dashboard where the monitored service sits in their system — which services call it and which services it depends on. That context lives in the Knowledge Graph's entity graph, so triaging "is it my service or something downstream?" means leaving Synthetic Monitoring and manually exploring the graph in another app.

## Solution

The check dashboard gets a new collapsible **Connected services** section that renders the check's one-hop neighbourhood from the Knowledge Graph: the check itself, the service it monitors, and that service's upstream callers and downstream dependencies, fetched via the KG datasource's Cypher query API.

Rather than the generic node graph panel, the graph is a SVG renderer  that mirrors the Knowledge Graph's own entity-graph design:

- Layered top-to-bottom layout: check + upstream callers → monitored service → downstream dependencies.
- KG node anatomy: insight severity rings (using the KG's colors), insight count badges, entity icons, and the Synthetic Monitoring logo on check nodes.
- Orthogonal elbow edges between rows; calls between services in the same row render as arcs over the row so they don't cut through neighbouring nodes.
- Hovering a node opens a popup with the entity's active insights and an "Open in Knowledge Graph" deep link (nodes no longer navigate away on click).
- Hovering an edge highlights it, fades the others, and names the connection (e.g. `checkout → cart`) — useful when sibling arcs overlap in busy rows.

The section handles all datasource states: loading, query errors with a retry action, an empty result while the KG is still discovering the check, and a zero state when the Knowledge Graph plugin isn't installed.

## Testing

Integration tests cover the graph rendering from Cypher result frames, the node insights popup and its deep link, edge hover highlighting, and the empty/error/zero states. Unit tests cover frame parsing, ring segment math, the layered layout (including same-row arcs), and the entity URL builders. Verified visually against a live stack with the OTel demo services.

image

https://github.com/user-attachments/assets/87b9da3f-7113-46ee-af52-28a43b779d63

### DIFF
--- a/src/features/knowledgeGraph/ConnectedServices.hooks.ts
+++ b/src/features/knowledgeGraph/ConnectedServices.hooks.ts
@@ -1,0 +1,73 @@
+import { useQuery } from '@tanstack/react-query';
+import { CoreApp, DataQueryRequest, DataQueryResponse, dateTime, LoadingState } from '@grafana/data';
+import { getDataSourceSrv } from '@grafana/runtime';
+import { DataQuery } from '@grafana/schema';
+import { isObservable, lastValueFrom } from 'rxjs';
+
+import { Check } from 'types';
+import { useKGDS } from 'hooks/useKGDS';
+
+import { buildServiceNeighbourhoodQuery, parseGraphFrames, ServiceNeighbourhood } from './ConnectedServices.utils';
+import { getSyntheticCheckEntityName } from './knowledgeGraph';
+
+/** The KG datasource's entityGraph query in Cypher mode (see the KG plugin's kgdatasource types). */
+interface KGEntityGraphQuery extends DataQuery {
+  queryType: 'entityGraph';
+  queryMode: 'cypher';
+  cypherQuery: string;
+}
+
+// The KG resolves entities against a time window; one hour matches the convention used by the
+// service-link lookups in knowledgeGraphApi.ts.
+const NEIGHBOURHOOD_WINDOW_MS = 60 * 60 * 1000;
+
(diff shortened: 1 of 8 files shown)


src/features/knowledgeGraph/ConnectedServices.hooks.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export function useServiceNeighbourhood(check: Check) {
+  const kgDS = useKGDS();
+  const checkEntityName = getSyntheticCheckEntityName(check);
+
+  return useQuery({
+    queryKey: ['kg-service-neighbourhood', kgDS?.uid, checkEntityName],
+    enabled: Boolean(kgDS),
+    queryFn: () => fetchServiceNeighbourhood(kgDS!.uid, checkEntityName),
+  });
```
g3john, line 74: Would it be worth setting the `refetchInterval` to match the rest of the dashboard? By default the query doesn't become stale during the session
```suggestion
  return useQuery({
    queryKey: ['kg-service-neighbourhood', kgDS?.uid, checkEntityName],
    enabled: Boolean(kgDS),
    queryFn: () => fetchServiceNeighbourhood(kgDS!.uid, checkEntityName),
    refetchInterval: STANDARD_REFRESH_INTERVAL,
  });
```

g3john (approved):
LGTM! Just one comment

src/features/knowledgeGraph/ConnectedServices.hooks.ts
VikaCep: Good call — added in ebd5d1e0. The graph now refetches on `STANDARD_REFRESH_INTERVAL` (60s), same as the other dashboard queries.

PR #1777, opened 2026-07-31
feat: add folder dashboard page with per-folder health overview

markjmeier:
## Problem

Folders let users organize checks into groups that mean something to them — a team's services, an environment, a customer. But today a folder is only a collapsible grouping in the check list: there is nothing that tells the user how the folder is doing *as a whole*. Users who structure their checks around teams or customers have no page to glance at (or keep open) that answers "is this group healthy, and if not, what broke and when?"

## Solution

Every folder now has its own dashboard, linked from a "View dashboard" action on each folder header in the check list, at `checks/folders/:uid`. The page is an opinionated, auto-refreshing overview of the folder's last 3 hours:

- **Header** with a clickable folder path and a "N need attention" badge when any check is down or has a firing alert.
- **KPI tiles**: checks up/down, average reachability, failed executions in the window (naming the most frequent offender), and estimated executions per month.
- **Execution swimlane**: one lane per check showing every execution as a tick — failures are highlighted and clickable, opening the check's dashboard with the time range centred on that execution.
- **Check table**: attention-ordered rows with state, firing-alert badges, reachability, latency, and a latency trend sparkline; linked to each check's dashboard.
- **Recent failures**: the latest failed executions with time, check, and probe.

Checks are matched to their metrics and logs exactly (by job + target pair, client-side — the same pattern the check list uses), so the page stays correct even when job names are reused across folders. Checks in subfolders are included.

The v1 view is deliberately fixed to "last 3 hours" with no time picker or ad-hoc filters: every section tells the same story, and per-check dashboards carry the full controls. If user feedback asks for time travel here, it can be added deliberately.

## Notes

- The page registers Grafana Assistant context (`sm-folder-dashboard`) with starter questions ("Why is this folder unhealthy?", "Compare the checks in this folder").
- A `folder_dashboard_viewed` tracking event (with check count) is included to measure adoption.
- The shared `PluginPage` test mock now renders `subTitle` so the breadcrumb is testable.
- Known follow-ups (intentionally out of scope): failure *reasons* in the recent-failures table (needs per-check-type log parsing), execution deep-links via the `execution_id` log label, "SLOs in this folder", and folder-level cost attribution stats.

## Screenshot

image


markjmeier:
bugbot run

markjmeier:
Addressed all four bot findings in a07a6d6e:

- **CodeQL (double-escaping)**: rewrote the LogQL value escaping as a single-pass transform so no character is ever re-escaped — correctness no longer depends on replace ordering.
- **Bugbot (analytics docs)**: ran `yarn build:analytics-events`; `folder_dashboard_viewed` is now in the generated docs.
- **Bugbot (healthy KPI while loading)**: the Checks tile now shows a loading state and a neutral "no state data yet" state instead of prematurely reporting green.
- **Bugbot (tracking fires once)**: the dashboard is now keyed by folder uid, so navigating between folders remounts it — the view event fires per folder with the right check count. This also fixes an adjacent unreported bug: the execution swimlane's time window was captured on mount and would have leaked from one folder to the next.

markjmeier:
## Open questions for discussion

Collected from testing against live folders and early feedback (@markjmeier + design/eng input welcome — tagging topics by who we most want to hear from).

### Design (@vesnadean )

1. **Should the executions swimlane and the check table merge into one section?** They overlap on the check-name column. A merged, KG-style row (name | state | reachability | latency | execution timeline) would shorten the page and remove the duplication — at the cost of lanes no longer sharing a strict visual baseline for cross-check comparison. A merged row would also force the swimlane bucketing work (see 5).
2. **KPI row density** — we're at five tiles (checks up, lowest reachability, firing alerts, failed executions, executions/month). Too many? "Firing alerts" + "Failed executions" could merge into a single "needs attention" tile.
3. **Alert link destination** — the firing-alerts tile links to Grafana's unified alerting list filtered to firing SM rules. That's deep in Grafana. Proposal: adaptive link — one check alerting → that check's dashboard; multiple → the filtered list. Right pattern?

### Product scope (@VikaCep )

4. **Is the fixed "last 3 hours" window right?** It matches the check list's stats and suits the live-board persona. The likely first complaint is the morning glance ("what happened overnight?"). If that materializes, we'd propose an opinionated 3h/24h toggle — not a free time picker. OK to ship fixed-3h and let the feedback widget arbitrate?
5. **What folder size should v1 officially support?** Display is bounded (attention-ordered top 25 + "show all"), but data fetching is not: one Loki query covers all the folder's checks (a 300-check folder ≈ 100k log lines per window). Planned iteration: bucketed swimlane rendering + bounded fetching. Do we gate anything on that, or ship and watch?
6. **Page boundary: observe vs manage.** Position: the check list (folder view) is where you *manage*; this page is where you *observe* — so no bulk actions, edit, or delete here, ever. One debatable exception later: acknowledge/silence from the attention surface during incidents. Agree with the boundary?
7. **Failure *reasons*.** The recent-failures table shows when/where/how long, not *why*. The raw error lines exist in Loki; surfacing them needs per-check-type parsing (reusing `parseCheckLogs`) plus grouping ("all 503s from EU-West since 21:14"). It's also a natural Grafana Assistant surface — the page already registers assistant context. Priority?

### Platform asks (worth tickets?)

8. **Folder label on metrics.** Everything folder-scoped in this PR works around the fact that metrics carry no folder dimension. A `grafana_folder` label on all SM metrics (matching the alerting convention) + `folder_uid` on `sm_check_info` would enable native folder filtering in dashboards/Explore, folder-scoped alert routing, and folder-level cost attribution. Connects to the planned `label_` prefix removal work.
9. **Execution deep-links.** Failure marks currently link to the check dashboard with a last-3h range. Newer agents stamp an `execution_id` on log lines — a check-dashboard URL param that scrolls the Timepoint Explorer to that exact execution would complete the triage loop.

markjmeier:
bugbot run

markjmeier:
Second Bugbot round addressed in 4007690f:

- **Premature healthy failures KPI**: the tile's detail now shows "loading execution history" instead of "no failures in window" while logs are in flight — same treatment as the Checks tile.
- **Worst offender ignores check target**: `getWorstOffender` now aggregates failures by the job + target pair (the check identity used everywhere else on the page) so reused job names can't merge into a wrong offender.
- **Disabled checks inflate monthly KPI**: `getExecutionsPerMonth` now skips disabled checks.

(The three older comments — analytics docs, Checks-tile loading state, once-only view tracking — were fixed in a07a6d6e/earlier commits; leaving them for Bugbot to re-verify.)

markjmeier:
bugbot run

VikaCep:
image

I've pushed some changes to this PR, summary of what changed:

- **Refactored to standard @grafana/ui components**: the check table and recent failures now use `InteractiveTable` with pagination, the latency trend uses Grafana's own `Sparkline`, and the section/KPI cards are built from `Box`/`Text` primitives via a shared `FolderSection` — replacing the hand-rolled tables, SVG plot, and card CSS.
- **Moved the PromQL into `src/queries/`** (`upStateByCheck`, `avgLatencyByCheck`, `latencyTrendByCheck`) for consistency with how the rest of the app organises queries, plus a shared, tested regex-escaping util now used by both the PromQL and LogQL builders.
- **Scoped the latency-trend query to the folder's checks** — it was a range query over every check in the stack, which wouldn't have scaled to bigger customers.
- **Graceful failure for the execution-history query**: if the Loki query fails (e.g. a very large folder exceeding query limits), the swimlane now shows an error instead of rendering every lane as "no executions", and the failed-executions KPI shows "unavailable" instead of a false zero.
- **UI fixes for laptop-size screens**: the KPI detail text wrapped badly at narrow tile widths (links breaking mid-phrase).
- **Accessibility**: the swimlane marks are hover-only, so each lane now carries a screen-reader summary ("N executions, M failed in window"); small aria fixes on the sparkline.
- **Moved the components from `src/scenes/Folder/` to `src/page/FolderPage/`** — this page isn't a @grafana/scenes dashboard like its former siblings, so colocating with the page matches our conventions better.
- **Firing-alerts tile wording**: "2 checks alerting" linked to the alerting *rule list*, which was misleading — the link now reads "view rules". (Touches your design question 3: I like the adaptive-link idea as an iteration.)

On the open questions:

4. **Fixed 3h window** — I think it's ok to ship fixed and let the feedback widget arbitrate. If the "what happened overnight?" complaint materialises, an opinionated 3h/24h toggle sounds like the right first iteration.
5. **Folder size** — With the error state above we now fail visibly rather than silently, which I think is enough not to gate on the bucketing/bounded-fetching work. I agree let's keep that as the planned iteration.
6. **Observe vs manage** — agree with the boundary as stated
7. **Failure reasons** — valuable, but perhaps for a next iteration. I'd rather ship and learn whether users actually ask "why" on this page or jump straight into the check dashboard. The assistant-context angle makes it cheaper to revisit later.

VikaCep:
bugbot run

vesnadean:
@markjmeier 

> Should the executions swimlane and the check table merge into one section? They overlap on the check-name column. A merged, KG-style row (name | state | reachability | latency | execution timeline) would shorten the page and remove the duplication — at the cost of lanes no longer sharing a strict visual baseline for cross-check comparison. A merged row would also force the swimlane bucketing work (see 5).

You can merge it, but then you will have a lot of columns and I worry that the executions viz would become unusable if there is a large number of executions or a large time range. In case you want to try to merge it, I propose the following:

- Link the check name to the dashboard, so there is no separate column for the "View dashboard" link.
- Move check type labels below check name
- Think about getting rid od the trend column. What does it bring to the table?

However, I am leaning towards keeping them separate considering they serve two different purposes – one is an overview of executions, and the other is a more general checks health overview.

> KPI row density — we're at five tiles (checks up, lowest reachability, firing alerts, failed executions, executions/month). Too many? "Firing alerts" + "Failed executions" could merge into a single "needs attention" tile.

I don't think 5 tiles is too many, but I don't have a strong opinion on this. Feels like either approach would work fine. 

> Alert link destination — the firing-alerts tile links to Grafana's unified alerting list filtered to firing SM rules. That's deep in Grafana. Proposal: adaptive link — one check alerting → that check's dashboard; multiple → the filtered list. Right pattern?

I would not use adaptive links to change destinations because this can be confusing. I propose always redirecting to a unified alerting list, even if that list consists of only one rule.

VikaCep:
I've made some changes to account for the bugbot review plus Vesna's feedback on

> - Link the check name to the dashboard, so there is no separate column for the "View dashboard" link.
> - Move check type labels below check name

image

VikaCep:
bugbot review

markjmeier:
Addressed Bugbot's stale-probe finding on the state query: the `last_over_time` window now sizes to 1.5× the slowest check frequency (floor 5m) instead of 2× (floor 15m) — wide enough to contain every active probe's latest sample, narrow enough that a probe that stops reporting drops out of the state calculation after roughly one missed cycle rather than holding a stale success for the full window.

Residual trade-off documented in the query's comment: a probe silent for up to ~1.5 intervals after a success can still briefly mask other probes' failures. The complete fix would be per-probe staleness masking or deriving state from the execution log stream the swimlane already loads — leaving that as a design call for this PR's owners.

(Not touching the two Aug 6 findings on the refactored trend query and KPI detail — those are on the recent eng commits, deferring to their authors.)

ckbedwell:
Going to close this for the time-being and we'll circle back to how to implement this once we have the new nav in.

PR #1778, opened 2026-07-31
feat: add cost attribution discovery nudges with click-through tracking

markjmeier:
## Problem

Cost attribution for Synthetic Monitoring recently went GA: checks can carry cost attribution labels (CALs) and spend can be tracked per team/service in the Cost Management and Billing (CMAB) app. Adoption has been slow — not because the feature falls short, but because it's hard to discover. A tenant admin has to already know the feature exists and go set it up in a different app; nothing in Synthetic Monitoring surfaces it, even in the moments where users are clearly thinking about labels or usage.

## Solution

This PR adds three discovery nudges, all self-hiding once a tenant has CALs configured. Note that cost attribution is GA and its `synthetic-monitoring-cost-attribution` flag is enabled for everyone — the nudges reuse the existing gate purely for consistency with the rest of the CAL surface (which is still flag-gated in code). Removing the now-redundant flag repo-wide is deliberately out of scope here and left as separate cleanup debt.

1. **Check list banner** — a dismissible info banner ("Attribute check costs to teams and services") linking to CMAB settings. It only appears for tenants with 5+ checks and no CALs configured, so small/new tenants aren't nagged before cost attribution is worth their while. Dismissal persists via localStorage (same pattern as the private probes alert).
2. **Inline hint in the check form's labels step** — users configuring labels are at a high-intent moment for this feature. The hint explains what CALs unlock, and the link opens in a new tab (so the unsaved check isn't lost) with a note that setup requires an admin role — we're deliberately not sending users somewhere cold.
3. **Tooltips on usage figures** — the "active series" and "executions / month" figures on check list items and the check form's usage estimate are cost proxies; anyone reading them without cost attribution configured is a likely adopter. They get an interactive tooltip linking to CMAB settings.

Click-throughs are tracked via a new `cost_attribution` event family (`cmab_link_clicked` with a `source` property per placement, plus `setup_banner_shown`/`setup_banner_dismissed` for banner CTR) — see the regenerated analytics docs.

### Testing it locally

Under `yarn dev:msw`, append `?features=synthetic-monitoring-cost-attribution&sm-demo-no-cals` to the plugin URL. The second param is a dev-only toggle added to the MSW `/sm/tenant/cals` handler that simulates the "no CALs configured" state (the fixture otherwise returns configured labels, which hides all three nudges). It has no effect in tests or production — happy to drop it if we'd rather not keep demo hooks in handlers.

### Open questions for reviewers

- `setup_banner_shown` fires on every check list visit until the banner is dismissed or CALs are configured. Useful for CTR math, but flagging the volume for whoever owns the analytics pipeline — happy to gate it (e.g. once per session) if preferred.
- The banner threshold (`MIN_CHECKS_FOR_CAL_BANNER = 5`) is a first guess — open to tuning.

### Screenshots

On checks page: 
image

tooltip on estimated executions:
image

hint during check creation on label step:
image


markjmeier:
Few things:
- I dont love the banner on the checks page, that takes up valuable real estate.  It might be okay for the short term, to ensure users can find the feature and set it up. 
- while workin with claude, we still have the feature flag present for the feature, we probably need to clean that up since its not GA.

src/components/Checkster/components/form/layouts/GenericLabelContent.tsx
```diff
@@ -1,118 +1,56 @@
...
-
-    const remainingLabels = currentLabels.filter(
-      (label) => !calNameSet.has(label.name) && !staleCalNameSet.has(label.name)
-    );
-
-    setValue('calLabels', calRows);
-    setValue('labels', remainingLabels);
-
-    prevCalNamesRef.current = calNames;
-  }, [calNames, getValues, setValue]);
```
ckbedwell, line 73: It was driving me mad that the GenericLabelContent component was responsible for CALs -- it made the rest of the PR difficult to reason with. Ideally, this should have been a separate refactor to make review easier but it made my life harder so I chose to put the burden on the reviewer -- such as life now in the agentic age of the SDLC

src/components/CostAttribution/CostAttributionUsageTooltip.tsx
ckbedwell, line 1: I don't love this because it has a11y issues (links in tooltips) but I'm not solving this in this PR.

ckbedwell (commented):
Self-review few comments about my changes

src/page/CheckList/CheckList.costAttributionBanner.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+    });
+
+    it('does not show the banner when CALs are already configured', async () => {
+      mockCalNames([`Team`, `Service`]);
+      await renderCheckList(buildChecks(5));
+
+      expect(screen.queryByText(BANNER_TITLE)).not.toBeInTheDocument();
+    });
+
+    it('hides the banner permanently when dismissed', async () => {
```
g3john: Ran this locally: the banner comes back after a remount (the X doesn't write to localStorage), but this test passes anyway because the assertion runs before the CALs query resolves on the second render, so it's green either way. Was this meant to click the "Understood, don't show again" button instead of the X? That would match the "permanently" in the name. Either way, let's make the remount checks here and in the test below wait for the CALs request to finish before asserting the banner is gone, otherwise they pass no matter what

src/components/CostAttribution/CostAttributionBanner.tsx
```diff
@@ -0,0 +1,79 @@
...
+            size="sm"
+            href={CMAB_URLS.settings}
+            onClick={() => trackCmabLinkClicked({ source: 'check_list_banner' })}
+          >
+            Set up cost attribution
+          </LinkButton>
+          <Button
+            variant="secondary"
+            size="sm"
+            onClick={() => {
```
g3john, line 68: `trackSetupBannerDismissed` only fires on the X, so we never track the "don't show again" clicks. Since the two dismissals are intentionally different, let's track both and tell them apart, something like `trackSetupBannerDismissed({ permanent: true })` here and `{ permanent: false }` in `onRemove` (+ regenerate the analytics docs)

src/components/CheckUsage.tsx
```diff
@@ -1,92 +1,213 @@
...
+            <Icon className={styles.footerIcon} name="info-circle" size="sm" />
+            <span>
+              {hideTelemetry ? USAGE_NUDGE_MESSAGE.withoutTelemetry : USAGE_NUDGE_MESSAGE.withTelemetry}{' '}
+              <TextLink
+                href={CMAB_URLS.settings}
+                external={true}
+                variant="bodySmall"
+                onClick={() =>
+                  trackCmabLinkClicked({
+                    source: 'check_form_usage_tooltip',
```
g3john: This one's a footer now but the event still says `check_form_usage_tooltip`. Worth renaming to something like `check_form_usage_footer` before the data starts coming in? Also `'check_form'` in `CostAttributionUsageTooltip`'s `source` prop isn't used anymore, we can drop it

src/components/Checkster/hooks/useHydrateCalLabels.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (calNames.length === 0) {
+      return;
+    }
+
+    const { calLabels, labels } = formMethods.getValues();
+    const partitioned = partitionCalLabels([...calLabels, ...labels], calNames);
+
+    formMethods.setValue('calLabels', partitioned.calLabels, { shouldDirty: false });
+    formMethods.setValue('labels', partitioned.labels, { shouldDirty: false });
+  }, [calNames, calNamesKey, defaultFormValues, formMethods]);
```
g3john: nit: `calNames` is already in the deps so `calNamesKey` never does anything on its own, one of them can go

g3john (commented):
Few comments from testing this locally

src/page/CheckList/CheckList.costAttributionBanner.test.tsx
VikaCep: Good catch — the X is session-only by design ('don't show again' is the permanent one), so this test was asserting the wrong thing and passing vacuously. Fixed in 44e6eb00: this test now asserts the banner comes back after a remount, and the permanent-dismiss test below waits for the CALs request to resolve on the remount before asserting the banner stays gone.

src/components/CostAttribution/CostAttributionBanner.tsx
VikaCep: Done in 44e6eb00 — both dismissals now fire trackSetupBannerDismissed, distinguished with permanent: true/false, and the analytics docs are regenerated.

src/components/CheckUsage.tsx
VikaCep: Renamed to check_form_usage_footer in 44e6eb00 (no data has shipped yet since the feature is still flagged off). Also dropped the unused 'check_form' source from CostAttributionUsageTooltip.

src/components/Checkster/hooks/useHydrateCalLabels.ts
VikaCep: Removed calNamesKey in 44e6eb00 — calNames is memoized in ChecksterContext so the array reference in the deps is stable.

src/components/CostAttribution/CostAttribution.hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+export function useCostAttributionSetupStatus() {
+  const { isEnabled: isCALsEnabled } = useFeatureFlag(FeatureName.CALs);
+  const { data, isLoading } = useTenantCostAttributionLabels();
+  const calNames = isCALsEnabled ? (data?.names ?? []) : [];
+
+  return {
+    isCALsEnabled,
+    calNames,
+    // Never true while loading, so nudges don't flash for tenants that have CALs configured.
+    needsSetup: isCALsEnabled && !isLoading && calNames.length === 0,
```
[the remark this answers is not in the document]
VikaCep: Fixed in 6e62e20a — the nudges now require the CALs query to have resolved successfully; a failed fetch (or disabled flag) no longer reads as "no labels configured".

src/components/Checkster/contexts/ChecksterContext.tsx
```diff
@@ -121,7 +138,13 @@ export function ChecksterProvider({
...
   const isNew = !check || !check.id;
 
-  const { schema, defaultFormValues } = useFormValuesMeta(checkType, check, probesWithMetadata, defaultFolderUid);
+  const { schema, defaultFormValues } = useFormValuesMeta(
+    checkType,
+    check,
+    probesWithMetadata,
+    calNames,
+    defaultFolderUid
+  );
```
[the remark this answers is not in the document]
VikaCep: Fixed in 19c5efd2 — CAL names no longer feed useFormValuesMeta, so defaultFormValues stays stable when the CALs query resolves. The repartitioning happens in useHydrateCalLabels against current form state without touching other fields.

src/components/CheckUsage.tsx
```diff
@@ -1,92 +1,208 @@
...
+              >
+                Attribute check costs to teams and services
+              </TextLink>
+            </span>
+          </div>
         )}
       </div>
     </div>
   );
 };
```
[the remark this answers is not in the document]
VikaCep: Fixed in 19c5efd2 — the footer nudge is no longer gated on billed telemetry rows, and scripted/multihttp/browser checks get execution-based copy instead.

src/components/CheckUsage.tsx
```diff
@@ -1,92 +1,208 @@
...
+            {billedTelemetryRows.map((row) => (
+              <UsageRow key={row.key} row={row} value={values[row.key]} />
+            ))}
+          </div>
+        )}
+        {showCmabNudge && (
+          <div className={styles.footer}>
+            <Icon className={styles.footerIcon} name="info-circle" size="sm" />
+            <span>
+              {hideTelemetry ? USAGE_NUDGE_MESSAGE.withoutTelemetry : USAGE_NUDGE_MESSAGE.withTelemetry}{' '}
```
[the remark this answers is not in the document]
VikaCep: Fixed in 8a4eac27 — the click handler now reports executions_per_month for check types that hide billed telemetry.

src/components/CostAttribution/CostAttributionBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <LinkButton
+            size="sm"
+            href={CMAB_URLS.settings}
+            onClick={() => trackCmabLinkClicked({ source: 'check_list_banner' })}
+          >
+            Set up cost attribution
+          </LinkButton>
+          <Button variant="secondary" size="sm" onClick={() => setDismissed(true)}>
+            Understood, don&apos;t show again
+          </Button>
```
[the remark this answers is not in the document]
VikaCep: Fixed in 8a4eac27 — the permanent dismiss now also updates the in-memory flag so the banner hides immediately.

src/components/Checkster/hooks/useHydrateCalLabels.ts
```diff
@@ -0,0 +1,24 @@
...
+      return;
+    }
+
+    const { calLabels, labels } = formMethods.getValues();
+    const partitioned = partitionCalLabels([...calLabels, ...labels], calNames);
+
+    formMethods.setValue('calLabels', partitioned.calLabels, { shouldDirty: false });
+    formMethods.setValue('labels', partitioned.labels, { shouldDirty: false });
+  }, [calNames, defaultFormValues, formMethods]);
+}
```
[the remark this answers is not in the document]
VikaCep: Confirmed and fixed in 0bec33f4 — hydration now keeps its last result as a last-priority value source when re-partitioning, so CAL values survive a defaults reset even when labels is dirty. User-cleared values stay cleared since those rows are dirty and take priority. Added a regression test (verified it fails against the previous hook).

src/components/Checkster/hooks/useHydrateCalLabels.ts
```diff
@@ -0,0 +1,34 @@
...
+    // Hydrated rows are written with shouldDirty: false so the form stays pristine, which means
+    // a defaults reset (probes resolving, folder status arriving) wipes them — and when `labels`
+    // is dirty, its CAL-valued rows are not restored either. The check's original labels still
+    // hold those values, so they are the last-priority value source. Rows already in the form
+    // win: a cleared CAL value is a dirty row that survives resets, so it stays cleared.
+    const { calLabels: hydrated } = partitionCalLabels(
+      [...calLabels, ...labels, ...defaultFormValues.labels],
+      calNames
+    );
+    const { labels: remaining } = partitionCalLabels(labels, calNames);
```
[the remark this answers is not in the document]
VikaCep: Fixed in aba0bf84 — the reconciliation is now a pure function (`rehydrateCalLabels`) whose value sources include the previous hydration's rows, so a value edited before partition survives a later defaults reset. Covered by the "preserves a CAL value edited before hydration across a defaults reset" test in `useHydrateCalLabels.test.tsx`.

src/components/Checkster/hooks/useHydrateCalLabels.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    formMethods.setValue('calLabels', hydrated, { shouldDirty: false });
+    formMethods.setValue('labels', remaining, { shouldDirty: false });
```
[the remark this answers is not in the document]
VikaCep: Fixed in aba0bf84 — when a name drops out of the tenant's CAL list, its row now moves back into the custom labels instead of being dropped from the check. Covered by the "moves a deconfigured CAL value back to the custom labels" test in `useHydrateCalLabels.test.tsx`.

g3john (approved):
LGTM!

PR #1782, opened 2026-08-04
chore(faro): duplicate rudderstack events to faro and enhance faro instrumentation

eskirk:
hi folks — we (Frontend Observability) are doing a pass over grafana plugins to improve faro dogfooding visibility. three small commits, each independently droppable:

1. **mirror reportInteraction events into faro** — a tiny echo backend (registered via the public `registerEchoBackend` API from `@grafana/runtime`) that forwards this plugin's `synthetic-monitoring_*` interaction events to your existing faro instance as events. rudderstack is unaffected; the same events just become visible next to your errors, web vitals, and traces in Frontend Observability. same pattern as grafana/logs-drilldown#2023.

2. **faro config hygiene** — adds a `beforeSend` filter so only signals from this plugin's pages are sent (`isolate: true` was already set 👍), and enables `experimental.trackNavigation` for navigation events (quality-of-life when exploring sessions in frontend o11y).

3. **five new events** — probe lifecycle (`probe_created` / `probe_updated` / `probe_deleted` / `probe_token_reset`, fired in the `useProbes` mutation onSuccess callbacks next to the existing faro `reportEvent`s) and `terraform_config_viewed`. these follow your `createSMEventFactory` convention. this commit is deliberately last so it's easy to drop if you'd rather pick your own events.

verified: lint, typecheck, and the probe/config-page test suites all pass. happy to adjust anything — feel free to push to the branch.

### user journey dogfooding

mirroring these interactions into faro also unlocks critical user journeys in frontend observability — journeys are built from faro events, so the duplicated interaction events can be used directly as journey steps to measure completion and drop-off of real product flows. (this app's faro apps are registered against the us-east-0 ops collector rather than the stack backing the frontend o11y team's app list, so I couldn't resolve a critical-journeys link — it lives under the app's "critical journeys" tab wherever your faro apps are managed.)

### DIFF
--- a/docs/analytics/analytics-events.md
+++ b/docs/analytics/analytics-events.md
@@ -344,6 +344,24 @@ Tracks when an alert is deleted successfully
 | ---- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------- |
 | name | `"ProbeFailedExecutionsTooHigh" \| "TLSTargetCertificateCloseToExpiring" \| "HTTPRequestDurationTooHighAvg" \| "PingRequestDurationTooHighAvg" \| "DNSRequestDurationTooHighAvg"` | The name of the alert |
 
+### probes
+
+#### synthetic-monitoring_probes_probe_created
+
+Tracks when a private probe is created.
+
+#### synthetic-monitoring_probes_probe_updated
+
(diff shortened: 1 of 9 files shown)


ckbedwell (approved):
LGTM!

PR #1785, opened 2026-08-04
Mem/proxy datasources

ckbedwell:
Draft PR to see the diff and as a place for discussion of the pros and cons of turning into a backend plugin.


pkg/plugin/namedqueries.go
```diff
@@ -0,0 +1,549 @@
...
+			}, nil
+		},
+	},
+
+	// src/queries/probeExecutionStats.ts -- tenant-wide, so no check parameters
+	"probe_execution_rate": {
+		target: targetMetrics,
+		build: func(params) (built, error) {
+			return built{
+				expr:    fmt.Sprintf(`sum(rate(probe_all_success_count[%s])) by (probe)`, defaultQueryFromTime),
```
ckbedwell, line 199: These two hardcode `defaultQueryFromTime` (3h) instead of using the request's time range, so a caller asking for the last hour gets a 3-hour rate regardless. Mirroring the frontend constant makes sense for the panels that use it, but as a backend API this is now a query endpoint that accepts a time range and silently ignores it for two of its 19 queries — not something a caller can work out from the outside.

Take the window from the request range, or put it in the name (`probe_execution_rate_3h`) so it's explicit?

pkg/plugin/query.go
```diff
@@ -0,0 +1,118 @@
...
+			Expr:          b.expr,
+			Datasource:    ds,
+			Range:         !b.instant,
+			Instant:       b.instant,
+			Interval:      b.interval,
+			LegendFormat:  b.legendFormat,
+			IntervalMs:    q.Interval.Milliseconds(),
+			MaxDataPoints: maxDataPoints(b, q),
+		})
+		ranges[ds] = q.TimeRange
```
ckbedwell, line 51: This is assigned per query inside the loop, so when several queries group onto the same target the last one's range wins silently. The comment above acknowledges the SDK-versus-`/api/ds/query` mismatch, but last-wins means two queries with different ranges on the same target return one answer computed over the wrong window, with nothing signalling it.

`/api/ds/query` genuinely does impose one range per request, so the options are probably to split the group per distinct range or reject mixed ranges outright. Either beats silent.

pkg/plugin/grafana.go
```diff
@@ -0,0 +1,130 @@
...
+	"strings"
+	"time"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana-plugin-sdk-go/config"
+)
+
+// POC: Grafana's own address. Inside the dev container this is Grafana itself.
+// GrafanaConfig.AppURL() is preferred and tried first; this is the fallback.
+const devAppURL = "http://localhost:3000"
```
ckbedwell, line 19: **Security — localhost fallback**

This should hard-fail outside dev rather than silently falling back to localhost.

src/datasource/plugin.json
```diff
@@ -3,6 +3,19 @@
...
   "id": "synthetic-monitoring-datasource",
   "metrics": true,
+  "logs": true,
+  "backend": true,
+  "executable": "gpx_sm_datasource",
+
+  "iam": {
+    "permissions": [
+      { "action": "datasources:read", "scope": "datasources:*" },
+      { "action": "datasources:query", "scope": "datasources:*" },
```
ckbedwell, line 13: **Security — iam blast radius**

`datasources:query` on `datasources:*` means the plugin SA can query any datasource in the org; `authorize()` is the only guard. Worth calling out in the PR description.

Can we scope iam to the linked metrics/logs UIDs, or is `*` required by the platform? Either way, authz is load-bearing — suggest non-admin test coverage before prod rollout.

pkg/plugin/grafana.go
```diff
@@ -0,0 +1,130 @@
...
+// to `//api/ds/query`, which Grafana answers with the SPA's HTML rather than the
+// API.
+func appURL(ctx context.Context) string {
+	if cfg := config.GrafanaConfigFromContext(ctx); cfg != nil {
+		if url, err := cfg.AppURL(); err == nil && url != "" {
+			return strings.TrimSuffix(url, "/")
+		}
+	}
+
+	return devAppURL
```
ckbedwell, line 114: **Security — localhost fallback** (see also `devAppURL` const ~line 19)

Should hard-fail outside dev rather than silently falling back to localhost. Cloud should always provide `AppURL()` via plugin context.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/instancemgmt"
+)
+
+// ID must match the id in src/datasource/plugin.json.
+const ID = "synthetic-monitoring-datasource"
+
+// linkedDatasource is a Prometheus or Loki datasource the SM datasource was
+// configured against. Mirrors LinkedDatasourceInfo in src/datasource/types.ts.
+type linkedDatasource struct {
```
ckbedwell, line 25: **Linked datasource `type`**

Backend requires both `uid` and `type`. Cloud init via `useAppInitializer` writes `type` correctly for normal tenants. Provisioning paths that POST jsonData without `type` will fail late — worth auditing non-init provisioning templates.

src/datasource/types.ts
```diff
@@ -5,6 +5,64 @@ export enum QueryType {
...
+  ScriptedDataSent = 'scripted_data_sent',
+  SumDurationByProbe = 'sum_duration_by_probe',
+}
+
+/**
+ * Query types served by the Go backend rather than in the browser.
+ *
+ * The backend also serves avg_request_latency, avg_request_success_rate,
+ * avg_request_expected_response and scripted_http_requests_error_rate, which the
+ * app does not use: the panels consuming them need Grafana's joined-table frame
```
ckbedwell, line 31: **Hybrid query path — blocking follow-up**

The four unwired registry entries aren't optional long-term. `ResultsByTargetTable` still joins three Prometheus table queries in the frontend (`findValueByName`, `(1 - failureRate) * 100`).

Need composite backend queries (e.g. `request_breakdown_by_target`) that join and transform server-side, returning a single ready-to-render frame. Goal: zero frontend query transforms and no direct Prometheus datasource usage from scenes.

src/scenes/Common/useVizPanelMenu.tsx
```diff
@@ -9,21 +9,70 @@ interface UseVizPanelMenuProps {
...
+ * The expression a named query resolved to, as reported by the datasource that ran
+ * it. Both datasources put this in frame metadata:
+ *
+ *   Prometheus: 'Expr: max by () (max_over_time(probe_success{...}[60s]))\nStep: 1m0s'
+ *   Loki:       'Expr: {probe=~".*", job="test"} | logfmt'
+ *
+ * so the expression runs to the end of the string or to the next `Detail:` line.
+ *
+ * Exported for testing.
+ */
```
ckbedwell, line 34: **Explore link fragility**

Smoke-tested: works — opens Prometheus Explore with the recovered expression. Parsing `executedQueryString` via regex is fragile if Prom/Loki metadata format changes. Accepted trade-off for now; worth a code comment noting the dependency.

pkg/plugin/authz.go
```diff
@@ -0,0 +1,210 @@
...
+
+// permissionsSearchResponse maps identity -> action -> scopes.
+type permissionsSearchResponse map[string]map[string][]string
+
+// authorize reports an error unless the request's user may query the datasource.
+//
+// Fails closed: any inability to establish permission -- no user on the request,
+// no plugin credential, a lookup failure -- denies the query. Falling back to the
+// plugin's own identity is what we are trying to avoid.
+func (a *authorizer) authorize(ctx context.Context, appURL, token string, user *backend.User, ds linkedDatasource) error {
```
ckbedwell, line 61: **Authz not exercised locally**

Anonymous Admin in dev makes authz look like a no-op; anonymous sessions without a user identity caused stat panels to show N/A (`"no user on the request"`).

Before Cloud prod: test with a non-admin user who can use SM but **cannot** query Prometheus directly.

pkg/plugin/namedqueries.go
```diff
@@ -0,0 +1,549 @@
...
+		target: targetMetrics,
+		build: func(p params) (built, error) {
+			job, instance, probe, err := p.check()
+			if err != nil {
+				return built{}, err
+			}
+
+			return built{
+				expr: fmt.Sprintf(
+					`group by(frequency, config_version) (max_over_time(sm_check_info{job="%s", instance="%s", probe=~"%s"}[$__range]))`,
```
ckbedwell, line 141: **Macro dependency**

Registry entries emit `$__range` / `$__rate_interval` — fine when Grafana executes the query, but expressions aren't self-contained if a consumer receives the string (gcx, CallResource). Flag for external-client work.

CONTRIBUTING.md
```diff
@@ -88,6 +89,102 @@ During the initialization process, the Synthetic Monitoring backend will:
...
+#### Grafana settings it needs
+
+The backend queries Prometheus and Loki by calling Grafana's own `/api/ds/query`,
+which is what makes each datasource's existing authentication and any label
+policy attached to its token apply unchanged. To do that it needs its own service
+account, which Grafana provisions from the `iam` block in the datasource's
+`plugin.json`. That requires two settings:
+
+```ini
+[auth]
```
ckbedwell, line 130: **Local dev gap**

Documents service-account settings here, but `dev/custom.ini` doesn't set `managed_service_accounts_enabled` or `externalServiceAccounts`. Consider adding to dev Docker env so `yarn server` works for named queries out of the box.

pkg/plugin/query.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/log"
+)
+
+// QueryData resolves each named query and forwards it to the datasource that
+// holds the data.
+//
+// Queries are grouped by backing datasource so that a request mixing metrics and
+// logs costs one round trip per datasource rather than one per query.
+func (d *Datasource) QueryData(ctx context.Context, req *backend.QueryDataRequest) (*backend.QueryDataResponse, error) {
```
ckbedwell, line 17: **Performance note**

Named queries: browser → backend → authz (up to 3 API calls) → `/api/ds/query` → Prom/Loki. Worth measuring on Cloud with TimepointExplorer. Authz caching per user/datasource is an obvious follow-up if latency shows up.

ckbedwell (commented):
Works well — tested locally (dem-dev) and on a deployed Cloud stack. Haven't tried it with different RBAC roles yet, that needs a look.

What I care about is how `gcx` and other third-party callers consume this long term. Right now gcx rebuilds these queries itself ([status.go](https://github.com/grafana/gcx/blob/main/internal/providers/synth/checks/status.go)), so this PR kills real duplication — but if gcx just calls the named queries it ends up hardcoding the 19 names and the `params` fields instead.

[`schemabuilder`](https://github.com/grafana/grafana-plugin-sdk-go/tree/main/experimental/schemabuilder) from the plugin-sdk would avoid that: one test generates `schema/v0alpha1/query.types.json` from the Go types and the file ships with the plugin ([grafana/grafana#122557](https://github.com/grafana/grafana/pull/122557)). No feature toggles needed to read it — `/public/plugins/grafana-testdata-datasource/schema/v0alpha1/query.types.json` returns a `QueryTypeDefinitionList` on my stack today. gcx could read that and hardcode nothing.

The catch is `params`. With all 19 queries sharing one struct the generated schema is 19 copies of "12 optional fields", which tells a caller nothing — `checks_uptime` needing job/instance/frequency is only expressed in Go. There are 8 real shapes in there (base `check()`, +frequency, +metric, +label, +labelName/value/method, +unsuccessfulOnly, the web vital one, none) so splitting isn't much work.

My other issue is that we are 100% not going to get the API right from the start and I haven't worked out a clear route that supports v1/named_query and v2/named_query simultaneously. Grafana's documentation seems to just imply 'get it right at the start and make everything backwards compatible...' 😅😅😅 

-----
Forgot to mention all the comments below are just reminders for myself that our agentic friends flagged as I was working through the PR.

Pokom:
> [schemabuilder](https://github.com/grafana/grafana-plugin-sdk-go/tree/main/experimental/schemabuilder) from the plugin-sdk would avoid that: one test generates schema/v0alpha1/query.types.json from the Go types and the file ships with the plugin (https://github.com/grafana/grafana/pull/122557). No feature toggles needed to read it — /public/plugins/grafana-testdata-datasource/schema/v0alpha1/query.types.json returns a QueryTypeDefinitionList on my stack today. gcx could read that and hardcode nothing.

Check out the latest push I made, which I codified this recommendation. I've attached a sample response. I think that helps define the shape of the query, but an agent likely needs one a set of examples as well. The one drag I see with this is that(at least as far as I can tell), we'll need to commit the generated json blob as code(🤮). This isn't a huge deal breaker, and maybe even something we could generate on the fly before building the bundle.

I've been working through using claude + gcx to gather the examples + query type definitions and use it to build queries. I'll try and post a video tomorrow demo'ing it. 

[query.types.json](https://github.com/user-attachments/files/31045106/query.types.json)

ckbedwell:
POC verified. Closing as we won't be using this branch.

Issue #1791, opened 2026-08-07
Cost Attribution labels are re-rendered under Custom Labels after save


lewisporterdev:
**What happened**:

When a Synthetic Monitoring check has both Cost Attribution labels and Custom Labels set, saving the check causes the Cost Attribution labels to be moved into the Custom Labels section on refresh. The labels that were configured as Cost Attribution labels (`namespace`, `check_type`) no longer render in the Cost Attribution area of the Labels tab.

**What you expected to happen**:

Cost Attribution labels should continue to render in the Cost Attribution section after saving, and Custom Labels should render only in the Custom Labels section. The two should remain visually and functionally distinct regardless of how many times the check is saved or reloaded.

**How to reproduce it (as minimally and precisely as possible)**:

1. Go to Cost Management and Billing
2. Set the Synthetic Monitoring Cost Attribution labels:
   1. `namespace`
   2. `check_type`
3. Go to a Synthetic Monitoring check > Labels section
4. Modify the `unattributed` value for the Cost Attribution labels
5. Add a Custom Label:
   1. `group`
6. Set a value for the Custom Label
7. Click `Save` and watch the refresh — the Cost Attribution labels get updated to sit under Custom Labels

**Anything else we need to know?**:

This appears to be related to how the frontend perceives the difference between Cost Attribution labels and Custom Labels. Both exist as a single line within JSON and within HCL for Terraform, so the issue is likely in the rendering/grouping of the labels rather than in what is persisted.

Note: This is more of a visual bug than anything, when going back to the check the labels return to normal.

Possibly related: https://github.com/grafana/synthetic-monitoring-api/pull/2396

https://github.com/user-attachments/assets/ab034ae5-362d-4d17-9bd5-677ff4b11eca

**Environment**: Production, Dev, UAT
- Grafana version (if running on prem): Cloud (Latest)
- SM plugin version: Cloud (Latest)
- OS Grafana is installed on: Cloud Hosted
- User OS & Browser: MacOS (Version 26.5.2 (25F84)) / Chrome (Version 151.0.7922.72 (Official Build) (arm64))
- Others: N/A

PR #1796, opened 2026-08-10
fix: keep cost attribution labels grouped after saving a check

g3john:
Addresses: https://github.com/grafana/synthetic-monitoring-app/issues/1791

## Problem

When a check has both cost attribution labels and custom labels, saving the check moves the cost attribution labels into the Custom labels section. The cost attribution rows are left blank, as though the values had been cleared.

Nothing is actually lost. The check is saved correctly and the labels return to their proper sections when you navigate away and come back, but it looks like the save has just discarded your cost attribution values.

## Solution

Sorting the labels into the two sections was a one-shot, only re-running when the tenant's cost attribution names changed. It now also runs whenever the cost attribution rows are out of sync with the names they should contain, which is the state a form reset leaves behind. It keys off the cost attribution rows rather than scanning the custom labels for cost-attribution-looking names

### Before

https://github.com/user-attachments/assets/adb055bf-5bad-4049-9cba-ef694d2f4eea

### After

https://github.com/user-attachments/assets/03eb6b77-61c6-413b-9af7-62a2d2e3bf75

### DIFF
--- a/src/components/Checkster/components/form/layouts/GenericLabelContent.test.tsx
+++ b/src/components/Checkster/components/form/layouts/GenericLabelContent.test.tsx
@@ -1,9 +1,14 @@
 import React from 'react';
-import { screen, waitFor } from '@testing-library/react';
+import { FormProvider, useForm } from 'react-hook-form';
+import { OpenFeatureTestProvider } from '@openfeature/react-sdk';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { SM_OPEN_FEATURE_DOMAIN } from 'services/featureFlags';
 import { TENANT_COST_ATTRIBUTION_LABELS } from 'test/fixtures/tenants';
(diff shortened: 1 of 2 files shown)


src/components/Checkster/components/form/layouts/GenericLabelContent.test.tsx
```diff
@@ -29,6 +34,35 @@ function renderGenericLabelContent(
...
+
+  render(
+    <OpenFeatureTestProvider domain={SM_OPEN_FEATURE_DOMAIN} flagValueMap={getTestFlagValues()}>
+      <Harness />
+    </OpenFeatureTestProvider>
+  );
+
+  return userEvent.setup();
+}
+
```
g3john, line 65: Not sure if its worth doing this just to have a test

g3john:
Can close this if/when https://github.com/grafana/synthetic-monitoring-app/pull/1778 goes in

PR #1800, opened 2026-08-11
fix: group cost attribution labels before the check form is populated

g3john:
Addresses the known limitation called out in #1796, which fixed https://github.com/grafana/synthetic-monitoring-app/issues/1791.

## Problem

After #1796 the labels end up in the right sections, but saving a check still produces a brief flash where the visible custom label rows show the cost attribution names and values before snapping back.

## Solution

Group the labels while the form values are derived, so the form is populated with them already separated and never holds a flat list.

Doing it here rather than in a component effect is what removes the flash: `reset()` writes into the label inputs before React re-renders, so by the time any effect could correct it the frame has already been painted.

### Before

https://github.com/user-attachments/assets/03eb6b77-61c6-413b-9af7-62a2d2e3bf75

### After

https://github.com/user-attachments/assets/ce34427f-f5d3-400c-9e37-79bc242dfb28

### DIFF
--- a/src/components/Checkster/components/form/generic/GenericNameValueField.test.tsx
+++ b/src/components/Checkster/components/form/generic/GenericNameValueField.test.tsx
@@ -37,6 +37,13 @@ jest.mock('data/useDefaultFolder', () => ({
   })),
 }));
 
+jest.mock('data/useTenantCostAttributionLabels', () => ({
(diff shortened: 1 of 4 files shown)


g3john:
Can close this if/when https://github.com/grafana/synthetic-monitoring-app/pull/1778 goes in

PR #1801, opened 2026-08-11
feat: attach global context props (page, check count) to all tracking events

g3john:
Addresses: https://github.com/grafana/synthetic-monitoring-app/issues/1598

## Problem

Our tracking events capture what a user did but not the situation they were in when they did it. When we analyse events (or read feature feedback), we can't segment by basic environmental facts — which page the user was on, or whether they run 2 checks or 20,000. We learned this the hard way with Time Point Explorer feedback, where we couldn't correlate reactions with the user's setup. Missing properties require a code change, a release, and weeks of data accumulation before we can answer a question.

## Solution

Every event created through `createSMEventFactory` now automatically carries global context properties, resolved lazily at the moment the event fires (so they are always current, with no navigation listeners):

- `page` — the current page as a low-cardinality route pattern (e.g. `checks/:id/edit`), resolved via ranked `matchRoutes` against a new `PAGE_ROUTE_PATTERNS` list; pages outside the plugin (e.g. external components) report their raw pathname
- `check_count` — the tenant's total check count, read passively from the react-query cache (never triggers a fetch; omitted until a page has loaded the check list, mirroring how `org_id`/`stack_id` behave before the tenant resolves)

Each resolver is individually guarded so a failure in one can never break the interaction being tracked nor take the other properties down with it. Explicit properties always win on key collision: event props > base props (`org_id`/`stack_id`) > global props. The Faro echo backend forwards all interaction properties, so the enrichment reaches Faro automatically.

## Rollout context

This is PR 1 of 3 for #1598. PR 2 adds mounted "scope" enrichment (check metadata + dashboard time range + check form state) and PR 3 adds Time Point Explorer-specific properties.

Made with [Cursor](https://cursor.com)

Issue #1598, opened 2026-02-20
feat: Enrich tracking events with contextual metadata


ckbedwell:
### Description

Our tracking events capture actions but lack the surrounding context needed to draw meaningful conclusions. We should attach additional metadata properties to events describing the user's state and environment.

### Use Case / Motivation

- Without contextual metadata, we can't segment or correlate user behaviour with their setup
- Learned this the hard way with Time Point Explorer feedback — we can't tell what time range, check frequency, or screen size users had when they gave feedback
- Properties that go unused can be ignored; missing properties require a code change, release, and weeks of data accumulation

Reference: [Session Replay PR review comment](https://github.com/grafana/grafana-sessionreplay-app/pull/67#pullrequestreview-3832327314)

### Proposed Solution

Audit existing tracking events and attach contextual metadata. Candidates:

**Global (most events):**
- Screen size / viewport
- Total check count
- Current time range
- Current page / section

**Check-specific:**
- Check type, frequency, probe count
- Whether alerting is enabled

**Feature-specific (e.g. Time Point Explorer, feedback):**
- Time range, check frequency, visible data points
- Section the user is interacting with

Favour over-logging — unused properties cost nothing, missing ones cost months.

PR #1802, opened 2026-08-11
feat: attach check, form and time-range scope props to tracking events

g3john:
Addresses: https://github.com/grafana/synthetic-monitoring-app/issues/1598

> Stacked on #1801 (PR 1 of 3) — review that first; this diff is only the scope layer.

## Problem

With global context in place (#1801), events still can't answer check-level questions. When a user gives Time Point Explorer feedback or clicks around a check dashboard, we don't know which kind of check they were looking at, how frequently it runs, how many probes it uses, whether alerting is set up, or what time range they had selected. The same applies in the check form: events don't say whether the user was creating, editing, or duplicating a check.

## Solution

A new `useTrackingScope(props)` hook attaches props to every tracking event fired while the calling component is mounted (kept current across re-renders, removed on unmount). Three scopes are mounted:

- **Check dashboard** (`DashboardContainer`): `check_id`, `check_type`, `check_frequency_seconds`, `probe_count`, `alerting_enabled`, `check_enabled`. The alerting logic is extracted from the check list filter into a shared `checkHasAlerting` util so both use the same definition.
- **Dashboard time range** (`TrackingTimeRangeScope`, mounted inside the `SceneContextProvider` of both the check dashboard and the homepage summary): `time_range_from` / `time_range_to` (relative expressions as-is, absolute ranges as ISO strings) + `time_range_seconds`.
- **Check form** (`ChecksterProvider`): `check_type`, `check_state` (`new`/`existing`), `check_id`, `check_is_duplicate`. Duplicates report as `new` because the duplicate flow strips the source check id.

This means every event fired on a check dashboard — including Time Point Explorer interactions and feature feedback — now carries the check's metadata and the selected time range, which is exactly the context we were missing when reading TPE feedback.

Scopes must use disjoint prop namespaces (`check_*`, `time_range_*`, ...): collision order between concurrently mounted scopes follows React effect timing and is deliberately not a supported semantic (pinned down in tests). Explicit event props always win over scope props.

## Rollout context

PR 2 of 3 for #1598. PR 3 adds Time Point Explorer-specific properties (`tpe_*`) and the analytics documentation update.


ckbedwell (dismissed):
LGTM!

My preference with stacked PRs is they go into main then the stacked PR redirects to `main` as it creates an easier audit trail but in this case we've ended up with a conflict 🤦 

So, get this fixed and we'll get the other two PRs in!

ckbedwell (approved):
LGTM

PR #1804, opened 2026-08-11
feat: attach Time Point Explorer display state to tracking events fired inside it

g3john:
Addresses: https://github.com/grafana/synthetic-monitoring-app/issues/1598

> Stacked on #1802 (PR 2 of 3), which is stacked on #1801 — this diff is only the Time Point Explorer layer and docs.

## Problem

Time Point Explorer feedback was the motivating example for #1598: users react with a thumbs up/down, but we can't correlate their reaction with what they were actually looking at — which view mode, how many data points were visible, or which minimap page/section they were exploring. The same context is missing from the explorer's own interaction events.

## Solution

The `TimepointExplorerProvider` now registers a tracking scope (using the `useTrackingScope` hook from #1802), so every event fired inside the explorer — all nine TPE interaction events, screenshot/trace events in the log viewer, and crucially the feature feedback events — automatically carries:

- `tpe_view_mode` (`uptime`/`reachability`)
- `tpe_visible_timepoints` (how many data points fit on screen)
- `tpe_total_timepoints` (how many exist in the selected time range)
- `tpe_page` / `tpe_section` (where in the minimap the user is)

Combined with the earlier layers, a TPE feedback event now reports the user's page, check count, the check's type/frequency/probes/alerting, the selected time range, and the explorer's display state. An integration test clicks the feedback button inside the rendered explorer and asserts the enriched payload end to end.

This PR also documents the full automatic-enrichment model (global props, base props, scopes) in `docs/analytics/analytics.md`, including how to add a scope for a new feature. `analytics-events.md` required no regeneration since no event definitions changed.

## Rollout context

PR 3 of 3 for #1598. With this, all three metadata tiers from the issue (global, check-specific, feature-specific) are implemented.


ckbedwell (dismissed):
LGTM

ckbedwell (approved):
LGTM

PR #1805, opened 2026-08-11
fix: replace expand/collapse all with a single toggle in folder view

vesnadean:
## Problem

In the check list folder view, Expand all and Collapse all were separate buttons. That took extra space and did not match how Grafana Alerting handles the same action on Notification policies, where a single control toggles between the two states.

## Solution

Folder view now uses one secondary button that switches between Expand all and Collapse all based on whether every folder is already expanded. The button label, icon, and aria-label update together so the control stays clear and accessible, aligned with the Alerting notification policies pattern.

https://github.com/user-attachments/assets/5433b084-4680-41e8-9b5c-9334c97ce77a

## Test plan

- [ ] Open Synthetic Monitoring → Checks and switch to folder view
- [ ] Confirm a single Expand all / Collapse all button appears next to the Folders heading
- [ ] With folders collapsed (or partially collapsed), click Expand all and verify all folders open
- [ ] With all folders expanded, confirm the button shows Collapse all and collapses every folder
- [ ] Confirm individual folder expand/collapse still works independently

### DIFF
--- a/src/page/CheckList/CheckList.folderView.test.tsx
+++ b/src/page/CheckList/CheckList.folderView.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { screen } from '@testing-library/react';
+import { screen, waitFor } from '@testing-library/react';
 import { CHECKS_TEST_ID } from 'test/dataTestIds';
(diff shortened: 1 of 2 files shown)


g3john (approved):
LGTM!

PR #1806, opened 2026-08-11
feat: introduce backend component to plugin

Pokom:
Part of a wider set of changes to provide a mechanism for external systems such as https://github.com/grafana/gcx ((private issue) https://github.com/grafana/synthetic-monitoring/issues/634) to query data the same way the plugin does. In order to accomplish that, I'm introducing a backend component that is initially just a stub. The long term goal is to migrate all of the datasource queries to the backend in the same way https://github.com/grafana/synthetic-monitoring-app/pull/1785 does. 

This PR introduces the first change to add a backend component to the SM
app. The included set of changes are:
- creates Magefile.go for building the backend
- sets pkg/main.go as the entry point
- updates plugin.json to specify that the sm app is a backend plugin
- defines a sane set of settings for golang-ci for linting
- creates an action to lint the codebase as part of CI

It does not migrate any of the existing plugin to use the backend
datasource for any calls. Plainly, this means that no functionality changes for the plugin. All of the existing datasource calls are still routed through the proxy. After this is released, we can stage the changes in the following ways:

1. (here) Add backend plugin component to the app  
2. Add in [`grafana.go`](http://grafana.go) [`authz.go`](http://authz.go) `` and `iam.go` ``  
   * First query: \`probe\_execution\_rate\` — tenant-wide, no parameters, so no string interpolation of user input, so \`promql.go\` and its escaping stay out. Reviewers judge the security model, not the escaping.  
   * Validate results with curl  
3. Copy app queries over to the backend component, but leave the app using the existing ones  
   1. Allows us to start wiring up gcx to call the queries  
4. Migrate parameterize queries to use backend component  
5. Migrate scene queries to use backend component  
6. Delete old queries from the app


ckbedwell (approved):
LGTM! Let's get in and see if we break dev 😄

scripts/validate-backend
```diff
@@ -0,0 +1,135 @@
...
+EOT
+
+cat > "$override" <<EOT
+services:
+  grafana:
+    container_name: $NAME
+    ports: !override
+      - $PORT:3000
+    volumes:
+      - $provisioning/sm-backend-validate.yaml:/etc/grafana/provisioning/datasources/sm-backend-validate.yaml
```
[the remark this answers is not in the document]
g3john: Ran this locally on macOS (Docker Desktop, virtiofs) and hit a harder version of this finding: the file-into-bound-dir mount prevents the container from starting at all. Docker has to create the mountpoint file *through* the outer `dev/provisioning` bind, and virtiofs refuses:

    OCI runtime create failed: ... create mountpoint for /etc/grafana/provisioning/datasources/sm-backend-validate.yaml mount: mountpoint "/run/host_virtiofs/.../dev/provisioning/datasources/sm-backend-validate.yaml" is outside of rootfs

So as written the script is Linux-only. (The stray host file is created even on the failed run — though it can't be committed: `dev/provisioning/datasources/.gitignore` ignores everything except `example.yaml`.)

Mounting the temp dir over the datasources dir fixes both problems: the mountpoint already exists so nothing gets created on the host, and the throwaway instance no longer sees local datasource provisioning:

```suggestion
      - $provisioning:/etc/grafana/provisioning/datasources
```

Verified on macOS (arm64): fails as-written; with this one-line change the full run passes (`PASS: backend builds, loads, registers and reports healthy`).

scripts/validate-backend
```diff
@@ -0,0 +1,135 @@
...
+PLUGIN_ID=synthetic-monitoring-datasource
+DS_UID=sm-backend-validate
+
+toplevel=$(git rev-parse --show-toplevel)
+cd "$toplevel"
+
+override=$(mktemp)
+provisioning=$(mktemp -d)
+cleanup() {
+	docker compose -f docker-compose.yaml -f "$override" -p "$PROJECT" down >/dev/null 2>&1 || true
```
g3john, line 29: Minor, found while running this locally: `down` without `--volumes` leaves the project's named volume behind after every run (`sm-backend-validate_grafana-storage`, from `grafana-storage` in the base compose file). Since the whole project is throwaway:

```suggestion
	docker compose -f docker-compose.yaml -f "$override" -p "$PROJECT" down --volumes >/dev/null 2>&1 || true
```

Pokom: Ah good find, sorry about that. Mind fixed that forward? I know Chris has some changes he wants to do as well

PR #1815, opened 2026-08-17
feat: add secret scanner to SM checks page

cristianoventura:

Adds hardcoded-secret detection to the check script editor, using `@grafana/plugin-ui/secret-scanner`. As users type a k6 script, the scanner flags likely secrets (API keys, tokens, passwords) inline and surfaces them in a panel above the editor.

Screenshot 2026-08-17 at 12 55 00 PM

Each finding can be moved into the Secrets Manager in one step - the same modal used to create a secret opens prefilled with a suggested name and the detected value, and on save the script is rewritten in place to reference the new secret (`await secrets.get('...')`) instead of the literal value. Findings can also be dismissed if they're false positives.

The Monaco editor also includes a "quick fix" option:

Screenshot 2026-08-17 at 12 55 28 PM

<br />
<br />

**How to test**:

- On the create/edit SM check page, add a few different hardcoded secrets (see examples below)
- Check that the panel above the editor shows the detected secret
- Click on "Move to secret" and check that the create-secret modal is displayed with pre-populated values
- Click "Ignore" and check that the secret no longer shows in the panel above the editor

Examples:

```js
# jwt
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NT.KMUFsIDTnFmyG3nMiGM6H9FNFUROf3wh7SmqJp-QV30

# named secret
const password = 'he110-w0rlD'

# gh token
const t = 'ghp_NOT_A_VALID_GITHUB_TOKEN'

# google api key
const k = `AIzaSyDaGmWKa4JsXZ-HjGw7ISLn_3namBGewQe_NOTVALID`

# generic api key
const host = 'http://quickpizza.grafana.com'
const url = `${host}?token=INcIbu6jVhjaZN+pFFK4MWAkPfGUBc/2gLz+oXRj`
```

Fixes: N/A

- New runtime dependency: `@grafana/plugin-ui@0.17.4` (detection, Monaco markers/quick-fix, and panel components live upstream; this PR only wires them up and owns the create-secret flow via our own `SecretEditModal`).
- `SecretEditModal` gained optional `initialValues`/`onCreated` props so it can be prefilled and driven from the scanner's migration flow, in addition to its existing standalone use on the config page.

### DIFF
--- a/src/components/Checkster/components/form/generic/GenericScriptField.test.tsx
+++ b/src/components/Checkster/components/form/generic/GenericScriptField.test.tsx
@@ -38,6 +38,22 @@ jest.mock('components/CodeEditor', () => ({
   )),
 }));
 
+// Secret scanner integration: the feature is off by default here, and its data
+// hooks require app-level providers this lightweight renderer doesn't supply.
+jest.mock('contexts/FeatureFlagContext', () => ({
+  ...jest.requireActual('contexts/FeatureFlagContext'),
+  isFeatureEnabled: jest.fn(() => false),
+}));
+
+jest.mock('data/permissions', () => ({
+  ...jest.requireActual('data/permissions'),
+  getUserPermissions: jest.fn(() => ({ canCreateSecrets: false, canReadSecrets: false })),
+}));
+
+jest.mock('data/useSecrets', () => ({
+  useSecrets: jest.fn(() => ({ data: [] })),
+}));
+
 const defaultProps = {
   field: 'value',
 } as any;
--- a/src/components/Checkster/components/form/generic/GenericScriptField.tsx
+++ b/src/components/Checkster/components/form/generic/GenericScriptField.tsx
@@ -1,14 +1,18 @@
 import React from 'react';
 import { useController, useFormContext } from 'react-hook-form';
-import { FieldValidationMessage, useTheme2 } from '@grafana/ui';
-import { css, cx } from '@emotion/css';
+import { SecretReferenceModal, SecretScannerPanel } from '@grafana/plugin-ui/secret-scanner';
+import { Box, FieldValidationMessage, useTheme2 } from '@grafana/ui';
+import { css } from '@emotion/css';
 
 import { CheckFormFieldPath } from '../../../types';
 import { CheckFormValues, K6Channel } from 'types';
 import { CodeEditor } from 'components/CodeEditor';
+import { SECRETS_EDIT_MODE_ADD } from 'page/ConfigPageLayout/tabs/SecretsManagementTab/constants';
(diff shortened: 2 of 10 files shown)


src/components/Checkster/components/form/generic/GenericScriptField.tsx
```diff
@@ -1,14 +1,18 @@
 import React from 'react';
 import { useController, useFormContext } from 'react-hook-form';
-import { FieldValidationMessage, useTheme2 } from '@grafana/ui';
-import { css, cx } from '@emotion/css';
+import { SecretReferenceModal, SecretScannerPanel } from '@grafana/plugin-ui/secret-scanner';
```
g3john, line 3: Not blocking, but something we can consider is lazy loading this import, which is increasing the load by 550KB (4.88%)

src/components/Checkster/components/form/generic/GenericScriptField.test.tsx
g3john, line 1: Let's add some unit tests!

src/components/Checkster/components/form/generic/GenericScriptField.tsx
```diff
@@ -23,35 +27,59 @@ export function GenericScriptField({ field }: GenericScriptFieldProps) {
...
-        css`
-          & > div:first-child {
-            flex: 1 1 0;
-            overflow: visible;
-          }
-          & > div > div {
+    <Column grow>
+      <Box padding={2} paddingBottom={0}>
+        <SecretScannerPanel scanner={scanner} readOnly={disabled} hidden={!secretsEnabled} />
+      </Box>
```
g3john, line 50: The Box is taking up space whether or not secrets is Enabled, and whether or not the Panel is visible. Let's only render the Box if we're showing the Panel.

g3john (commented):
Looking great! Something we might want to think about is having the banner visible when they scroll far down in the script. They might miss it in this case

src/components/Checkster/components/form/generic/GenericScriptField.tsx
ckbedwell: Going to leave this for this PR but it is worth us doing at some point and identifying logical places to do chunk splitting and lazy loading. The thing is we lazy-load the `module.js` and then the whole Synthetic Monitoring app when needed so despite the large bundle size it doesn't concern me too much.

ckbedwell:
> Looking great! Something we might want to think about is having the banner visible when they scroll far down in the script. They might miss it in this case

Going to leave this one for now just in the interest of getting the PR in and can look at when we do some dogfooding UI pickups.

g3john (approved):
Looks good!

PR #1817, opened 2026-08-18
fix: dedupe probe_success join in response latency panel

The-9880:
Relates to https://github.com/grafana/synthetic-monitoring/issues/549

Closes https://github.com/grafana/synthetic-monitoring/issues/653

When a tenant triggers the label migration, `probe_success`'s label set changes (`region` → `sm_region`, plus dual-written user labels). For the ~5-minute staleness window both series generations are live and share the same `(instance, job, probe, config_version)` match group, so the panel's `group_left` join fails with *found duplicate series on the right hand-side* and the panel errors for the whole window.

Aggregating the right side down to exactly the match-group labels collapses both generations into one series per group. `max` is semantically neutral here (same underlying 0/1 value), duplicate left-hand series are legal under many-to-one matching, and the outer `avg by (probe)` absorbs the doubled duration series without changing the result.

This conforms the one outlier to the pattern the other joining panels already use — `SummaryTableViz`, `ErrorRateMapViz`, and `SummaryErrorPctgViz` all aggregate their right-hand side with `max by (instance, job, probe, config_version, …)`.

----

With fix:
image

Without fix:
image

### DIFF
--- a/src/scenes/Common/ResponseLatencyByProbe.tsx
+++ b/src/scenes/Common/ResponseLatencyByProbe.tsx
@@ -27,7 +27,7 @@ export const ResponseLatencyByProbe = () => {
     maxDataPoints: 100,
     queries: [
       {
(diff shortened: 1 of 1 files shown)


g3john (approved):
LGTM!

PR #1818, opened 2026-08-19
Reliability Inbox

oleiade:

Introduces an experimental, feature-gated (pending) Reliability Inbox in Synthetic Monitoring. 

It identifies likely gaps in synthetic coverage, presents prioritized recommendations with supporting evidence, and lets users review and customize a proposed check through Grafana Assistant before explicitly choosing whether to create it.

This is the first narrow experiment of the broader Reliability Inbox concept: turning observed reliability opportunities into safe, product-native actions while keeping users in control.

CleanShot 2026-08-19 at 08 56 05

CleanShot 2026-08-19 at 08 56 23

Note that the suggestions are generated by our [k6-experiments](https://github.com/grafana/k6-experiments) service and will eventually be cached in the user's  browser session.


src/datasource/DataSource.ts
```diff
@@ -270,6 +271,51 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
...
+  //--------------------------------------------------------------------------------
+
+  /**
+   * Whether a reliability-inbox instance serves this stack's region. Callers
+   * should hide the feature when this is false, rather than calling and handling
+   * the error.
+   */
+  supportsReliabilityInbox(): boolean {
+    return reliabilityInboxURL(this.instanceSettings.jsonData.apiHost) !== undefined;
+  }
```
ckbedwell: Can we move this out of the datasource? This is just a utility function so shouldn't be here.

src/datasource/DataSource.ts
```diff
@@ -270,6 +271,51 @@ export class SMDataSource extends DataSourceApi<SMQuery, SMOptions> {
...
+  async getReliabilityInboxSuggestions(): Promise<unknown> {
+    const url = reliabilityInboxURL(this.instanceSettings.jsonData.apiHost);
+    if (!url) {
+      throw new Error('The reliability inbox is not available in this region');
+    }
+
+    // Called directly instead of through the datasource proxy. The proxy would
+    // inject the token from secureJsonData, but its route cannot exist until
+    // these plugin.json changes ship in a released plugin version — Grafana
+    // registers proxy routes server-side from the installed plugin. So we mint a
```
ckbedwell: Totally get this in the context of doing development but we need to change this getting ready to deploy to production.

I'm enjoying thinking about the graveyard of long-lived tokens this call has been creating in dev (they don't expire and are non manageable right now...) 😛

src/features/reliabilityInbox/assistantActionStyles.test.ts
ckbedwell, line 1: Get rid of this test, it adds little value. (Any tests asserting specific styles has limited to no value)

src/features/reliabilityInbox/assistantActionStyles.ts
ckbedwell, line 1: Util files which are dedicated just to styles is a code smell. What you want to do here instead is create a generic AssistantButton component which has these styles and caters for the two places they are used.

It will have to be relatively generic and work out whether to render under the hood as a Link or Button but it is cleaner and more consistent than trying to maintain these styles which might clash with the props of the consumer.

src/features/reliabilityInbox/data.test.tsx
ckbedwell, line 1: I'd just kill this test. It does some horrible mocking with type assertions, it couples with the implementation too closely and it provides little value over the UI tests you have in other places like `ReliabilityInboxPage.test.tsx` and `ReliabilityInboxBanner.test.tsx`.

src/features/reliabilityInbox/data.ts
```diff
@@ -0,0 +1,127 @@
...
+    smDS.instanceSettings.jsonData.metrics.hostedId
+  );
+}
+
+function useReliabilityInboxQuery(generateSuggestions: boolean) {
+  const smDS = useSMDS();
+  const apiHost = smDS.instanceSettings.jsonData.apiHost;
+  const stackId = smDS.instanceSettings.jsonData.metrics.hostedId;
+  const storageKey = reliabilityInboxStorageKey(apiHost, stackId);
+  const snapshot = readSnapshot(storageKey);
```
ckbedwell: As you've probably discovered at this point -- optimistic data hydration is just a huge can of worms.

I'd just drop this feature. Keep the dismissed storage IDs (really, I'd rather that was stored in an api somewhere but I'm fine with this as a temporary workaround) and the versioned storage key but putting large payloads of API data in localStorage becomes problematic quickly.

src/features/reliabilityInbox/data.ts
```diff
@@ -0,0 +1,127 @@
...
+      // The experiment reports degraded failures as HTTP 200 with warnings.
+      // Keep the last useful inbox instead of replacing it with that empty response.
+      if (result.suggestions.length === 0 && result.warnings.length > 0) {
+        throw new Error(result.warnings.join('; '));
+      }
+
+      writeSnapshot(storageKey, result.suggestions);
+
+      return toOpportunities(result.suggestions);
+    },
```
ckbedwell, line 145: Keep this `queryFn` pure with no side effects or data transformation. Make a call whether you want to lift the transformation to the datasource or lower it to either the select function or render time.

src/features/reliabilityInbox/model.ts
```diff
@@ -0,0 +1,169 @@
...
+  const timeout = prompt.match(/timeout ([^,\s]+)/)?.[1];
+  const statusCodes = prompt.match(/expect HTTP status \[([^\]]*)\]/)?.[1];
+  const probeIds = prompt.match(/probe IDs \[([^\]]*)\]/)?.[1];
+
+  return {
+    job,
+    frequencyMs: frequency ? parsePromptDuration(frequency) : undefined,
+    timeoutMs: timeout ? parsePromptDuration(timeout) : undefined,
+    validStatusCodes: parseNumberList(statusCodes),
+    failIfNotSSL: /fail if not SSL/i.test(prompt),
```
ckbedwell: I'd pull these out so this function stays a generic any-check-type parameters function and then have a map of dedicated check-type functions to handle different check types. The only caller of this function right now is a HTTP one so it could even just move to that function for now so as not to predict the future?

Just want to preempt the type union nightmare this function would currently create as it grew.

src/features/reliabilityInbox/model.ts
ckbedwell, line 1: This file should be split. Reading through it it's not entirely what I'd expect when I see a `model.ts`.

Keep the pure model logic in this file and for the parts which are doing the assistant-shaped proposed check make a new file called `proposedCheck.ts`.

src/features/reliabilityInbox/ReliabilityInboxBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+    trackInboxExposure({
+      opportunityCount: opportunities.length,
+      topOpportunityId: topOpportunity.id,
+    });
+  }, [opportunities.length, topOpportunity]);
+
+  return (
+    <Box
+      element="section"
+      aria-label="Reliability Inbox"
```
ckbedwell: This doesn't need an `aria-label`.

src/features/reliabilityInbox/ReliabilityInboxPage.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+import { generateRoutePath, getRoute } from 'routing/utils';
+
+import { useReliabilityInboxDismissals, useReliabilityInboxSuggestions } from './data';
+import { toReliabilityOpportunity } from './model';
+import { ReliabilityInboxPage } from './ReliabilityInboxPage';
+
+jest.mock('./data', () => ({
+  useReliabilityInboxDismissals: jest.fn(),
+  useReliabilityInboxSuggestions: jest.fn(),
+}));
```
ckbedwell, line 25: The much more durable and easier to reason with way of writing these tests is using MSW and mocking the server rather than these implementation-specific functions. 

If you point an agent at `src/test/handlers/index.ts` and ask it to do some exploration there, it should come to the conclusion of what factories to write powered by Fishery, what handlers to add and how to rewrite these tests so it uses the mock server rather than taking this approach.

When you go this route we can much more easily start testing the unhappy paths as well.

src/features/reliabilityInbox/ReliabilityInboxPage.tsx
```diff
@@ -0,0 +1,645 @@
...
+      </Stack>
+    </PluginPage>
+  );
+}
+
+function ReliabilityInboxReview() {
+  const styles = useStyles2(getStyles);
+  const { canWriteChecks } = getUserPermissions();
+  const { isAvailable: isAssistantAvailable, isLoading: isAssistantLoading, openAssistant } = useAssistant();
+  const { data, isLoading, isFetching, isError, refetch } = useReliabilityInboxSuggestions();
```
ckbedwell: It's worth getting the error rather than `isError` as from that you can work out what UI to display. Right now we don't check if it is a permissions issue or a 400 / 500 or what, we just show a UI that says it is a permissions issue.

src/features/reliabilityInbox/ReliabilityInboxPage.tsx
ckbedwell, line 1: This file is pretty big, can we split this out by the components to make it more manageable 😃

src/features/reliabilityInbox/ReliabilityInboxPage.tsx
```diff
@@ -0,0 +1,645 @@
...
+  },
+};
+
+export function ReliabilityInboxPage() {
+  return (
+    <PluginPage
+      pageNav={RELIABILITY_INBOX_PAGE_NAV}
+      renderTitle={() => (
+        <Stack alignItems="center" gap={1.5}>
+          <h1>Reliability Inbox</h1>
```
ckbedwell: They have a `<Text>` component you can use here instead of using the native html element.

src/features/reliabilityInbox/ReliabilityInboxPage.tsx
```diff
@@ -0,0 +1,645 @@
...
+  checkSummary: css({
+    alignItems: 'center',
+    background: theme.colors.background.secondary,
+    borderRadius: theme.shape.radius.default,
+    display: 'flex',
+    gap: theme.spacing(2),
+    justifyContent: 'flex-start',
+    marginTop: theme.spacing(2),
+    padding: theme.spacing(1.5),
+    [`@media (max-width: ${theme.breakpoints.values.md}px)`]: {
```
ckbedwell: Can we convert this to a [container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) rather than a media query please. We've stopped using media queries since container queries are better supported now.

src/features/reliabilityInbox/types.ts
```diff
@@ -0,0 +1,48 @@
...
+        families: z.array(z.string()).default([]),
+        provenance: z
+          .object({
+            datasource: z.string(),
+            queries: z.array(z.object({ expr: z.string(), instant: z.boolean().optional() }).loose()),
+            range: z.object({ from: z.string(), to: z.string() }),
+          })
+          .optional(),
+      })
+      .loose(),
```
ckbedwell: Do we need `.loose()` on these types, I can't see any benefit in having them?

ckbedwell (changes requested):
Alright, gone through it in some detail! I'm going to put up a suggestion PR in a moment just with some UI/UX improvement suggestions -- that kind of feedback makes more sense in a visual medium (and is quicker to iterate on and get in...)

ckbedwell:
Suggestion is https://github.com/grafana/synthetic-monitoring-app/pull/1827. I'd take the page split, `Text` for the title, dropping the banner `aria-label`, dismissed-as-a-tab, and the NewCheck `prefilledCheck` path.

I'm not doing the rest of this review there. Tokens, moving `supportsReliabilityInbox`, AssistantButton, dropping the localStorage snapshot, keeping `queryFn` pure, splitting `model.ts`, killing `data.test` / the style tests, error vs `isError`, `.loose()`, and MSW for the page tests still belong here.

If you take the suggestion, don't take the client-side PromQL rewrite, the `animationend`-gated dismiss, or Create manually without `canWriteChecks`. Those came from iterating on the UI and I'd rather they not land on this branch as-is.

oleiade:
@ckbedwell As mentioned on Slack, I believe I have addressed most of your comments in a commit in #1827

Let me know if I forgot anything, or if you have any further/follow-up comments 🙇🏻

pkg/plugin/datasource.go
```diff
@@ -32,18 +38,41 @@ var PluginVersion = "development"
...
+	return &Datasource{
+		accessToken: settings.DecryptedSecureJSONData["accessToken"],
+		httpClient: &http.Client{
+			Timeout: 90 * time.Second,
+			CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
+				return http.ErrUseLastResponse
+			},
+		},
+		suggestionsURL: reliabilityInboxURL(jsonData.APIHost),
+	}, nil
```
codebien, line 87: Is this expected to land in the final version?

AGENTS.md
```diff
@@ -25,8 +25,8 @@ Standard commands are in `package.json`:
 The plugin has a Go backend under `pkg/`. It backs the **nested
 `synthetic-monitoring-datasource` plugin** (`src/datasource/plugin.json`), not the
 app itself — Grafana routes panel and scene queries through datasources, so an app
-backend could not serve them. It currently only starts up and reports its health —
-it serves no queries or resources and no frontend code calls it.
+backend could not serve them. It reports its health and proxies Reliability Inbox
+suggestion requests so the browser never receives the stored access token.
```
codebien: This change sounds a bit out of context.

pkg/plugin/datasource.go
```diff
@@ -58,3 +87,135 @@ func (d *Datasource) CheckHealth(_ context.Context, _ *backend.CheckHealthReques
...
+	if len(hostParts) < 2 {
+		return ""
+	}
+
+	label := hostParts[0]
+	domain := strings.Join(hostParts[1:], ".")
+
+	environment := ""
+
+	switch {
```
vortegatorres, line 237: We should include `prod` here to have the implementation ready for prod and avoid requiring another PR.

oleiade: Can you make a suggestion please? I have no idea what the case should look like for prod 🙇🏻

pkg/plugin/datasource.go
oleiade: What's the concern, and what would be an alternative to address it?

Bear in mind I have zero experience with deploying to gck plugins/apps, so I relied on AI for this one.

AGENTS.md
oleiade: This should be fixed.

pkg/plugin/datasource.go
oleiade: This should be addressed, but can you take another look to be sure @vortegatorres ?

pkg/plugin/datasource.go
oleiade: From what I understand, yes, this is intended to be part of the final implementation. The backend proxy keeps the stored token server-side and derives the correct regional endpoint, which a static plugin route cannot do. 

I’m definitely not an expert on Grafana plugin backends though, so please let me know if I’m missing a simpler or more idiomatic approach 🙇🏻

src/datasource/reliabilityInboxRegion.ts
```diff
@@ -0,0 +1,37 @@
+const DEPLOYED_DOMAINS = [
+  ['.grafana-dev.net', 'dev'],
+  ['.grafana-ops.net', 'ops'],
+] as const;
+
+const SM_API_PREFIX = 'synthetic-monitoring-api-';
+const SUGGESTIONS_PATH = '/api/v1alpha1/reliability-inbox/suggestions';
+
+/** Returns the co-located experimental service URL, failing closed elsewhere. */
+export function reliabilityInboxURL(apiHost: string): string | undefined {
```
ckbedwell: This builds a URL that nothing actually uses. `data.ts` only compares it to `undefined`, and the real request goes through the datasource resource route, where `reliabilityInboxURL` in `datasource.go` derives the same thing again. So the region rules live twice, in two languages — the production-regions commit (since reverted) had to edit both copies and both test files to add a region.

I'd rather the backend owned this outright. It already returns 404 "not available in this region" before it makes any outbound call, so the frontend gate isn't saving a paid request.

It's not a straight delete though, and this is the bit worth deciding on: with the gate the query stays idle and the page falls through to the "No reviewable opportunities" empty state. Without it, unsupported regions hit the 404 and get "The Reliability Inbox service is unavailable. Try again later." — which is misleading when the truth is "not deployed here". So if we drop the local copy, we'd want the backend to answer availability (a GET on the resource, or something in `CheckHealth`) and gate on that.

No strong view on the mechanism. I just don't want two copies of the region maths.

ckbedwell: My preference is we drop this now that we have the backend datasource handling constructing the URL.

However, it is still being used as an identifier and a gate of whether the region has the check suggestion API so should display it or not.

Can we add another route to your API (and then to the datasource) which is a simple GET -- if we get a 200 it is available, otherwise it isn't and we don't show any check suggestion UI?

ckbedwell (dismissed):
Let's goooooooooo!

I'm approving despite the comment I left. I would like to see that prioritised as right now it is confusing it is defined in two places but I'm favouring velocity and pragmatism to get this in.

There's some big UI components that could be better split up (which is 100% my agents' doing from the suggestion branch as I didn't review their work in detail...) that could be organised better but that doesn't need to be addressed in this PR and a questionable prop on the toggletip but I am comfortable with taking that tech debt on and we address it in a follow-up PR.

In a follow-up PR we should explore registering the route in the Grafana navigation too. It needs to be gated but I can't remember the implementation so needs investigation.

Great job all! Excited to see how this performs!

src/datasource/reliabilityInboxRegion.ts
vortegatorres: PRs for this:
- https://github.com/grafana/synthetic-monitoring-app/pull/1839

g3john (approved):
Approving to move it forward, bugbot findings aren't blockers

PR #1821, opened 2026-08-19
feat(faro): instrument user actions across check and probe flows

digitalcrab:
Addresses: https://github.com/grafana/synthetic-monitoring/issues/650

## Problem

We have no visibility into what people actually click through the check and probe
flows - check creation, the check wizard, check list actions, probe setup/reset/delete.
Without that, we cannot build user journeys later on.

## Solution

Instruments those flows with Faro user actions via `trackFaroUserAction(FaroUserAction, attributes)`,
which wraps `faro.api.startUserAction()`. See: https://grafana.com/docs/grafana-cloud/observe-and-act/monitor-applications/frontend-observability/instrument/user-actions/

Went with the imperative API instead of the declarative `data-faro-user-action-name`
attribute - that path is broken upstream (I opened https://github.com/grafana/faro-web-sdk/pull/2239
to fix it), so imperative was the only way not to block this work.

Tracks clicks through:
- check creation entry point + type selection
- check wizard navigation (next/prev/tab) and create vs. update submit
- check list actions: edit, duplicate, enable/disable, delete + confirm/cancel
- probe list/edit/delete/reset + confirm/cancel, ad hoc check test, probe setup modal copy

Also removes the localhost opt-out in `App.tsx` - Faro doesn't spam the console with
errors anymore, so it's fine to leave enabled everywhere, and we need it locally to
actually verify this instrumentation.

### DIFF
--- a/src/components/AddNewCheckButton/AddNewCheckButton.tsx
+++ b/src/components/AddNewCheckButton/AddNewCheckButton.tsx
@@ -4,9 +4,12 @@ import { LinkButton } from '@grafana/ui';
 import { trackAddNewCheckButtonClicked } from 'features/tracking/checkCreationEvents';
 import { ACTIONS_TEST_ID } from 'test/dataTestIds';
 
+import { FaroUserAction } from 'faro';
 import { AppRoutes } from 'routing/types';
 import { generateRoutePath } from 'routing/utils';
 import { getUserPermissions } from 'data/permissions';
+
+import { trackFaroUserAction } from '../../features/tracking/userAction';
 
 interface AddNewCheckButtonProps {
   source: 'check-list-empty-state' | 'check-list' | 'homepage';
@@ -17,6 +20,7 @@ export function AddNewCheckButton({ source }: AddNewCheckButtonProps) {
 
   const handleClick = useCallback(() => {
     trackAddNewCheckButtonClicked({ source });
+    trackFaroUserAction(FaroUserAction.CreateNewCheckClicked, { source: source });
   }, [source]);
 
   return (
@@ -28,9 +32,7 @@ export function AddNewCheckButton({ source }: AddNewCheckButtonProps) {
       onClick={handleClick}
       variant="primary"
     >
-      <Trans i18nKey="addNewCheckButton.createNewCheck">
-        Create new check
-      </Trans>
+      <Trans i18nKey="addNewCheckButton.createNewCheck">Create new check</Trans>
     </LinkButton>
   );
 }
--- a/src/components/AddNewCheckButton/AddNewCheckButton.tsx
+++ b/src/components/AddNewCheckButton/AddNewCheckButton.tsx
@@ -20,7 +20,7 @@ export function AddNewCheckButton({ source }: AddNewCheckButtonProps) {
 
   const handleClick = useCallback(() => {
     trackAddNewCheckButtonClicked({ source });
-    trackFaroUserAction(FaroUserAction.CreateNewCheckClicked, { source: source });
+    trackFaroUserAction(FaroUserAction.CreateNewCheckClicked, { source });
   }, [source]);
 
   return (
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -24,37 +24,38 @@ import { SMOpenFeatureProvider } from './SMOpenFeatureProvider';
 
 const { env, url, name } = getFaroConfig();
 
-// Used to be disabled on localhost because it was filling up the console with error logs.
-// That's no longer the case, so it's safe to leave enabled everywhere - which we now need
-// to actually test user action tracking during local development.
-getAppPluginVersion('grafana-synthetic-monitoring-app').then((version) => {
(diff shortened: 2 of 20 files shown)


src/components/App.tsx
```diff
@@ -24,37 +24,37 @@ import { SMOpenFeatureProvider } from './SMOpenFeatureProvider';
 
 const { env, url, name } = getFaroConfig();
 
-// faro was filling up the console with error logs, and it annoyed me, so I disabled it for localhost
-if (window.location.hostname !== 'localhost') {
```
ckbedwell, line 28: I'd rather default to not sending Faro events for localhost, it doesn't really provide us any value sending them outside testing it works.

Let's just document that if you need to test Faro events whilst developing you comment this line out or just use Graft on a cloud stack of your choosing.

src/components/AddNewCheckButton/AddNewCheckButton.tsx
```diff
@@ -17,6 +20,7 @@ export function AddNewCheckButton({ source }: AddNewCheckButtonProps) {
 
   const handleClick = useCallback(() => {
     trackAddNewCheckButtonClicked({ source });
+    trackFaroUserAction(FaroUserAction.CreateNewCheckClicked, { source: source });
```
ckbedwell: The nittest of nits: this is more of an inform than a nit as I know Typescript isn't your favoured language but there's a beautiful shorthand in Typescript where when the variable's assignment matches the key you want to assign to an object you can just do { key }, so this line can become:

```
trackFaroUserAction(FaroUserAction.CreateNewCheckClicked, { source });
```

src/components/Checkster/components/form/FormFooter.tsx
```diff
@@ -66,6 +79,9 @@ export function FormFooter() {
           data-testid={CHECKSTER_TEST_ID.form.submitButton}
           variant={isStepsComplete || !next ? 'primary' : 'secondary'}
           disabled={disableSubmit}
+          // Do not add data-faro-user-action-name="save" here: this shared submit button
+          // is used for both create and edit flows, so the action must be named where
+          // the submit handler can distinguish create vs. update. (@see useHandleSubmitCheckster)
```
ckbedwell: ✂️ 

I do think it is worth documenting somewhere more general that we are favouring the manual API rather than data- attributes to bypass this problem, though.

src/faro.ts
```diff
@@ -24,6 +24,36 @@ export enum FaroEvent {
...
+  NewProbeCreateModalDismissClicked = 'new-probe-create-probe-modal-dismiss-clicked',
+  EditPrivateProbeClicked = 'edit-private-probe-clicked',
+  AddPrivateProbeClicked = 'add-private-probe-clicked',
+  DeletePrivateProbeClicked = 'delete-private-probe-clicked',
+  DeletePrivateProbeConfirmationClicked = 'delete-private-probe-confirmation-clicked',
+  DeletePrivateProbeCancellationClicked = 'delete-private-probe-cancelation-clicked',
+  ResetAccessTokenClicked = 'reset-access-token-clicked',
+  ResetAccessTokenConfirmationClicked = 'reset-access-token-confirmation-clicked',
+  ResetAccessTokenCancellationClicked = 'reset-access-token-cancelation-clicked',
+}
```
ckbedwell, line 54: Because we have a lot of refactoring coming up in our future I was thinking we should add a test to make sure each of these is referenced once in the application. Quick agent-driven test looks like:

```
import fs from 'fs';
import path from 'path';

...

function collectAppSource(dir: string): string {
  return fs
    .readdirSync(dir, { withFileTypes: true })
    .flatMap((entry) => {
      const fullPath = path.join(dir, entry.name);

      if (entry.isDirectory()) {
        if (entry.name === '__tests__' || entry.name === 'node_modules') {
          return [];
        }

        return collectAppSource(fullPath);
      }

      if (entry.name === 'faro.ts' || /\.test\.(ts|tsx)$/.test(entry.name) || !/\.(ts|tsx)$/.test(entry.name)) {
        return [];
      }

      return [fs.readFileSync(fullPath, 'utf8')];
    })
    .join('\n');
}

function unusedMembers(enumName: string, members: string[], source: string) {
  return members.filter((member) => !source.includes(`${enumName}.${member}`));
}

describe(`Faro enums are referenced in the app`, () => {
  const source = collectAppSource(path.join(__dirname));

  it(`uses every FaroEvent at least once`, () => {
    expect(unusedMembers('FaroEvent', Object.keys(FaroEvent), source)).toEqual([]);
  });

  it(`uses every FaroUserAction at least once`, () => {
    expect(unusedMembers('FaroUserAction', Object.keys(FaroUserAction), source)).toEqual([]);
  });
});
```

That actually caught that we no longer use `DisablePlugin` in`FaroEvent`s so point made 😂

ckbedwell (changes requested):
Few changes but otherwise this is good to go! Great work 👏 ⭐

ckbedwell (approved):
LGTM! Great job

PR #1824, opened 2026-08-19
feat: rename colliding labels from the label migration tab

The-9880:
Closes https://github.com/grafana/synthetic-monitoring/issues/657

The Label Migration UI allows users to attempt to change the state of labelling in the synthetic-monitoring-agent and eventually move off of the `label_` prefix.

The API call to do so returns a `409` if any of the user's custom labels clash with our system-reserved labels.

This enhances the UI by providing an interface for users to rename custom labels across their checks to assist them in resolving the clash so they can proceed with the label migration.

Backed by the bulk rename endpoint (`POST /check/labels/{key}`, [OpenAPI spec](https://synthetic-monitoring-api.grafana.net/api/v1/swagger#/operations/POST_/api/v1/check/labels/:key)), and a **Retry enabling dual-write** button gated until every label has been renamed.

## Behavior

- Rename targets are validated client-side (label syntax, not a reserved name, not the target of another row) before any request; the API's both-keys 409 renders inline on the row.
- A successful rename locks its row and reports how many checks changed.
- **Scope: checks only.** The collision pre-check also scans probe labels, but the rename endpoint covers checks; a rename that matches no checks shows a hint that the label may live on a probe, which must be edited directly (accepted v1 limitation — the 409 response doesn't distinguish sources). The retry gate still opens in that case: the retry simply 409s again with the remaining labels and the flow remounts fresh against the new list.
- Each new 409 remounts the rename flow, so state never goes stale across attempts.

## Testing

Seven new tests: the full collide → rename (path + body asserted) → retry → dual-write flow; client-side rejection of reserved/duplicate/invalid targets with zero API calls; inline both-keys conflict; probe hint with the gate deliberately open; row lock after success; fallback error text; and the second-409 remount asserting no stale state survives.

### DIFF
--- a/src/data/useRenameCheckLabels.ts
+++ b/src/data/useRenameCheckLabels.ts
@@ -1,0 +1,16 @@
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+
+import { QUERY_KEYS as CHECK_QUERY_KEYS } from 'data/useChecks';
+import { useSMDS } from 'hooks/useSMDS';
+
+export function useRenameCheckLabels() {
+  const smDS = useSMDS();
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: ({ from, to }: { from: string; to: string }) => smDS.renameCheckLabels(from, to),
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: CHECK_QUERY_KEYS.list });
+    },
+  });
+}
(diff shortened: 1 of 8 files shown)


g3john (changes requested):
After clicking "Enable dual-write" the first time and it returns 409, it'ss still there and remains clickable. Let's hide or at least disable the button while the conflict alert is showing, and bring it back if they dismiss the alert

https://github.com/user-attachments/assets/a92d7a15-cd91-4eb1-9aed-1784dd38bd51

The-9880:
@g3john Updated - the original `Enable dual-write` button disappears while the panel is open and only the `Retry` one becomes enabled after you rename all the offending labels:

image

g3john (approved):
LGTM!

PR #1825, opened 2026-08-19
feat(reliability-inbox): Gate the feature on Assistant availability

codebien:
## What
Hides the Reliability Inbox banner whenever the Grafana Assistant is unavailable or its availability is still resolving.

## Why
We need Assistant to use the feature in a functional way. This method also guarantees that the user accepted Assistant T&C.


ckbedwell:
Closing as this work was included in the main PR.

PR #1826, opened 2026-08-20
fix: prevent creating checks in folders the user cannot edit

VikaCep:
Addresses: https://github.com/grafana/support-escalations/issues/23773

## Problem

A user with Synthetic Monitoring checks writer role but only **View** permission on the default `Grafana Synthetic Monitoring` folder (valid cases are their folder Edit grants live on subfolders) gets a misleading check-creation experience:

- The form opens with the read-only default folder preselected (remember the folder exists and is readable, but not editable).
- Saving succeeds, but the check is then governed by the default folder's permissions — so its own creator cannot edit, pause or delete it. Only Editors can modify it.

The folder dropdown already offered only editable folders, but the the default folder was a special case that was pre-selected regardless of its permissions.

## Solution

Preselection is now permission-aware, and a folder is required whenever folder data is available:

- **Edit on the default folder (the vast majority)** 
Preselected as before — no change 
image

- **Edit on exactly one folder** 
That folder is preselected
image

- **Edit on several folders, not the default** 
No preselection, lists all the editable folders.
image

- **Edit on no folder**
 The picker is replaced by an explanation asking an administrator for Edit access to the default folder or one of its subfolders 
image

**Note:**
When the feature flag is off, the folder isn't provisioned, or the user lacks `folders:read`, the rule is inert and checks save without a folder exactly as before.

## Notes for reviewers
- Disabled Edit/Duplicate actions in the check list now get the same dimmed styling as the neighboring icon buttons, so read-only checks read as read-only.
- This is deliberately scoped to preselection, validation and labeling. The picker still lists only the default folder's subtree; broader folder support (any editable folder, server-side permission filtering via the standard `FolderPicker`) is actively worked on in [#1790](https://github.com/grafana/synthetic-monitoring-app/pull/1790), which replaces the `FolderSelector` internals. 
- This is UI-level guidance, not enforcement: the SM API and Terraform accept any `folderUid` regardless of folder permissions. True enforcement would be an API-side decision.

## Testing

- New test helpers simulate the three permission setups (read-only default with several/single/no editable folders).
- Coverage for every row of the behavior matrix: `FolderSelector` unit tests, form UI tests, and two end-to-end journeys.

### DIFF
--- a/src/components/FolderSelector/FolderSelector.tsx
+++ b/src/components/FolderSelector/FolderSelector.tsx
@@ -3,6 +3,7 @@ import { Alert, Button, Combobox, ComboboxOption, Field, Input, LoadingPlacehold
 import { trackFolderCreated, trackFolderSelected } from 'features/tracking/folderEvents';
 
 import { GrafanaFolder } from 'types';
+import { FolderAccessState } from 'data/folderPermissions';
 import { useDefaultFolder } from 'data/useDefaultFolder';
 import { useFolderPermissions } from 'data/useFolderPermissions';
 import { getFolderPathParts, useCreateFolder, useFolderChildren } from 'data/useFolders';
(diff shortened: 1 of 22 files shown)


src/components/FolderSelector/FolderSelector.tsx
```diff
@@ -60,17 +83,29 @@ export function FolderSelector({ value, onChange, disabled, autoSelectDefault =
...
+  }, [defaultFolder, permissionsSettled, editableFolders]);
 
   useEffect(() => {
-    if (autoSelectDefault && value === undefined && defaultFolderUid) {
-      onChange(defaultFolderUid);
+    if (autoSelectDefault && value === undefined && preselectUid) {
+      onChange(preselectUid);
     }
-  }, [autoSelectDefault, value, defaultFolderUid, onChange]);
+  }, [autoSelectDefault, value, preselectUid, onChange]);
```
g3john: With just one editable folder, opening the new check form and leaving straight away pops the 'You have unsaved changes' modal without touching anything. The auto-select here goes through `field.onChange` (`FormFolderField.tsx:29`) which marks the form dirty on mount, and `Checkster.tsx:50` feeds `isDirty` into `ConfirmLeavingPage`, so a clean form prompts. The default folder case doesn't hit it because that one is seeded through form defaults.

Let's do the auto-pick without dirtying the form, e.g. `setValue('folderUid', uid, { shouldDirty: false })` from `FormFolderField`. One thing to watch: the reset in `ChecksterContext.tsx:200` only keeps dirty values, so a non-dirty pick can get cleared and re-picked when the defaults recompute. Might be worth a test for this!

src/schemas/forms/BaseCheckSchema.ts
```diff
@@ -28,8 +28,29 @@ export const baseCheckSchema = z.object({
...
+  schema: ZodType<T, any, any>,
+  { requiresFolder = false }: CheckSchemaRefinementOptions = {}
+) {
   return schema
+    .superRefine((data, ctx) => {
+      if (requiresFolder && !data.folderUid) {
+        ctx.addIssue({
+          path: ['folderUid'],
+          code: 'custom',
+          message: 'Select a folder to store this check in.',
```
g3john, line 50: In the no-editable-folders state the picker is replaced by the permission alert, but submitting still shows this under it, and there's nothing to select. We shoulr probably skip the field error when the alert is up?

Image

g3john (dismissed):
Awesome!

g3john (approved):
LGTM!

PR #1827, opened 2026-08-21
Suggestion/reliability inbox

ckbedwell:


oleiade:
Love the updates you've made to the design @ckbedwell 🫶 They're spot on.

I only have small nitpicks on sizings mostly, bear with me, and feel free to ignore them too, it's just me sharing my thoughts; I'm probably not right.

On the review suggestions banner on the homepage, I actually love that you made it so big! It's bold and clearly attracts my eyes. But in terms of proportions I find the vertical negative space to be a bit too much. Codex says it's 33px currently. I wonder if 24px would work? I don't know what sizing scale Grafana uses generally. An alternative could be to balance that with larger text, but I think we might already be at the limit on that front.
CleanShot 2026-08-21 at 11 22 49

Regarding the box in which we display the suggestions, I noticed that it seems we don't apply the same spacing on the left, and on the right. Once I noticed it, I couldn't unsee it. If I had to chose, I'd probably go with the smaller spacing (left), and apply the same on the right.
CleanShot 2026-08-21 at 11 26 15

src/features/reliabilityInbox/ReliabilityInboxPage.test.tsx
```diff
@@ -74,53 +86,235 @@ describe('ReliabilityInboxPage', () => {
...
-    expect(within(evidence).getByRole('link', { name: 'Explore telemetry' })).toHaveAttribute(
+    expect(screen.queryByRole('region', { name: 'Why this recommendation' })).not.toBeInTheDocument();
+
+    const evidenceDisclosure = within(suggestedCheckHeader!).getByRole('button', { name: 'Why this check?' });
+    expect(evidenceDisclosure).toHaveAttribute('aria-expanded', 'false');
+    expect(within(suggestedCheck).queryByRole('region', { name: 'Recommendation evidence' })).not.toBeInTheDocument();
+
+    await user.click(evidenceDisclosure);
+
+    const evidence = within(suggestedCheck).getByRole('region', { name: 'Recommendation evidence' });
```
ckbedwell: This looks up `region` named `Recommendation evidence`, heading `Traffic signals`, and `Last 60 minutes`. The component renders an unnamed `role="region"` whose heading is `No matching uptime check was found`.

Those assertions fail as written. Either restore the labelled region and copy, or retarget the tests. `role="region"` also needs an accessible name.

src/features/reliabilityInbox/components/SuggestedCheckPanel.tsx
```diff
@@ -0,0 +1,459 @@
...
+        </Button>
+      </footer>
+    );
+  }
+
+  return (
+    <footer className={styles.footer}>
+      <div className={styles.leftContent}>
+        <div className={styles.createActions}>
+          <Button variant="secondary" onClick={onCreateManually}>
```
ckbedwell: `canWriteChecks` only feeds the Assistant button. This always `locationService.push`es.

Viewers already get a disabled NewCheck form, so it isn't a privilege bug, but it is a dead-end `AddNewCheckButton` doesn't have. Disable/tooltip both create actions the same way, and add a viewer case.

src/features/reliabilityInbox/evidence.ts
```diff
@@ -17,5 +41,46 @@ export function getEvidenceExploreUrl(evidence: ReliabilityEvidence): string | u
...
   }
 
-  return getExploreUrl(provenance.datasource, provenance.queries, { from, to });
+  return {
+    datasource: provenance.datasource,
+    from,
+    to,
+    queries: provenance.queries.map((query, index) => {
+      const refId = query.refId?.trim() || String.fromCharCode(65 + index);
+      const trendExpression = query.expr.replaceAll('[1h]', `[${INLINE_TREND_RATE_WINDOW}]`);
```
ckbedwell: `replaceAll('[1h]', '[5m]')` isn't PromQL-aware, `refId === 'B'` couples the 5xx rewrite to query order, and the result is both posted to `/api/ds/query` on expand and stuffed into Explore.

#1818 opened Explore with the backend's queries, on click. Pinning the SM metrics DS is the right mitigation; spreading `...query` from a `.loose()` schema is not. Keep Explore on the original `expr`s. Allowlist `{ refId, expr }`.

src/features/reliabilityInbox/components/ReliabilityInboxReview.tsx
```diff
@@ -0,0 +1,249 @@
...
+        buttonText="Retry"
+        content="Check your permissions and the live Reliability Inbox service, then try again."
+        onClick={() => refetch()}
+        title="Unable to load Reliability Inbox"
+      />
+    );
+  }
+
+  const refreshStatus = <RefreshStatus isFetching={isFetching} isError={isError} onRetry={() => refetch()} />;
+  const runAfterExit = (transition: 'fade' | 'dismiss', action: () => void) => {
```
ckbedwell: This sets `pointer-events: none` and no-ops further actions until `animationend`. Tests paper over it with `fireEvent.animationEnd`.

If `matchMedia('(prefers-reduced-motion: no-preference)')` and `theme.transitions.handleMotion('no-preference')` disagree, or the event never fires, dismiss/select/restore stick. Timeout fallback, or apply the state change immediately and keep the animation visual-only.

src/features/reliabilityInbox/components/SuggestionsRefreshControl.tsx
```diff
@@ -0,0 +1,64 @@
...
+              <div>
+                <Button onClick={refresh}>Refresh suggestions</Button>
+              </div>
+            </Stack>
+          }
+          onClose={() => setShowConfirmation(false)}
+          show={showConfirmation}
+          strategy="fixed"
+        >
+          <Button
```
ckbedwell: Controlled `show` plus this `onClick={() => setShowConfirmation(true)}` fights `useClick` on Toggletip. Second click on an open trigger: close, then the button sets it true again, so you can't dismiss via the trigger.

`Feedback` already passed `show`/`onClose`; this PR makes that live, and the primitive has no tests of its own. In controlled mode, don't let `useClick` open.

src/features/reliabilityInbox/components/InboxDisclosure.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useStyles2 } from '@grafana/ui';
+import { css, cx } from '@emotion/css';
+
+interface InboxDisclosureProps {
+  summary: string;
+  children: React.ReactNode;
+  inline?: boolean;
+}
+
+export function InboxDisclosure({ summary, children, inline }: InboxDisclosureProps) {
```
ckbedwell: Dead after the Button disclosure replaced `<details>`. Delete it.

src/features/reliabilityInbox/components/RecommendationEvidence.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const { user } = render(
+      <RecommendationEvidence
+        opportunity={toReliabilityOpportunity(HTTP_RELIABILITY_SUGGESTION)}
+        headerContent={<span>Suggested check</span>}
+      />
+    );
+
+    await user.click(await screen.findByRole('button', { name: 'Why this check?' }));
+
+    expect(screen.getAllByText('10:00')).toHaveLength(3);
```
ckbedwell: Fixture range is UTC `10:00`–`11:00`, asserted via `dateTimeFormat(..., { format: 'HH:mm' })` with no `timeZone`. In BST that's `11:00`/`12:00`. Pass `{ timeZone: 'utc' }` (or assert via a formatted expected value).

ckbedwell (commented):
Editing this comment as I can't delete it -- agent overstepped and decided posting its self-review on the PR was the correct thing to do...

PR #1828, opened 2026-08-21
fix: support publishing releases from vX.Y.Z tags

VikaCep:
## Problem

The manual deployment workflow deploys whatever `main`'s HEAD is at the time it runs. Because ops and prod are deployed at different times, merges landing in between mean the two environments can ship **different builds under the same version number**. This is exactly what broke [ops](https://raintank-corp.slack.com/archives/C0162C1K9E1/p1787323607369809) on 1.59.0: the ops catalog ended up expecting the prod build's `moduleHash` while the ops CDN served the earlier ops build, so SRI checks blocked `module.js` and users saw "App not found".

Separately, deploying a PR branch to dev has been silently broken since `trigger-argo` was narrowed to `main` only (#1689): the deploy publishes to the dev catalog but the Argo rollout never runs, so the preview never reaches instances. The same gap made tag deploys (e.g. `v1.45.0`) appear to succeed while never rolling out.

## Solution

Release tags (`vX.Y.Z`) are now treated as release references alongside `main`:

- **Ops and prod releases should be deployed from the release tag**, guaranteeing both environments build from the exact same commit and produce identical artifacts for a given version.
- Tag deploys get the clean version from `plugin.json` (no `+<sha>` suffix) and trigger the Argo rollout, so they actually reach instances.
- Deploying a PR branch to dev triggers the Argo rollout again, restoring dev previews. Branch deploys keep the `+<commit SHA>` suffix so preview versions stay unique.

Dev auto-deploys from `main` are unaffected: they run via `validate_main.yml`, which sets `plugin-version-suffix` explicitly.

## Behavior matrix

| Ref | Environment | Version | Argo rollout |
|---|---|---|---|
| `main` | any | clean | yes |
| `vX.Y.Z` tag | any | clean | yes |
| PR branch | dev | `+<sha>` suffix | yes (restored) |
| PR branch | ops/prod | blocked by `cd.yml` (non-release ref) | — |

## Follow-ups

- Update `docs/synthetic-monitoring/release-process.md` in `deployment_tools` to document deploying from tags (separate PR).
- After the next release, dry-run by deploying its `vX.Y.Z` tag to dev and confirming the clean version publishes and Argo triggers.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
     inputs:
       branch:
-        description: Branch to publish from (can be used to deploy PRs to dev)
+        description: Branch or tag to publish from.
         default: main
       environment:
         description: Environment to publish to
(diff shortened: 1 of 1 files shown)


ckbedwell (approved):
Great find. I wondered about this when I was doing the deployment actually but figured it would be fine 😅

PR #1830, opened 2026-08-24
feat: add feature feedback widget to the Connected services section

VikaCep:
## Problem

The Connected services section (the Knowledge Graph service neighbourhood shown on a check's dashboard) recently shipped, but we have no way of knowing how users feel about it. Other new features like the Timepoint Explorer and per-check alerts surface our feature feedback widget so users can react and leave comments, and this section should follow the same pattern while it's still new.

Additionally, the "Open in Knowledge Graph" action in the section header was rendered as a plain text link, making it easy to miss as the primary way to jump into the Knowledge Graph app.

## Solution

The section header now includes the shared feedback widget — a "New feature!" badge with thumbs up/down buttons and an optional comment form — tracked under the `knowledge-graph` feature name via the existing `feature_feedback` analytics events. It renders in all section states (graph, zero state, error), so users can leave feedback even before linking a check to a service.

image

The "Open in Knowledge Graph" link has also been promoted to a secondary button (keeping anchor semantics via `LinkButton`), matching how prominent navigation is presented elsewhere in the app.

An integration test covers the widget rendering in the section header; the feedback form and tracking behaviour are already covered by the `Feedback` component's own tests.

### DIFF
--- a/src/features/knowledgeGraph/ConnectedServices.tsx
+++ b/src/features/knowledgeGraph/ConnectedServices.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { GrafanaTheme2 } from '@grafana/data';
-import { Icon, IconButton, Stack, Text, TextLink, useStyles2 } from '@grafana/ui';
+import { Icon, IconButton, LinkButton, Stack, Text, TextLink, useStyles2 } from '@grafana/ui';
 import { css } from '@emotion/css';
 
 import { Check } from 'types';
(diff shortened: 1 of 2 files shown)


src/features/knowledgeGraph/ConnectedServices.tsx
```diff
@@ -76,10 +77,17 @@ function ConnectedServicesSection({ check }: ConnectedServicesProps) {
             {CONNECTED_SERVICES_SUBTITLE}
           </Text>
         </div>
+        <Feedback feature="knowledge-graph" about={{ text: `New feature!` }} />
```
g3john: Should we be more specific? Something like `knowledge-graph-connected-services` so we know exactly which feature we're collecting feedback on, or in case we add more knowledge graph related features that we want to get feedback from?

g3john (dismissed):
LGTM!

src/features/knowledgeGraph/ConnectedServices.tsx
VikaCep: Good catch!

g3john (approved):

PR #1831, opened 2026-08-25
fix: populate the KG graph search from the connected-services drawer links

rdubrock:
## What

The **Open in Knowledge Graph** links in the connected-services graph popover only carried the `ed[...]` drawer params:

```
/a/grafana-asserts-app/entities?ed[type]=Service&ed[name]=checkoutservice&ed[scope][env]=...
```

In the KG app, `ed` maps only to the entity-details drawer state. The entities page derives its underlying graph search from `filterCriteria` — nothing derives a search from `ed`. So the link opened the drawer floating over an empty page, and the incomplete URL also broke the KG app's back-button handling (its URL-sync rewrites the URL with the full serialized state, stacking a history entry the back button can never escape).

## Fix

`getEntityDrawerUrl` now also appends the `filterCriteria` params that the KG's own **Explore connected entities → View Graph** action dispatches: an `EQUALS` search on `name` (plus `env`/`namespace` when scoped), connected to Services. The KG page hydrates that into its search state, so the graph loads underneath the drawer — same result as navigating from inside the KG app.

## Testing

- `ConnectedServices.utils.test.ts` updated: asserts drawer params, search matchers, and empty-scope handling via parsed `URLSearchParams`.
- Verified the param shape round-trips through the KG app's `qs.parse` hydration (`fillSliceWithQueryParams` → `searchObject.filterCriteria`) against the asserts-app-plugin source.

### DIFF
--- a/src/features/knowledgeGraph/ConnectedServices.utils.test.ts
+++ b/src/features/knowledgeGraph/ConnectedServices.utils.test.ts
@@ -253,17 +253,50 @@ describe('getEntityDrawerUrl', () => {
     };
   }
 
+  function paramsOf(url: string): URLSearchParams {
+    return new URLSearchParams(url.split('?')[1]);
+  }
+
   it('builds the KG entity-drawer link with type, name and the non-empty scope values', () => {
     const url = getEntityDrawerUrl(buildNode());
+    const params = paramsOf(url);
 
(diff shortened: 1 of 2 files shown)


VikaCep (approved):
LGTM!

PR #1834, opened 2026-08-26
fix(suggestions): shrink banner and sharpen CTA visibility

codebien:
## What

Make the reliability inbox suggestion banner on the Home page more compact, and make its call-to-action button stand out more clearly.

## Why

The min height sounds oversized.
The CTA also needed to stay visible after the shrink, since this banner exists to drive adoption of the new suggestions feature. The idea is that a solid primary outline fill keeps the border/text as the primary for contrast without a competing with other components on the page.


ckbedwell (approved):
LGTM!

PR #1836, opened 2026-08-26
Hide label_ prefix hint in custom labels once tenant leaves PREFIXED mode

The-9880:
## Summary
- The Custom Labels tab in the check editor always showed a `label_` prefix hint (decorator + tooltip) on every label-name input, regardless of the tenant's label migration state.
- `GenericLabelContent` now reads `useLabelMode()` and only renders the prefix hint while the tenant is in `PREFIXED` mode; it's omitted once the tenant has moved to `DUAL_WRITE` or `UNPREFIXED`.
- While the label mode is still loading, the hint defaults to shown (matches the pre-existing always-on behavior, fails open rather than flickering).

Fixes grafana/synthetic-monitoring#663

## Test plan
- [x] `GenericLabelContent.test.tsx` — new `label_ prefix hint` cases: shown in PREFIXED, hidden in DUAL_WRITE, hidden in UNPREFIXED, shown while loading
- [x] `yarn typecheck`
- [x] `yarn test:ci` (full suite, no regressions)

### DIFF
--- a/src/components/Checkster/components/form/layouts/GenericLabelContent.test.tsx
+++ b/src/components/Checkster/components/form/layouts/GenericLabelContent.test.tsx
@@ -7,6 +7,8 @@ import { server } from 'test/server';
 import { mockFeatureToggles, testUsesCombobox } from 'test/utils';
 
 import { FeatureName } from 'types';
+import { LabelMode } from 'datasource/responses.types';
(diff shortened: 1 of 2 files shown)


VikaCep (approved):
LGTM!

PR #1837, opened 2026-08-27
feat: render invalidLabels from 409 collision response

The-9880:
## What this PR does

We return a new `invalidLabels` field in the `409` error response for changing the tenant's labelling mode.

That field captures which labels are blocking due to being invalid for Prometheus compatibility.

There was already an existing `collidingLabels` field capturing labels that clash with our system-reserved names.

Changes:
- `CollisionError` widened: both label array fields are optional.
- The 409 detection previously keyed on `collidingLabels` presence alone; it now fires when **either** array is non-empty, so an invalid-only conflict renders the guidance alert instead of the generic error. A 409 carrying only empty arrays deliberately routes to the generic path (shows the API `msg`) rather than an empty alert.
- Invalid names render as a second section in the existing conflict alert with their own explanation; lists addressed via new `CONFIG_TEST_ID.labelMigration` registry entries.

Without this change the shipped app still detects the new 409 (empty `collidingLabels` is truthy), just with degraded display - merge order with the API PR is not critical.

## Testing

Invalid-only 409 renders the invalid names in the conflict alert; colliding-only regression (invalid section absent); mixed 409 with `within()`-scoped proof each name sits only in its own list; empty-arrays 409 routes to the generic path with the API `msg`. TDD verified failing pre-implementation; two adversarial review rounds; 26/26 tests, typecheck clean, lint 0 errors.

### DIFF
--- a/src/page/ConfigPageLayout/tabs/LabelMigrationTab/LabelMigrationTab.test.tsx
+++ b/src/page/ConfigPageLayout/tabs/LabelMigrationTab/LabelMigrationTab.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import { CONFIG_TEST_ID } from 'test/dataTestIds';
 import { TENANT, TENANT_LABEL_MODE } from 'test/fixtures/tenants';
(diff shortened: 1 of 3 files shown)


VikaCep (approved):
LGTM

PR #1838, opened 2026-08-27
fix: sync grouped probe checkbox state

d0ugal:
## Summary
- synchronize grouped probe header checkbox state after the final probe is selected
- add regression coverage for the partial-to-complete selection transition

## Bug

If you check the last box in a list..
Screenshot 2026-08-27 at 10 37 18
The region checkbox is then unchecked
Screenshot 2026-08-27 at 10 37 26

## Test plan
- [x] Targeted ProbesList Jest test
- [x] TypeScript typecheck
- [x] ESLint


VikaCep (approved):
LGTM

PR #1843, opened 2026-09-03
feat: implement grafana datasource querying for backend component

Pokom:
Adds QueryData support to the backend datasource: named queries (currently just `probe_execution_rate`) resolve to an expression against the tenant's linked Prometheus/Loki datasource and are proxied through Grafana's own /api/ds/query as the calling user, not as the plugin. 

Authorization is delegated to Grafana's RBAC rather than reimplemented: before running anything, the backend checks the caller's datasources:query permission on the target datasource via `/api/access-control/users/permissions/search`. This fails future requests for the following conditions
- no user on the request
- no plugin service account credential
- failed lookup for the user 

`src/datasource/schema/v0alpha1/query.types.json` is an automatically generated file. It's the query-type schema Grafana serves to describe each named query's shape.  `pkg/plugin/schema_test.go`'s `TestUpdateSchema` regenerates the json file from the registry via `grafana-plugin-sdk-go's` schemabuilder, and fails if the checked-in copy is stale.  Run it directly after adding or changing a named query, then commit the result. 
`TestSchemaCoversRegistry` is the separate guard that keeps the schema list and the real query registry in namedqueries.go from silently drifting apart as queries are added.

The plugin needs its own service-account credential to talk to Grafana (the iam block in plugin.json), which requires externalServiceAccounts and managed service accounts enabled - both already wired into the gitignored `docker-compose.override.yaml` for local dev.

- Relates https://github.com/grafana/synthetic-monitoring/issues/634


docker-compose.override.yaml
```diff
@@ -0,0 +1,23 @@
...
+    extra_hosts:
+      - 'host.docker.internal:host-gateway'
+
+    environment:
+      # The datasource backend gets its service account from the `iam` block in
+      # src/datasource/plugin.json. Without both of these Grafana issues no
+      # credential and every named query is refused -- it fails closed rather
+      # than querying as itself. See pkg/plugin/authz.go.
+      GF_FEATURE_TOGGLES_ENABLE: localizationForPlugins externalServiceAccounts
+      GF_AUTH_MANAGED_SERVICE_ACCOUNTS_ENABLED: 'true'
```
VikaCep: Should this configuration be moved to `docker-compose.yaml` and gitignore this file?

pkg/plugin/datasource.go
```diff
@@ -6,14 +6,19 @@
...
-// It currently does nothing beyond starting up and reporting its health. It
-// exists to verify that the plugin binary is built, packaged, signed, shipped and
-// started correctly. It serves no queries and no resources, and no frontend code
-// calls it, so enabling the backend does not change plugin behaviour.
+// The app's metric and log queries are normally assembled in the browser and
+// sent straight to the Prometheus or Loki datasource. This backend lets the app
+// ask for a query *by name* instead -- `probe_execution_rate`, say -- and
+// resolves that name to an expression against whichever backing datasource is
+// appropriate, querying it as the calling user rather than as itself. See
+// authz.go for why that distinction matters.
```
VikaCep: Worth noting that the app can't actually reach this yet. `SMDataSource` extends `DataSourceApi` and its `query()` only recognises the `Probes`, `Checks` and `Traceroute` query types, so anything else falls through the if/else chain and returns `{ data: [] }`. 

I understand this is not te be addressed in this PR, but curious what would be the way for connecting the two: switching to `DataSourceWithBackend` and deferring unknown query types to `super.query()`, or having callers POST directly like `queryLogsV2` already does? Flagging mainly because the current failure mode is a silently empty panel rather than an error.

CONTRIBUTING.md
```diff
@@ -116,6 +116,64 @@ During the initialization process, the Synthetic Monitoring backend will:
 
 Users can then create checks to monitor their remote targets. Metrics and logs will flow into the selected Cloud stack.
 
+### Query the datasource backend directly
```
VikaCep, line 119: This section is hard-wrapped at ~80 columns while the rest of the file is one line per paragraph

VikaCep (dismissed):
This is great! 👏👏 👏 Just left a few questions.

My main question is what's the process for moving this forward? Do we port queries one at a time and swap each frontend caller as we go, or build the registry up first and migrate the app in one go? I ask because nothing in the app consumes this path yet, so my worry is the registry growing with queries we haven't compared against what the UI actually renders today.

Whichever way we go, could we write the porting workflow down (fine as a follow-up). The `CONTRIBUTING` section you added is a great start for calling a named query by hand and regenerating the schema. The parts I'd find most useful on top of that:

- The steps to add a query end to end: registry entry in `namedqueries.go`, parameter shape in `queryshapes.go`, description in `schema_test.go`, then regenerate and commit `query.types.json`. The schema half is covered already, the rest isn't.
- What tests a new entry is expected to come with, and which cases they should cover.
- What happens to the frontend tests when a caller is swapped over. `src/test/handlers/metrics.ts` mocks Prometheus' `/api/v1/query` and matches on the exact expression from `getProbeExecutionRateQuery()`, so a ported caller needs a different handler shape — and presumably the old `src/queries/` function and its handler branch get deleted in the same PR.

docker-compose.override.yaml
Pokom: That's a good point. Let me do that

pkg/plugin/datasource.go
Pokom: That's a good question, and I don't know. We discussed over DM's putting together a design doc. I think now is the right time to do that so we can come up some solution. I imagine the right thing to do is pick a few core parts of the app to port over that exercises each path properly.

VikaCep (approved):
LGTM! 👏 👏 👏

PR #1846, opened 2026-09-11
feat: add affordance for checks impacted by label migration

Pokom:
Previously, it was near impossible to understand the number of checks impacted before starting the label migration or during it. This adds a check _prior_ to starting the migration that lists out the total number of checks impacted along with a collapsable nav section with a list of checks and their associated labels. This allows a user to understand before and during the label migraiton process what checks need to udpating. I've provided a demo of it against a personal stack and showed using gcx to fix the checks prior to starting the migration.

Fixes #

https://github.com/user-attachments/assets/531e234f-1519-4baa-89f2-ece6a065ad0a

### DIFF
--- a/src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
+++ b/src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import { Stack, Text, TextLink } from '@grafana/ui';
+
+import { type Check } from 'types';
+import { AppRoutes } from 'routing/types';
+import { generateRoutePath, getRoute } from 'routing/utils';
+
+function getChecksListHref(label: string) {
+  const params = new URLSearchParams({ search: label });
+  return `${getRoute(AppRoutes.Checks)}?${params.toString()}`;
+}
+
+// BlockingChecksList shows which checks currently carry a given label, so an
+// admin can jump straight to them instead of guessing from the label name
+// alone. A label with zero matches most likely lives on a probe, which this
+// list can't see (probe labels aren't tied to any Check object).
+export function BlockingChecksList({ label, checks }: { label: string; checks: Check[] }) {
+  const blockingChecks = checks.filter((check) => check.labels.some((l) => l.name === label));
+
+  if (blockingChecks.length === 0) {
+    return (
+      <Text color="secondary" variant="bodySmall" data-testid={`blocking-checks-empty-${label}`}>
+        No checks currently carry this label — it may be set on a probe, which must be edited directly.
+      </Text>
+    );
+  }
+
+  return (
+    <Stack direction="row" gap={1} wrap="wrap" alignItems="center" data-testid={`blocking-checks-${label}`}>
+      <Text color="secondary" variant="bodySmall">
+        Blocking checks:
+      </Text>
+      {blockingChecks.map((check, i) => (
+        <React.Fragment key={check.id}>
+          <TextLink variant="bodySmall" href={generateRoutePath(AppRoutes.EditCheck, { id: check.id! })}>
+            {check.job}
+          </TextLink>
+          {i < blockingChecks.length - 1 && <Text color="secondary">,</Text>}
+        </React.Fragment>
+      ))}
+      <TextLink variant="bodySmall" href={getChecksListHref(label)} data-testid={`blocking-checks-view-all-${label}`}>
+        View all {blockingChecks.length} in Checks list
+      </TextLink>
+    </Stack>
+  );
+}
--- a/src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
+++ b/src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
@@ -10,11 +10,28 @@ function getChecksListHref(label: string) {
   return `${getRoute(AppRoutes.Checks)}?${params.toString()}`;
 }
 
+interface BlockingChecksListProps {
+  label: string;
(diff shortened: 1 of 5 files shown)


src/page/ConfigPageLayout/tabs/LabelMigrationTab/LabelMigrationTab.tsx
```diff
@@ -45,6 +47,7 @@ export function LabelMigrationTab() {
 
   const { data: state, isLoading, error: loadError, refetch, isRefetching } = useLabelMode();
   const setLabelModeMutation = useSetLabelMode();
+  const { data: checks } = useChecks();
```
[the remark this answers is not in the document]
Pokom: Fixed: added an explicit `checksError` flag threaded from `useChecks()` (in `LabelMigrationTab.tsx`) down through `ImpactedChecksWarning` and `CollidingLabelRename`/`BlockingChecksList`. On a failed fetch, both now render a distinct "couldn't verify" state instead of silently falling through to the confident zero-collisions/probe-only-label copy. Added regression tests covering both surfaces (`surfaces a failed check fetch instead of silently reporting no impacted checks`, `surfaces a failed check fetch in the reactive collision alert instead of a false "no checks" hint`). Not pushed yet — will follow shortly, please take another pass once it lands.

src/page/ConfigPageLayout/tabs/LabelMigrationTab/ImpactedChecksWarning.tsx
```diff
@@ -0,0 +1,79 @@
...
+      >
+        <Text>
+          There are {impactedCheckCount} check{impactedCheckCount === 1 ? '' : 's'} that currently use label names
+          reserved by the migration and will block this transition until renamed.
+        </Text>
+      </Alert>
+      <Collapse
+        label={`${isOpen ? 'Hide' : 'Show'} ${impactedLabels.length} impacted label${impactedLabels.length === 1 ? '' : 's'}`}
+        isOpen={isOpen}
+        onToggle={() => setIsOpen((v) => !v)}
```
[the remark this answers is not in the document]
Pokom: Fixed: the Collapse label is now static — `Impacted label(s) (N)` regardless of open state — instead of toggling between "Show"/"Hide". Grafana's `Collapse` already manages `aria-expanded" on its own toggle button, so the constant label plus that built-in state is sufficient. Not pushed yet — will follow shortly, please take another pass once it lands.

Pokom:
Pushed a fix for both Bugbot findings (checks fetch failure handling, static Collapse label). Requesting another review pass.

src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </Text>
+    );
+  }
+
+  return (
+    <Stack direction="row" gap={1} wrap="wrap" alignItems="center" data-testid={`blocking-checks-${label}`}>
+      <Text color="secondary" variant="bodySmall">
+        Blocking checks:
+      </Text>
+      {blockingChecks.map((check, i) => (
```
VikaCep: There's no cap here, so a tenant with 400 checks carrying instance renders 400 links inline. Can we slice to the first five and let the "View all" link below handle the rest?

g3john: To add to that, can we move the "View all" link away from the check links? Maybe right align it so that it's clearly a "View all" vs one of the checks
Screenshot 2026-09-11 at 2 14 53 PM

src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <React.Fragment key={check.id}>
+          <TextLink variant="bodySmall" href={generateRoutePath(AppRoutes.EditCheck, { id: check.id! })}>
+            {check.job}
+          </TextLink>
+          {i < blockingChecks.length - 1 && <Text color="secondary">,</Text>}
+        </React.Fragment>
+      ))}
+      <TextLink variant="bodySmall" href={getChecksListHref(label)} data-testid={`blocking-checks-view-all-${label}`}>
+        View all {blockingChecks.length} in Checks list
+      </TextLink>
```
g3john, line 66: The count here is an exact match on label name, but the link uses `?search=<label>` and `search` substring matches job, target, label names and label values (`CheckList.utils.ts:18-34`), so clicking "View all 3" can land you on a much bigger list. Pretty likely for the reserved names, `?search=job` will pull in anything with "job" in a job name or label value too.

There's no name-only label filter today (`labels` is exact `name: value`), so should we just drop the count from the link text for the short term? Long term would be adding name-only matching to the checks filter but may not be worth the time

src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
Pokom: Both sounds reasonable, thanks!

Pokom: Here's a screenshot from the latest changes:

image

src/page/ConfigPageLayout/tabs/LabelMigrationTab/BlockingChecksList.tsx
Pokom: Here's a screenshot from the latest changes:
image

VikaCep (approved):

PR #1851, opened 2026-09-15
fix: guard bulk check deletion behind a typed confirmation

VikaCep:
## Problem

The check list's select-all checkbox selected every check matching your current filters, across every page. The card view shows 15 at a time, so on a list of 2,000 checks a single click selected all 2,000. 
Deleting from there took two clicks. The confirmation asked "Are you sure you want to delete these checks?", kept the count tucked away in the modal title, and had its confirm button live from the moment the modal opened. Select all, delete, confirm, and the list is gone. This has already happened to someone internally; on a customer tenant it would be an unrecoverable and expensive mistake.

## Solution

Bulk deletion now requires typing "Delete" before the confirm button becomes active, using the same `ConfirmModal` affordance we already use for deleting a secret. The modal states the number of checks in the body instead of saying "these checks", and says plainly that the action cannot be undone.

In the card and list views, select-all now selects only the checks on the page you are looking at.
Folder view selects all checks within the folders of the current page.

image

**Folders view:**
https://github.com/user-attachments/assets/f506bb64-db45-4a33-aeed-c102b78a1413

**List view:**
https://github.com/user-attachments/assets/e62ab148-dd50-438d-a636-dbaca518d39f


VikaCep:
@vesnadean this is what we discussed over on Slack. Added the requirement to type "Delete", added the number of checks to be deleted in the confirmation modal, and in non-folder views, clicking the select-all checkbox only selects the checks in the current page.

vesnadean (approved):

PR #1855, opened 2026-09-16
fix: honour the ?features= URL override for OpenFeature-routed flags

g3john:
Follow-up to #1726.

## Problem

`?features=<flag>` has long been the way to see a gated feature on any stack without changing its configuration. #1726 kept it for legacy toggles but dropped it for flags routed through OpenFeature, so as flags migrate the override quietly stops working for them, and the first migrated flags (`CheckSuggestions`, soon `Recommendations`) can only be turned on via GOFF.

## Solution

`useFeatureFlag` honours `?features=` for both backends. A flag named in the URL is reported enabled and ready without consulting the provider, so a forced flag never waits on OFREP. Routed flags answer to either their `FeatureName` or their OpenFeature key, which is the name people see in `deployment_tools`. `getBooleanFlag`, the non-React path, applies the same rule so the two can't disagree.

The override is a client-side short-circuit: the provider never sees it, so GOFF-side evaluation and any tracking derived from it record the real value. That is the same behaviour the legacy override has always had.

https://github.com/user-attachments/assets/df7b5473-df9b-42b7-8517-3f945682da7e


ckbedwell:
I'd go the other way and remove this completely as Grafana support the param `?featureControl` by default now.

https://ops.grafana-ops.net/?featureControl

image

g3john:
Chris' PR makes this obsolete. Closing this in favour of that
https://github.com/grafana/synthetic-monitoring-app/pull/1872

PR #1856, opened 2026-09-17
feat: select a probe by default

dgzlopes:

New checks start with no probe selected, which makes the "Test" button fail right away. 

In this PR, we change the default to _one probe_, instead of none (invalid) or all of them (expensive).

Slack thread: https://raintank-corp.slack.com/archives/C0C1TQQ98F9/p1789652585418319


src/components/Checkster/contexts/ChecksterContext.tsx
```diff
@@ -79,6 +80,29 @@ interface StashedValues {
   settings: Record<string, unknown> | undefined;
 }
 
+function getDefaultProbeId(probes: ProbeWithMetadata[], checkType: CheckType) {
+  const availableProbes = getAvailableProbes(probes, checkType).filter((probe) => !probe.deprecated);
+  const onlineProbes = availableProbes.filter((probe) => probe.online);
+  const defaultProbe = onlineProbes.find((probe) => probe.public) ?? onlineProbes[0] ?? availableProbes[0];
```
g3john, line 86: Should we put more thought into which probe we select first, instead of just the first one (I believe its sorted alphabetically)?

g3john (commented):
Is this solving an actual problem? Do customers not care about which probe(s) they use? I also imagine cases where they just create a check without even seeing which probe is selected

src/components/Checkster/contexts/ChecksterContext.tsx
dgzlopes: As I see it, offering a reasonable default doesn't mean customers can't pick other probes _if they care_ 👀 

This change should speed up check creation for the average user and enable the test button ASAP, so they don't have to jump to the probes tab and then go back to tweak the logic. That said, I used both DD and SolarWinds checks a lot before the acquisition, and both select a bunch of regions for you by default. 

Actually, if I remember correctly, DD selects all of them (and that's a bunch of $$$)

dgzlopes: > I also imagine cases where they just create a check without even seeing which probe is selected

Why do you feel this is a problem? It can be edited afterward.

> Should we put more thought into which probe we select first, instead of just the first one?

I didn't want to overcomplicate it. What do you have in mind? 👀

ckbedwell (approved):
This does solve a real problem. I've had a lot of conversations with people asking why we don't take a more opinionated stance on setting things up for them so there's less friction to getting started.

We're potentially creating an issue where we might see an imbalance of checks being created on the default probe we're choosing here but we'll cross that bridge when we get to it and make the probe picking logic better (one idea is we analyse what probes they use already and pre-pick them -- if a customer has 100 checks and they use the same 3 probes on 90% them then we should just select them on their behalf...)

PR #1857, opened 2026-09-17
fix: remove "Shared With Me" from the folder selector if empty

dgzlopes:

"Shared with me" appears in the folder picker as a selectable row, but it's a virtual folder that can't be selected.

image

This PR excludes it from the picker instead of showing it if it is empty (very common).

Slack thread: https://raintank-corp.slack.com/archives/C0C1TQQ98F9/p1789656887347699


dgzlopes:
bugbot run

ckbedwell (commented):
I'm on the fence with this one because I've never really used the Shared with me folder functionality. I just went to ops and selected the AI folder under Shared with me? I get I can't share the folder itself but this PR just hides the whole thing.

Image

Image

dgzlopes:
@ckbedwell, let me test this a bit. It should only be hidden if it is empty 👀

PR #1858, opened 2026-09-17
fix: hide version number in the probe picker

dgzlopes:

Users pick the k6 channel in the first step of the check wizard (and can still change it when editing), so repeating the version per probe in the picker is redundant. This PR removes it. 

That said, unsupported probes stay disabled, but the tooltip now shows which channels they do support.

**Before:**
image

**After:**
image

Slack thread: https://raintank-corp.slack.com/archives/C0C1TQQ98F9/p1789655869817869


dgzlopes:
bugbot run

g3john:
Only the major version is shown in the selector (ex: v2.x) so I wonder if it's useful to be showing the exact version. @VikaCep WDYT?

ckbedwell (approved):
I agree with this change as that badge is just noise for public probes especially. There's a small argument we should show the version for private probes but even then I'd lean towards it being noise.

I'm approving it but I'll let @dgzlopes decide if he wants to wait for any other opinions 😄

PR #1868, opened 2026-09-18
feat: add graft manifest file

ckbedwell:
Adds a `.graft.jsonc` so Graft v4 can load this checkout as one package and run our existing Yarn 4 install/dev workflow.

This repo ships two Grafana plugins from one webpack process: the app (`src/plugin.json` → `dist`) and the nested datasource (`src/datasource/plugin.json` → `dist/datasource`). Graft can infer a conventional single-plugin layout; the manifest is what lets it serve both units and expose setup/watch as package actions.

The `dev` workflow calls the vendored Yarn 4.15.0 binary directly (`node .yarn/releases/yarn-4.15.0.cjs …`) so Graft does not need a global `yarn` or corepack. Install is skipped when `node_modules` already exists. Watch is `yarn dev`.

image


VikaCep (approved):
Nice!

PR #1869, opened 2026-09-18
feat: add browser check templates for broken links and SSL

going-confetti:
## What changed

Adds a proof of concept for creating checks from templates (behind the `synthetic-monitoring.check-templates` feature flag). Users select a clickable card, enter a URL, then continue to the existing check form with an editable script and suggested name.

- Broken links and SSL certificate templates, both defaulting to a one-hour interval.
- SSL requires HTTPS and uses a 30-day expiry threshold. It remains browser-based for this proof of concept. API endpoint checks and simpler certificate-expiry alert setup are follow-up work.
- Responsive template cards with icons and check-type badges, gated by `synthetic-monitoring.check-templates` (default off).
- Analytics for template selection and draft generation, plus template attribution on existing check-form events. These additions do not send entered URLs or script contents.

## Testing instructions

Use the normal local Grafana setup with `yarn dev` and `yarn server`. Enable the flag in the existing `[feature_toggles]` section of `dev/custom.ini`:

```ini
[feature_toggles]
synthetic-monitoring.check-templates = true
```

Restart Grafana after changing the ini and hard-refresh the browser. No TypeScript default changes are needed. Use `yarn dev`, not `yarn dev:msw`, because the MSW flag endpoint returns an empty flag set. If switching from MSW, unregister its service worker first.

1. Open **Create check** and verify that both template cards appear under **Start from a template**.
2. Select **Detect broken links**, enter `https://grafana.com`, and continue. Verify the browser form has a suggested name, a script using the entered URL, and a one-hour interval.
3. Repeat with **Check SSL certificate**. Verify the script uses `warnDays: 30`, `failOnExpired: true`, and `failOnNearExpiry: true`.
4. Verify empty or invalid URLs show validation errors. The SSL template must also reject an HTTP URL. Cancel should return to the cards without creating a check.
5. Change the flag to `false`, restart Grafana, and refresh. The template section should disappear.

## Validation

- Type checking and targeted ESLint.
- Focused tests covering template handoff, validation, permissions/limits, feature flags, and analytics attribution.
- Analytics reference regenerated.

Live execution of the generated scripts against a probe has not been verified. The GoFF service definition in deployment_tools is not included in this PR.


src/page/ChooseCheckGroup/ChooseCheckGroup.tsx
```diff
@@ -29,6 +33,20 @@ export const ChooseCheckGroup = () => {
               return <CheckGroupCard key={group.label} group={group} />;
             })}
           </div>
+          <FeatureFlag name={FeatureName.CheckTemplates}>
+            {({ isEnabled }) => isEnabled ? (
+              <Box marginTop={2}>
```
going-confetti, line 38: Image

src/page/ChooseCheckGroup/components/CheckTemplate.tsx
```diff
@@ -0,0 +1,98 @@
...
+        description={template.description}
+        checkType={checkTypeOption.label}
+        disabled={disabled}
+        onSelect={() => {
+          trackCheckTemplateSelected({ check_template_id: template.id });
+          setIsOpen(true);
+        }}
+      />
+      {isOpen && (
+        <Modal title={template.title} isOpen onDismiss={() => setIsOpen(false)}>
```
going-confetti, line 65: Not a perfect solution by any means, we'll try to improve the UX of passing the URL to the template in a follow-up PR

Image

PR #1870, opened 2026-09-18
feat: experimental FEO context panel in Timepoint Explorer

markjmeier:
## Summary

Adds an experimental "Real user context" panel to the browser check Timepoint Explorer, bridging Faro RUM data (Frontend Observability) alongside a check execution's own results — so a user can answer "if I have FEO set up, is my check's failure/pass representative of what real users are actually seeing?"

**This is explicitly a draft, sent early to collect feedback from FE** before committing to the current design/data model. Please don't read this as review-ready — read it as "does this direction make sense."

For a failed check, the panel checks whether real users are hitting the same failure (shared exception, or a named action's real-user failure rate). For a passed check, it checks representativeness: has the app deployed recently, and does the check's own timing diverge meaningfully from real users' (in either direction — a check that's *unrepresentatively fast* is the more dangerous case, since its pass reads as reassurance that isn't earned).

Everything is read from the same Faro session the existing "View Frontend Session" link already points to, so this and that feature always tell the same story.

## Where to start reviewing

Given the size, two natural entry points:
- `FrontendContext.utils.ts` — the data model and rules engine (pure functions: parsing Faro records into pages/actions/exceptions, the LogQL query builders, the fidelity-verdict and summary-verdict logic). No React in here — probably the higher-value read for judging whether the *logic* makes sense.
- `FrontendContext.tsx` — the panel UI itself, consuming the above.

`FrontendContext.hooks.ts` is the thin React Query layer connecting the two.

## Known gaps (not fixing in this PR)

- **No automated test coverage.** None of the three new files have tests yet. Deliberate for this draft — didn't want to lock in test assertions against a data model and rules engine that's still up for debate. Should be added before this leaves experimental status.
- **No feedback-collection mechanism.** This ships as a plain panel with no way for a user to tell us what they think of it — the app's existing feedback-widget pattern would need to be wired up here too. Flagging as a real gap, but intentionally leaving it for a follow-up rather than doing it in this PR.
- **Nesting page vitals/requests under their owning action** (rather than the current flat page-then-action layout) was raised as a design improvement but is a bigger architecture change, deferred out of this pass.

## Test plan

- [ ] Manual: open a browser check's Timepoint Explorer against an app that has Frontend Observability configured, for both a passing and a failing execution, and confirm the panel's verdict/detail match what FEO's own app shows for the same session.
- [x] `yarn typecheck`, `yarn lint`, `yarn test src/scenes/components/TimepointExplorer` all pass.

### DIFF
--- a/src/scenes/components/TimepointExplorer/FrontendContext.hooks.ts
+++ b/src/scenes/components/TimepointExplorer/FrontendContext.hooks.ts
@@ -1,0 +1,343 @@
+import { useQuery } from '@tanstack/react-query';
+import { DataFrame, FieldType } from '@grafana/data';
+import { parseLokiLogs } from 'features/parseLokiLogs/parseLokiLogs';
+import { queryDS } from 'features/queryDatasources/queryDS';
+import { queryLoki } from 'features/queryDatasources/queryLoki';
+
+import { useLogsDS } from 'hooks/useLogsDS';
+import {
+  AppVersionChange,
+  buildAppVersionHistoryLogQL,
+  buildExceptionRealSessionsLogQL,
+  buildFaroExecutionContextLogQL,
+  buildRealUserExceptionsLogQL,
+  buildRealUserHttpErrorsLogQL,
+  buildRealUserPageLoadsLogQL,
+  buildRealUserVitalP75LogQL,
+  buildSimilarSessionsLogQL,
+  FaroExecutionContext,
+  getAppVersionChange,
+  parseFaroExecutionContext,
+  parseSimilarSessions,
+  SimilarSession,
+  WEB_VITALS,
+  WebVitalName,
+} from 'scenes/components/TimepointExplorer/FrontendContext.utils';
+
+const REF_ID_FARO_EXECUTION_CONTEXT = 'faroExecutionContext';
+
+interface UseFaroExecutionContextProps {
+  executionId: string;
+  from: number;
+  to: number;
+  enabled?: boolean;
+}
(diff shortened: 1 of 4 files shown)


markjmeier:
Screenshots. Honestly, the section is huge and has zero design consideration. its information overload, but its information thats available and could be displayed for user using both FEO and SM.  I very much want/need real user feedback on this connection so we can improve it, which is part of the intent of the PR to begin with. Basically lets throw a bunch of thigns on the page then trim it to what we believe is useful/usable by end users.

image

image


My broad thoughts:

- the app version update seems like one of the most useful pieces of info
- claude got cute and tried to give a one sentence summary. i think its a good idea, but we need to probably just be better about what it says.  An at a glance summary is desirable and easy to read, but maybe thats why assistant is there and we can just use the assistant button
- user actions - if a customer uses user actions in FEO, they return timings.  I like that we can encourage users to set those up in FEO and then we can compare our run vs real users.  But - is this useful to see?
- page load time we get from FEO while we lack the page timing in k6 🫠 . i had a hard time getting the page load time from subsequent pages/seemed to only be present if we also have web vitals
- real user sessions with similar journey.  Ive been testing against a demo env which uses k6 browser for background load. So no real users against it. This, in part, is why im listing all the sessions, so i could find intersting ones.  the pattern is bad and the search bar for session id has potentially questionable usefulness.  We could seem to detect sessions that hit the same pages in the same order as well as other sessions that only hit some of them - this is useful.

PR #1872, opened 2026-09-18
fix: respect Grafana Feature control overrides

ckbedwell:
Feature control can show an enabled local override while Synthetic Monitoring still evaluates the server value. Compose Grafana's shared local override provider before the existing OFREP provider so mapped OpenFeature flags respect both true and false overrides. Removing an override restores server evaluation, and changes notify plugin consumers without a reload.

Target the latest Grafana Cloud runtime through a normal named import. Update the aligned Grafana packages to 13.2.2 and remove the older-host fallback, compatibility cast, and import lint suppression. AGENTS.md now documents this Cloud runtime policy.

The package upgrade requires React 19. Align React and its types, upgrade Scenes to 8.17.0 (the latest release allowed by the repository's three-day package age policy), and make the associated ref, event, popover, and test-harness changes. Add @react-hookz/web as a direct runtime dependency for React 19-compatible resize observers; this package is also used by the upgraded Grafana dependencies. Preserve accessible names on the view-switcher controls and update affected test selectors.

Legacy config.featureToggles flags remain outside Feature control's override path.

Validation:
- Six focused Feature control regression cases.
- Full TypeScript and ESLint checks; ESLint has existing console warnings.
- Yarn immutable install succeeds. Upstream peer-range warnings remain, including @grafana/alerting declaring Grafana <=12.x despite its 13.2.2 release; no force flags or peer bypasses were used.
- Full Jest suite: 250 suites and 2,232 tests passed (one suite skipped, seven tests skipped, one todo), using two workers.
- Browser/Graft verification and a production build have not been run.


src/services/featureFlags.ts
```diff
@@ -1,6 +1,6 @@
-import { config } from '@grafana/runtime';
+import { config, createOpenFeatureLocalStorageProvider } from '@grafana/runtime';
```
g3john, line 1: Should we bump `grafanaDependency` in `src/plugin.json` to `>=13.2.0-0` in order to use `createOpenFeatureLocalStorageProvider`?

g3john (approved):
:shipit:

PR #1873, opened 2026-09-18
feat(knowledgeGraph): improve Connected Services component

VikaCep:
## What

The Connected services section now renders insights the way the Knowledge Graph itself does, and disambiguates the environment fan-out that comes from linking an env-less check to services by name+namespace.

**Own vs connected insight split (mirrors the KG's `assertion` / `connectedAssertion`):**

- Dual rings: the outer ring carries the service's own insights; a service with check-propagated insights gets an inset connected ring. The insight count badge is removed — the KG's entity graph encodes severity purely in rings and has no count badge.
- Grouped node card: "Service insights" vs "Propagated from check", each row with a severity-colored icon (matching the KG drawer's row anatomy) instead of the previous dots that all carried the node's overall severity. Icons carry a `title` so color isn't the only severity signal. Long assertion names wrap in their own text column instead of orphaning the marker.
- The card's "Open in Knowledge Graph" link now opens in a new tab, matching the section header button.

**Environment disambiguation (one env-less check fans out to a service's twin in every environment):**

- Services whose display name repeats across environments carry an `Env: <env>` sublabel under the node (the SVG reserves an extra label line so nodes don't move).
- Edge hover tooltips are env-qualified (`otel-demo/frontend (prod) → …`) so parallel edges to environment twins are tellable apart; the edge arrowheads are removed in favor of these named tooltips.
- The node card labels its scope line (`Service · Env: prod · Namespace: x`) instead of showing bare values.
- The check's own card states the fan-out explicitly: "Linked to otel-demo/frontend in 3 environments: deaf98, otel-demo, staging" — the graph alone can't tell which environment the check's target actually hits.

## How

One batched request to the KG datasource's assertions API (`includeConnectedAssertions: true`) fetches assertion timelines for all of the neighbourhood's services when the graph renders, refetching on the same interval as the neighbourhood query. Connected insights are classified by explicit cross-scope provenance labels (`asserts_origin_*`, `asserts_source_entity_type=SyntheticCheck`, non-aggregated source name) — never by name matching.

Connected insights are derived from the timelines rather than the Cypher frame's `insightNames`: on some KG versions the frame lists the entity's own assertions only, so a propagated insight can exist solely in the timelines (verified on a dev stack where the propagated check failure was missing from the frame). Per-insight severities also come from timeline labels, falling back to the node's overall severity. Loading or lookup failure degrades to the previous behavior — the frame's mixed single ring and flat insight list, plus an "Insight origin unavailable" note on the card.

## Test support

- New msw handler for the KG assertions endpoint and a fixture covering all three classification cases (provenance-labelled, own, and timeline-only propagated insight).
- New test id for the connected ring.

## Testing

`yarn jest src/features/knowledgeGraph` — 72 tests, including: batched request shape, timeline-only propagated insights surfacing without frame support, the connected ring rendering on exactly the affected node, error fallback, provenance strictness (no inference from name or source type alone, aggregated label arrays rejected), scope-mismatch isolation between same-named entities, env sublabels only on repeated names, env-qualified edge titles, and the check card's fan-out note.

Verified live against a dev stack: a check failure propagated to the linked service renders as the inset connected ring + "Propagated from check" group, matching the KG entity graph's rendering of the same entities.

https://ops.grafana-ops.net/a/grafana-synthetic-monitoring-app/checks/1552?from=now-30m&to=now&timezone=America%2FArgentina%2FBuenos_Aires&var-probe=$__all&var-job=grafana.com%20homepage&var-instance=https:%2F%2Fgrafana.com%2F
image

image

image
image


VikaCep:
bugbot review

VikaCep:
Closing in favor of https://github.com/grafana/synthetic-monitoring-app/pull/1874

PR #1875, opened 2026-09-21
feat: redesign script editor toolbar

dgzlopes:

This PR redesigns the script editor toolbar by:
- Moving k6 version selection to the script editor itself.
- Adding a help menu (load example/docs).
- Adding an AI-assisted generation entry point.
- Adding an expand button for the editor.

This consolidates the existing feature set and makes it easier to ship new experiments (e.g., Playwright) faster.

https://github.com/user-attachments/assets/5c773e6a-3010-4f31-9b00-cd2d956ea8e3

Related Slack thread: https://raintank-corp.slack.com/archives/C0C1TQQ98F9/p1790012128810089


dgzlopes:
bugbot run

src/components/Checkster/components/form/generic/GenericScriptField.tsx
```diff
@@ -45,8 +53,46 @@ export function GenericScriptField({ field }: GenericScriptFieldProps) {
...
+      className={cx(
+        css`
+          & > div[data-fs-element='Code editor'] {
+            flex: 1 1 0;
+            overflow: visible;
+          }
+          & > div > div {
+            min-height: unset; // code editor
+          }
+        `
```
ckbedwell: 🤮 Can we make this cleaner? At the least put it in styles block with the other classes and put a note in that it is abusing child styles. Can we just add these styles to the code editor or have an injection prop to handle this usecase?

src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -260,12 +271,17 @@ export const CodeEditor = forwardRef(function CodeEditor(
       ${containerStyles};
       height: ${editorHeight}px;
       min-height: ${MIN_EDITOR_HEIGHT}px;
+      resize: vertical;
+      overflow: auto !important;
+      // Outer wrapper already draws the frame; drop Grafana's default inner one.
+      border: none !important;
+      border-radius: 0 !important;
```
ckbedwell: Do we need the `!important`s here? Is there no another way to handle this?

src/components/Checkster/components/form/generic/GenericScriptField.tsx
```diff
@@ -29,8 +34,11 @@ export function GenericScriptField({ field }: GenericScriptFieldProps) {
   const fieldErrorProps = getFieldErrorProps(errors, field);
 
   const theme = useTheme2();
+  const styles = useStyles2(getStyles);
 
-  const k6ChannelId = (getValues('channels.k6') as K6Channel | undefined)?.id;
+  const k6Channel = getValues('channels.k6') as K6Channel | undefined;
+  const k6ChannelId = k6Channel?.id;
+  const isDeprecatedChannel = !!k6Channel && new Date(k6Channel.deprecatedAfter) < new Date();
```
ckbedwell: Can we lift this out and make it injectable as a prop to make the GenericScriptField Generic?

One day the script component might be supporting scripts that aren't k6 (Playwright... 😅) so I'd rather we future-proof now rather than adding tech debt.

ckbedwell (changes requested):
I know this is still in draft but as I was passing by 😉

dgzlopes:
bugbot run

src/components/Checkster/components/form/generic/GenericScriptField.tsx
dgzlopes: GenericScriptField is k6-agnostic now; k6 logic moved to ScriptedCheckContent 👀

src/components/CodeEditor/CodeEditor.tsx
dgzlopes: We don't. Good catch! Fixed.

src/components/Checkster/components/form/generic/GenericScriptField.tsx
dgzlopes: Yup. Agree. I moved them to the code editor.

